PR #321, opened 2024-01-31
🚧 Fix some minor issues with build-lz-options

janjakubnanista:
### In this PR

- Fix small typos in the `build-lz-options` README
- Silence the console for `build-lz-options` since we were getting warnings for `bigint-buffer` [see issue here](https://github.com/no2chem/bigint-buffer/issues/31). For this, a new `io-devtools` function has been introduced - `withMutedConsole` and the code that imports `lz-v2-utilities` has been wrapped in an async `import`


ryandgoulding (approved):

janjakubnanista:
Closing in favor of a smaller #329 - the native bindings issue (just a console warning) turns out to be more work than it's worth. Will be addressed later

PR #329, opened 2024-02-01
🪲 Wording fixes for build-lz-options

janjakubnanista:
### In this PR

- Fixes to the `README` of `build-lz-options`
- Small wording change to the prompt that asks whether we want to keep adding options


sirarthurmoney (approved):
📭

PR #330, opened 2024-02-02
🗞️ Retry helpers

janjakubnanista:
### In this PR

- After exploring the options of retrying failed calls on several levels (SDK, configuration and RPC provider), I created composable retry functionality that can be used on all three levels - yet after seeing that retrying on the provider level is very error-prone, I think we should retry on the configuration level.
- The way the code is structured, allows us to create retry strategies that will for example bump the gas limit on transactions if we need to

```typescript
// This is not a financial advice
//
// This strategy will be bumping the. `gasLimit` on a transaction to infinity
// exponentially by factor of 2
//
// (Being a sketch, it does not account for gasLimit being a string or being 0)
const strategy: RetryStrategy<OmniTransaction> = (attempt, error, [previousTransaction]) => [{
  ...previousTransaction,
  gasLimit: previousTransaction.gasLimit * 2
}]

// This will create a wrapper that can retry any function that matches the strategy signature
const retryWithMyStrategy = createRetryFactory(strategy)
const retriedSignAndSend = retryWithMyStrategy(signAndSend)
await retriedSignAndSend({ data: '0x...' }) 
```

### DIFF
--- a/packages/devtools/src/common/promise.ts
+++ b/packages/devtools/src/common/promise.ts
@@ -1,5 +1,6 @@
 import { Factory } from '@/types'
 import assert from 'assert'
+import { backOff } from 'exponential-backoff'
 
 /**
  * Helper type for argumentless factories a.k.a. tasks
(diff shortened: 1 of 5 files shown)


packages/devtools/src/common/promise.ts
```diff
@@ -72,3 +73,131 @@ export const firstFactory =
     <TInput extends unknown[], TOutput>(...factories: Factory<TInput, TOutput>[]): Factory<TInput, TOutput> =>
     async (...input) =>
         await first(factories.map((factory) => () => factory(...input)))
+
+/**
+ * RetryStrategy is a way of having a fine grain control
```
lzBryan, line 78: nit: RetryStrategy does this, but it is not that, what is it? I guess I am asking about the type or pattern in which it fits.

packages/devtools/src/common/promise.ts
```diff
@@ -72,3 +73,131 @@ export const firstFactory =
...
+ * over execution of retried function.
+ *
+ * It will be executed on every failed attempt and has the ability to modify the
```
lzBryan, line 81: What is It in here? Can we be more explicit?

packages/devtools/src/common/promise.ts
```diff
@@ -72,3 +73,131 @@ export const firstFactory =
...
+ * // whose argument is the amount of money we want to pay for a service
+ * const functionThatCanFail = (money: number): Promise<void> => { ... }
+ *
+ * // We can create a strategy that will keep adding 1 to the amount of money
+ * const strategy: RetryStrategy<[money: number]> = (attempt, error, [previousMoney]) => [previousMoney + 1]
+ *
+ * // Or we can create a strategy that will adjust the money based on the initial value
+ * //
+ * // In this made up case it will take the original amount and will add 2 for every failed attempt
+ * const strategy: RetryStrategy<[money: number]> = (attempt, error, _, [originalMoney]) => [originalMoney + attempt * 2]
```
lzBryan, line 100: because originalMoney is the name of the parameter each time, it seems this actually compounds right?
So its not like 10, 12, 14,... it would be 10, 12, 16, ...

packages/devtools/src/common/promise.ts
```diff
@@ -72,3 +73,131 @@ export const firstFactory =
...
+ *
+ * // Or we can go insane with our logic and can, because without objective morality
+ * // everything is permissible, update the amount on every other attempt
```
lzBryan, line 103: In nature, everything that is not necessary is allowed to change. Why should we hold ourselves to a different standard?

packages/devtools/src/common/promise.ts
```diff
@@ -72,3 +73,131 @@ export const firstFactory =
...
+ * const strategy: RetryStrategy<[money: number]> = (attempt, error, _, [originalMoney]) => attempt % 2 ? [previousMoney + 1] : true
```
lzBryan, line 104: A Collatz retrier would be cool.

lzBryan:
This helper is great, but it would be greater if we could augment it with a way to actually execute the retries instead of relying on client code to manage that end of the contract. (Unless its already there and I missed it), as well as a mechanism to introduce pauses in between retries (unless thats not relevant in the domain in which this will be used)

lzBryan (changes requested):

packages/devtools/src/common/promise.ts
janjakubnanista: `RetryStrategy` is a type that represents a function. This function with its signature is used in `createRetryFactory` to control how and if retrying will be performed.

packages/devtools/src/common/promise.ts
janjakubnanista: Fixed the code comments here not to confuse people - the `_` parameter placeholder has been replaced

packages/devtools/src/common/promise.ts
janjakubnanista: I will ask @St0rmBr3w whether that's something the users are after :)

packages/devtools/src/common/promise.ts
janjakubnanista: True - I will create a philosophy debt JIRA ticket to fix this ASAP

lzBryan (approved):

ryandgoulding:
There are a few problems I envision with this.

1.  Usually, transactions are stuck in mempool because the gasPrice is to low, not gasLimit
2. Bumping the gasLimit will do 1/2 things.  It will occasionally fix an issue where ethers estimates the gasLimit too low (good).  It will also cause things caught by the RPC that it suspects would cause an error to fail on chain (bad, as it costs money now).
3. We may hit the block gas limit, in which case it becomes impossible to even cancel the transaction.

PR #332, opened 2024-02-02
🧹 Preparation work for lz:deploy task

janjakubnanista:
### In this PR

- Miniscule code reorganization needed for `lz:deploy` task - moving some schemas around, exporting `isLogLevel` function and removing `CommaSeparatedValuesSchema` in favor of a transformation function that splits a CSV string into individual values, discarding any whitespace


sirarthurmoney (approved):
🪵

PR #338, opened 2024-02-05
 🪲Fixing configureEnforcedOptions

sirarthurmoney:
### In this PR

- Fixing configureEnforcedOptions logic to collect EnforcedOptions[] by OApp before executing the transaction. This results in fewer transactions to configure enforced options.

### DIFF
--- a/packages/ua-devtools-evm/src/oapp/sdk.ts
+++ b/packages/ua-devtools-evm/src/oapp/sdk.ts
@@ -97,15 +97,7 @@ export class OApp extends OmniSDK implements IOApp {
     protected serializeExecutorOptions(
         oappEnforcedOptionParam: OAppEnforcedOptionParam[]
     ): SerializedEnforcedOptions[] {
-        const serializedEnforcedOptions: SerializedEnforcedOptions[] = []
-        for (const oAppEnforcedOptionParam of oappEnforcedOptionParam) {
-            serializedEnforcedOptions.push({
-                eid: oAppEnforcedOptionParam.eid,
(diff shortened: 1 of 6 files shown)


packages/ua-devtools-evm/src/oapp/sdk.ts
```diff
@@ -71,42 +71,49 @@ export class OApp extends OmniSDK implements IOApp {
...
+    ): SerializedEnforcedOptions[] {
+        const serializedEnforcedOptions: SerializedEnforcedOptions[] = []
+        for (const oAppEnforcedOptionParam of oappEnforcedOptionParam) {
+            serializedEnforcedOptions.push({
+                eid: oAppEnforcedOptionParam.eid,
+                msgType: oAppEnforcedOptionParam.option.msgType,
+                options: oAppEnforcedOptionParam.option.options,
+            })
         }
+        return serializedEnforcedOptions
```
janjakubnanista, line 108: This is just a code-style thing - in a situation like this, I'd normally go for a `.map` rather than pushing into an array. There is nothing wrong with this approach ofc and it works as expected - what `.map` enforces is a concept of immutability - basically the fact that once an object (an array in this case) is created, it is not being changed. The code also gets a bit smaller:

```typescript
return oappEnforcedOptionParam.map(({ eid, option: { msgType, options } }) => ({ eid, msgType, options })
```

janjakubnanista (dismissed):
Looks good, take a look at the comment about the serialization, that's the only thing

janjakubnanista (approved):

PR #344, opened 2024-02-06
🪲 Introduce node engine versions and improve error reporting from create-lz-oapp

janjakubnanista:
### In this PR

- Add `engines` to examples' `package.json` files
- Add `--log-level` optional CLI flag to `create-lz-oapp` and log activity during cloning and installation for debugging purposes
- Improve error reporting from failed dependency installs. The error now looks something like this:

Screenshot 2024-02-06 at 9 57 47 AM

I am going to try and think how we could trigger this error on purpose so that we can add a user test but understandably it's quite tricky (the easiest would be to keep a branch that has nothing in it, just an empty NPM project with some errors in `package.json`, then we'd point the CLI to use that package using the `LAYERZERO_EXAMPLES_REPOSITORY_REF` environment variable)


packages/create-lz-oapp/src/utilities/installation.ts
```diff
@@ -1,9 +1,22 @@
...
 export const installDependencies = (config: Config) =>
     new Promise<void>((resolve, reject) => {
+        const logger = createModuleLogger('installation')
+
+        // We'll store combined stdout and stderr in this variable
+        const std: string[] = []
+
+        // This function will handle stdout/stderr streams from the child process
+        const handleStd = (chunk: string) => {
+            std.push(chunk)
```
ryandgoulding, line 15: Could cause overflow, but unlikely in this case.

ryandgoulding (approved):

ryandgoulding:
oh, this was green when i checked last.  will check again soon after update goes through

packages/create-lz-oapp/src/utilities/installation.ts
janjakubnanista: Yeah we roughly know how much output these produce so that should be fine

PR #347, opened 2024-02-06
🗞️ warn user if no lzReceive gas is supplied

ryandgoulding:
Warn the user if:
* There is no lzReceive option.  compose gas does not count towards lzReceive gas for the default ExecutorFeeLib implementation.
* lzReceive option is specified, but with 0 destination gas.

### DIFF
--- a/packages/build-lz-options/src/components/outputOptions.tsx
+++ b/packages/build-lz-options/src/components/outputOptions.tsx
@@ -15,8 +15,14 @@ export const OutputOptions: React.FC<OptionOutputProps> = ({
 }: OptionOutputProps) => {
   return (
     <Box flexDirection="column">
+      {props.warning.length > 0 && (
+        <Text color={"yellow"}>
+          <Text bold={true}>Warning:</Text> {props.warning}
+        </Text>
+      )}
       <Text>
-        Result: <Text color={"green"}>{props.hex}</Text>
+        <Text bold={true}>Result:</Text>{" "}
+        <Text color={"green"}>{props.hex}</Text>
       </Text>
     </Box>
   );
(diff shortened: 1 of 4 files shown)


janjakubnanista:
🗞️ is for a feature :)

packages/build-lz-options/src/components/outputOptions.tsx
```diff
@@ -15,8 +15,14 @@ export const OutputOptions: React.FC<OptionOutputProps> = ({
...
     <Box flexDirection="column">
+      {props.warning.length > 0 && (
+        <Text color={"yellow"}>
+          <Text bold={true}>Warning:</Text> {props.warning}
+        </Text>
+      )}
       <Text>
-        Result: <Text color={"green"}>{props.hex}</Text>
+        <Text bold={true}>Result:</Text>{" "}
+        <Text color={"green"}>{props.hex}</Text>
```
janjakubnanista, line 25: Truly not a functional comment, just FYI code-style:

Truthy boolean attributes don't require a value in React:

```jsx
<Text bold>Bold text</Text>
```

And string attributes don't require `{}`:

```jsx
<Text color="green">Green text</Text>
```

packages/build-lz-options/src/components/outputOptions.tsx
```diff
@@ -15,8 +15,14 @@ export const OutputOptions: React.FC<OptionOutputProps> = ({
 }: OptionOutputProps) => {
   return (
     <Box flexDirection="column">
+      {props.warning.length > 0 && (
+        <Text color={"yellow"}>
+          <Text bold={true}>Warning:</Text> {props.warning}
+        </Text>
+      )}
```
janjakubnanista, line 22: I think here `props.warning` is an array of strings so something like this would make more sense:

```jsx
<Box flexDirection="column">
  {props.warning.map(warning => (
    <Text color="yellow">
      <Text bold>Warning:</Text> {warning}
    </Text>
  )}
</Box>
```

janjakubnanista (changes requested):

janjakubnanista (approved):

PR #364, opened 2024-02-07
:beetle: ensure valid address in options builder CLI

ryandgoulding:


packages/build-lz-options/src/utilities/prompts.ts
```diff
@@ -93,6 +93,7 @@ const promptForNativeDropAddress: PromptObject<'nativeDropAddress'> = {
     name: 'nativeDropAddress',
     message: 'What native gas drop do you want to set?',
     initial: makeBytes32(),
+    validate: (str: string) => str.length <= 66, // 32 bytes is 64 nibbles + '0x'
```
janjakubnanista, line 96: `makeBytes32` will throw if passed an invalid `bytes32` value - you can use that to validate this (by `try/catch`)

janjakubnanista (changes requested):

packages/build-lz-options/src/utilities/prompts.ts
ryandgoulding: good point, will do.

packages/build-lz-options/src/utilities/prompts.ts
```diff
@@ -93,6 +93,14 @@ const promptForNativeDropAddress: PromptObject<'nativeDropAddress'> = {
     name: 'nativeDropAddress',
     message: 'What native gas drop do you want to set?',
     initial: makeBytes32(),
+    validate: (str: string) => {
+        try {
+            makeBytes32(str)
```
lzBryan, line 98: Addresses are 20bytes long, it might be the case that some strings pass this validation but are not valid addresses; do we have a makeBytes20 or makeAddress() ?

ryandgoulding: Different addresses can be different lengths.  For example:
- bitcoin is 32
- aptos is 32
- evm is 20

lzBryan (approved):

janjakubnanista (approved):

PR #372, opened 2024-02-08
🪲 Fix lz:deploy task not being able to find external deployments

janjakubnanista:
### In this PR

- The `run` function on `DeploymentsExtension` would clear its database if `resetMemory` wasn't explicitly set to `false`. This also cleared the external deployments and made `EndpointV2` deployment not accessible in the deploy script.
- The fix is very tricky to test so I introduced two user tests that basically check that `nox hardhat lz:deploy` will not fail on a missing deployment. First I made these tests fail ([see here](https://github.com/LayerZero-Labs/devtools/actions/runs/7833644196/job/21375087379)) and then introduced a fix. The trickiness of this test is the fact that we'd need to completely erase the runtime between every test - something we might still want to do. In the meantime I covered the task with basic tests and made sure that from the users' perspective the deploy task works [*]
- Added a small fix - The `--ci` should be a CLI flag, not a parameter that takes a boolean value

[*] Well what we do is we test that the deploy task fails, but it fails on insufficient funds. It might be nice to add a nightly test that uses a funded testnet mnemonic / private key to actually run all the deployment and wiring


packages/devtools-evm-hardhat/src/tasks/deploy.ts
```diff
@@ -177,11 +180,35 @@ const action: ActionType<TaskArgs> = async (
...
+                    //
+                    // In effect this means the deployments for LayerZero artifacts would not be available
+                    resetMemory: false,
                     writeDeploymentsToFiles: true,
+                    deletePreviousDeployments: reset,
                 })
 
+                // Now we do a little diff on what contracts had been changed
+                const contracts = Object.fromEntries(
+                    Object.entries(deploymentsAfter).filter(
```
ryandgoulding, line 207: This does not take into account if `Object.entries(deploymentsAfter)` is a subset of `deploymentsBefore`.  I.e., if `deploymentsBefore` actually has more deployments than after.  Thus, it is an incomplete diff.

janjakubnanista: Yeah this logic is very basic and it only cares about the new and/or changed deployments - we are not telling the user which contracts have been deleted.

ryandgoulding: Got it, so long as you are aware of and fine with this limitation, LGTM

ryandgoulding (approved):

PR #374, opened 2024-02-08
🗞️ Move sign & send logic into a subtask

janjakubnanista:
### In this PR

- Move the sign & send logic into a subtask so that it can be both reused & modified by users


sirarthurmoney (approved):
⛲️

sirarthurmoney (approved):
↪️

PR #385, opened 2024-02-10
🗞️ Investigate migration to Ethers v6

janjakubnanista:
### In this PR

- See if we can migrate to `ethers@6` without any major issues (unlikely - ~~the last time we tried we ran into transitive dependency issues with `zksync-web3`~~ we are now blocked on our own packages - `lz-v2-utilities` depends on `ether@5.7.2` and imports directly from `ethers/lib/utils`)


janjakubnanista:
We might want to revisit this later or just switch to `viem`, closing for now

Berbex:
Why the branch was deleted? Is not in the roadmap continue with the bump of ethers version? 🤔

PR #415, opened 2024-02-16
🗞️  gnosis safe signer support

ryandgoulding:
Allows users to optionally define gnosis safe settings in `hardhat.config.ts`. If defined, then a gnosis signer will be used.  If not defined, a normal ethers signer will be used.

`@safe-global/protocol-kit@1.3.0` is the newest ethers V5 compatible version, so we rely on this version and the corresponding versions for dependencies.

### DIFF
--- a/packages/devtools-evm-hardhat/src/transactions/signer.ts
+++ b/packages/devtools-evm-hardhat/src/transactions/signer.ts
@@ -3,6 +3,7 @@ import type { OmniSignerFactory } from '@layerzerolabs/devtools'
 import { GnosisOmniSignerEVM, OmniSignerEVM } from '@layerzerolabs/devtools-evm'
 import { createProviderFactory } from '@/provider'
 import { createGetHreByEid } from '@/runtime'
+import { ConnectSafeConfigWithSafeAddress } from '@safe-global/protocol-kit'
 
 export const createSignerFactory = (
     addressOrIndex?: string | number,
@@ -20,7 +21,7 @@ export const createGnosisSignerFactory = (
     addressOrIndex?: string | number,
     providerFactory = createProviderFactory(),
     networkEnvironmentFactory = createGetHreByEid()
-): OmniSignerFactory<GnosisOmniSignerEVM> => {
+): OmniSignerFactory<GnosisOmniSignerEVM<ConnectSafeConfigWithSafeAddress>> => {
     return pMemoize(async (eid) => {
         const provider = await providerFactory(eid)
         const signer = provider.getSigner(addressOrIndex)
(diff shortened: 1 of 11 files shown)


packages/devtools-evm-hardhat/src/transactions/signer.ts
```diff
@@ -1,16 +1,22 @@
...
     return pMemoize(async (eid) => {
         const provider = await providerFactory(eid)
         const signer = provider.getSigner(addressOrIndex)
 
-        return new OmniSignerEVM(eid, signer)
+        const env = await networkEnvironmentFactory(eid)
+        const safeConfig = env.network.config.safeConfig
+        return safeConfig
+            ? new GnosisOmniSignerEVM(eid, signer, safeConfig.safeUrl, safeConfig)
+            : new OmniSignerEVM(eid, signer)
```
janjakubnanista, line 20: Okay for now but in general this will not allow people to directly influence the output of this function (since the decision is made internally based on the hardhat config). I would prefer a more explicit way of doing this:

- Create a new factory for the safe signer of type `OmniSignerFactory<GnosisOmniSignerEVM>`
- Then you can, e.g. based on some CLI flag, pass this or that factory to `subtask.sign-and-send.ts`

This way the behavior can be controlled directly by the consumer

packages/devtools-evm/src/signer/sdk.ts
```diff
@@ -37,41 +64,91 @@ export class OmniSignerEVM implements OmniSigner {
...
-        // data?: BytesLike,
-        // chainId?: number
+    protected safeSdk: Safe | undefined
+    protected safeService: SafeServiceClient | undefined
 
-        // type?: number;
-        // accessList?: AccessListish;
+    constructor(
+        public override readonly eid: EndpointId,
+        public override readonly signer: Signer,
```
janjakubnanista, line 92: You don't need the `public override readonly` since the parent class defines these, just pass them to `super`

janjakubnanista (dismissed):

packages/devtools-evm-hardhat/src/transactions/signer.ts
ryandgoulding: Let's handle this in a follow-up.

TRileySchwarz:
I would add an optional flag that can be passed into task to override the gnosis, vs. commenting/uncommenting the network config in the event your project requires both versions of signers

ryandgoulding:
dependency alerts are not related to this change.

ryandgoulding:
`@safe-global/protocol-kit` and others depend on old web3 dependencies.  These are just warnings, they are not actual security vulnerabilities.

TRileySchwarz:
@SocketSecurity ignore npm/web3-bzz@1.10.4
@SocketSecurity ignore npm/web3-shh@1.10.4
@SocketSecurity ignore npm/web3@1.10.4

TRileySchwarz (approved):

PR #417, opened 2024-02-18
🧹 Remove prettier from plugins

Aca1995:
Hey everyone :) Upon reviewing the ESLint configuration in this project, I observed that the 'prettier' plugin was included within the plugins array.

Prettier is primarily designed as a code formatter, focusing on ensuring consistent code formatting. Conversely, ESLint is better suited for addressing **code-quality** concerns, such as identifying potential bugs and enforcing coding standards.

By using **eslint-plugin-prettier** package and retaining 'prettier' within the plugins array, we inadvertently assign ESLint tasks that are ideally handled by Prettier. This misalignment could lead to conflicts between ESLint's stylistic rules and Prettier's formatting standards, potentially affecting the consistency and effectiveness of the codebase.

To align with industry-standard practices, I propose removing 'prettier' from the plugins array in your ESLint configuration. This adjustment will enable Prettier to fulfill its intended role as the primary code formatter, while ESLint remains dedicated to code-quality analysis.

For your reference, you can find more information on this topic in the relevant section of the Prettier documentation where they stated that:

[Prettier Documentation: Integrating with Linters](https://prettier.io/docs/en/integrating-with-linters)

Thank you for considering this proposal. I welcome any feedback or suggestions regarding this adjustment.


janjakubnanista:
Hey @Aca1995 !

Thanks for the PR, looks good! One thing that also needs to change if you want to remove the prettier plugin is to adjust the scripts so that the `lint` and `lint:fix` will run both `prettier` and `eslint`. Could you make that change as well?

package.json
```diff
@@ -6,8 +6,8 @@
     "build": "$npm_execpath turbo run build",
     "clean": "$npm_execpath turbo run clean",
     "dev": "$npm_execpath turbo run dev",
-    "lint": "$npm_execpath turbo run lint",
-    "lint:fix": "$npm_execpath turbo run lint:fix",
+    "lint": "(pnpm prettier . --check || true) && $npm_execpath turbo run lint",
```
janjakubnanista, line 9: I think it's better to fail fast here - also this will also consume the prettier errors. Could you remove pls?

janjakubnanista (changes requested):

package.json
janjakubnanista: Also in this case we want to run `lint` in all workspaces using `turbo` so this command does not need to be adjusted, it's the workspace commands that need to be changed

janjakubnanista:
@Aca1995 please also rebase on the newest `main` otherwise your workflows will not run, thanks!

janjakubnanista:
Closing due to inactivity

PR #433, opened 2024-02-28
🪲 Add resolutions to examples

janjakubnanista:
### In this PR

- Adding resolutions for `@openzeppelin/contracts-upgradeable` to example projects since NPM installation in the user tests started failing because of a peer dependency requirement not being met

### DIFF
--- a/examples/oapp/package.json
+++ b/examples/oapp/package.json
@@ -18,6 +18,7 @@
   },
   "resolutions": {
     "@openzeppelin/contracts": "^5.0.1",
+    "@openzeppelin/contracts-upgradeable": "^5.0.1",
     "ethers": "^5.7.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


lzBryan:
Is this the right fix? Overrides seem like scartissue instead of addressing the undelying problem; which if I understand correctly, is under our control to solve. For example, providing a version that depends on OZ 5 only.

janjakubnanista:
@lzBryan yeah the ultimate fix is to use OZ@5 in monorepo. Until then we need to tell the package managers that we know what we are doing and that they should not fail screaming when we're trying to install OZ@5

TRileySchwarz:
Is this not resolved by doing something like this?

"@openzeppelin/contracts": "^4.8.1 || ^5.0.0", inside of the root packages?

PR #443, opened 2024-03-01
🧹 Set correct exit codes for failed hardhat tasks

janjakubnanista:
### In this PR

- Set correct error exit codes when hardhat tasks fail. This is necessary to be able to correctly detect the state of the task on the command line since in case of error we don't want to throw or exit (for composability reasons) to be able to pass the exact error state to the consumer [*]

[*] JavaScript errors are not typed so for us to be able to pass a typed information about the error state we would need to resort to a bit more advanced techniques like [`Either`](https://itnext.io/either-monad-a-functional-approach-to-error-handling-in-js-ffdc2917ab2) which we can do but not as of yet

### DIFF
--- a/packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
+++ b/packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
@@ -61,10 +61,14 @@ const action: ActionType<TaskArgs> = async (
         )
     )
     // Now sign & send the transactions
-    const signAndSendResult = await hre.run(SUBTASK_LZ_SIGN_AND_SEND, {
+    const signAndSendResult: SignAndSendResult = await hre.run(SUBTASK_LZ_SIGN_AND_SEND, {
         transactions,
         ci,
     } satisfies SignAndSendTaskArgs)
+
+    // Mark the process as unsuccessful if there were any errors (only if it has not yet been marked as such)
+    const [, failed] = signAndSendResult
+    if (failed.length !== 0) process.exitCode = process.exitCode || 1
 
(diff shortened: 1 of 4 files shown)


packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
```diff
@@ -61,11 +61,15 @@ const action: ActionType<TaskArgs> = async (
...
     // Now sign & send the transactions
-    const signAndSendResult = await hre.run(SUBTASK_LZ_SIGN_AND_SEND, {
+    const signAndSendResult: SignAndSendResult = await hre.run(SUBTASK_LZ_SIGN_AND_SEND, {
         transactions,
         ci,
     } satisfies SignAndSendTaskArgs)
 
+    // Mark the process as unsuccessful if there were any errors (only if it has not yet been marked as such)
+    const [, failed] = signAndSendResult
+    if (failed.length !== 0) process.exitCode = process.exitCode || 1
```
ryandgoulding, line 71: Can we please use french braces, even for single line conditional clauses?

```
if (failed.length !== 0) {
    process.exitCode = process.exitCode || 1
}
````

TRileySchwarz: Im indiffeferent, but should just be consistent everywhere~

TRileySchwarz: Consensus is {}, will do in a seperate PR

TRileySchwarz (approved):

PR #455, opened 2024-03-10
doc(README): add in missing --oapp-config flag

nazreen:
closes #454

PR #456, opened 2024-03-10
🗞️ Default oappConfig path

### LINKED ISSUE #454: bug: README's wiring command is missing the --oapp-config flag
**Describe the bug**
The README says to run `npx hardhat lz:oapp:wire` when wiring contracts.

**To Reproduce**

1. Run `npx create-lz-oapp@latest` to scaffold a new LayerZero OApp/ONft
2. (Assuming you have a valid pk defined in the `.env` and that address has sufficient gas tokens on all chains) Run `npx hardhat lz:deploy`
3. Run `npx hardhat lz:oapp:wire`

**Expected behavior**
Wiring continues successfully.

**Screenshots**
image

**Environment (please complete the following information):**

- OS: macOS
- OS Version: 14.2.1
- VM: node
- VM Version: 18.19.0

**Additional context**

nazreen:
this is to resolve #454 

NOTE: either this PR or #455 should be solved, not both.

### DIFF
--- a/packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
+++ b/packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
@@ -14,13 +14,13 @@ import type { SignAndSendTaskArgs } from '@layerzerolabs/devtools-evm-hardhat/ta
 import './subtask.configure'
 
 interface TaskArgs {
-    oappConfig: string
+    oappConfig?: string
     logLevel?: string
     ci?: boolean
 }
 
 const action: ActionType<TaskArgs> = async (
-    { oappConfig: oappConfigPath, logLevel = 'info', ci = false },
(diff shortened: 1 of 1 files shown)


janjakubnanista:
@nazreen hello there! We decided to remove the default value since there is no good default here:

- Some projects might use JavaScript and the config extension will be `.js`
- Some projects might have more configs (e.g. `layerzero.config.testnet.ts` and `layerzero.config.mainnet.ts`)
- Combinations of the above

In order to support at lease the first point, we'd need to make the import extension agnostic, after that we can add the default value. Projects with more configs will then need to supply the `--oapp-config` explicitly.

janjakubnanista (changes requested):
To go ahead with the default value, please add an extension-agnostic config path resolution (along with tests). Thanks!

nazreen:
should we just close this one for now?

janjakubnanista:
Closing due to inactivity

PR #457, opened 2024-03-11
Patch: add in missing @return annotation tag in MyOApp.sol

wyattwestdev:
### **Purpose:**

- A patch to add in a missing `@return` annotation tag in the `quote` function from MyOApp.sol.

### **Overview:**

- The annotation tag description is as follows: "A `MessagingFee` struct containing the calculated gas fee in either the native token or ZRO token.".
- The annotation tag description is equivalent to the description defined in the `_quote` function from LayerZero-v2/oapp/contracts/oapp/OAppSender.sol.


janjakubnanista:
Thanks @wyattwestdev !

PR #478, opened 2024-03-13
🗞️ Add docker compose spec generation utilities for localnet [4/N]

janjakubnanista:
### In this PR

- Adding utilities & assets required to create a compose spec consisting of anvil-based EVM nodes & an nginx-based RPC proxy
- The proxy is added to avoid exposing N networking ports from the forked nodes - instead, these stay internal to the compose stack and the client needs to use the proxy to get to them
- In order to avoid copying files, the `Dockerfile` and `nginx.conf` file are imported as strings to the code itself using `raw`/`text` loaders. This required a `raw` loader for `jest` (instead of installing a package [with three outdated lines of code](https://github.com/keplersj/jest-raw-loader/blob/master/index.js) I just added a local version of the transformer - `jest.transformer.raw.js`) and a `text` loader for `esbuild`. In addition to this a type definition was required so that TypeScript is not freaking out over importing unknown file types
- Two additional arbitrariness for fuzzy testing were added to make generating optional types easier
- One new arbitrary for generating mnemonics

### DIFF
--- a/packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
+++ b/packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
@@ -2,10 +2,10 @@ events {}
 
 http {
   # We will modify the log  format to include the target_network
-  log_format unique '$remote_addr - $remote_user [$time_local] '
-                    '"$request" $status $body_bytes_sent '
-                    '"$http_referer" "$http_user_agent" '
-                    'Network: "$target_network"';
+  log_format proxied '$remote_addr - $remote_user [$time_local] '
+                     '"$request" $status $body_bytes_sent '
+                     '"$http_referer" "$http_user_agent" '
+                     'Network: "$target_network"';
 
   server {
     # This proxy server will listen on port 8545
@@ -35,8 +35,8 @@ http {
     # 
     # # http://localhost/fuji/some/url -> http://fuji:8545/
     location / {
-      # Set the log format to be our custom log format
-      access_log /var/log/nginx/access.log unique;
+      # Set the log format to be our custom 'proxied' log format
+      access_log /var/log/nginx/access.log proxied;
 
       resolver 127.0.0.11;
       autoindex off;
--- a/packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
+++ b/packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
@@ -25,7 +25,7 @@ http {
     }
 
     # In this section we'll proxy all the requests to this server
-    # to the respectative network nodes
+    # to the respective network nodes
     # 
     # The requests are proxied based on the first path segment:
     # 
--- a/packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
+++ b/packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
@@ -1,0 +1,56 @@
+events {}
+
+http {
+  # We will modify the log  format to include the target_network
+  log_format unique '$remote_addr - $remote_user [$time_local] '
+                    '"$request" $status $body_bytes_sent '
+                    '"$http_referer" "$http_user_agent" '
+                    'Network: "$target_network"';
+
+  server {
+    # This proxy server will listen on port 8545
+    # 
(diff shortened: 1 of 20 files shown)


packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+
+    # We will add a simple endpoint for healthcheck
+    location /health-check {
+      access_log	off;
+      error_log	off;
+      return 200 'ok';
+    }
+
+    # In this section we'll proxy all the requests to this server
+    # to the respectative network nodes
```
lzBryan, line 28: `respectative` -> `respective`

packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+    # 
+    # The requests are proxied based on the first path segment:
+    # 
+    # http://localhost/fuji -> http://fuji:8545/
+    # 
+    # For now the remaining path segments are not being preserved:
```
lzBryan, line 34: This might be a gotcha thats going to be hard to debug, it does not look hard to capture the remaining path in the regex, what do you think about preserving it?

packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
```diff
@@ -0,0 +1,56 @@
...
+    # 
+    # # http://localhost/fuji/some/url -> http://fuji:8545/
+    location / {
+      # Set the log format to be our custom log format
```
lzBryan, line 38: is `unique` our custom log format? Sounds like an option that is not custom.

packages/devtools-evm-hardhat/src/simulation/compose.ts
```diff
@@ -0,0 +1,93 @@
...
+        target: 'proxy-evm',
+    },
+    // This service will expose its internal 8545 port to a host port
+    //
+    // The internal 8545 port is hardcoded both here and in the nginx.conf file,
+    // the source for which is located in src/simulation/assets/nginx.conf
+    ports: [`${port}:8545`],
+    depends_on: pipe(
+        networkServices,
+        // This service will depends on the RPCs to be healthy
```
lzBryan, line 53: `depends` -> `depend`

lzBryan (changes requested):
There is a couple of typos, otherwise LGTM.

packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
janjakubnanista: For this particular use case, there's not much point since the only endpoint on the RPC we are interested in is `/` - that's where all the requests are being sent. So not much value for the added complexity

packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
janjakubnanista: Changed to `proxied`

packages/devtools-evm-hardhat/src/simulation/compose.ts
janjakubnanista: Removed that `'s'` and replaced it with a `''` using a regular expression service running in docker for portability reasons, to save time next time I need to do this.

It's called productivity, look it up.

lzBryan: I asked ChatGPT to create a script to look it up for me, then pipe it into text to speech so I can learn it while still reviewing CRs.

packages/devtools-evm-hardhat/src/simulation/assets/nginx.conf
janjakubnanista: Okay okay

lzBryan (approved):

packages/devtools-evm-hardhat/src/simulation/compose.ts
janjakubnanista: Productivity stand-off happening right here right now

PR #479, opened 2024-03-14
🗞️ Add workflow and fix typos

miles-six:
I fix many typos and want to add workflow for codespell check to eliminate typos in the future.

I have added a relatively simple filtering file to exclude certain words. 

If you find it meaningful, you can modify the filtering conditions according to your needs.

The code quality is exceptionally high, so sincerely hope my PR can help make this repository more standardized.


janjakubnanista:
Hey @miles-six! Thanks for the PR!

Since this is a node-based codebase for most of it, I'd rather use the [cspell](https://www.npmjs.com/package/cspell) from NPM, that should be an easy change (that way you can add it to the root `devDependencies` and forget about `requirements.txt`).

The second change I'd make is to add the spellcheck to the existing [code check workflow](https://github.com/LayerZero-Labs/devtools/blob/main/.github/workflows/reusable-test.yaml#L22) as a separate step, no need to spin up a new workflow just for this one.

packages/export-deployments/src/common/fs.test.ts
```diff
@@ -10,10 +10,10 @@ describe('common/fs', () => {
 
         it('should return a left either wrapped in error if readFileSync fails', () => {
             readFileSyncMock.mockImplementation(() => {
-                throw 'borken'
+                throw 'broken'
             })
 
-            expect(readFileSyncSafe('filename')).toEqual(E.left(new Error('borken')))
```
janjakubnanista, line 16: This one here is intentional, it's a broken _broken_

packages/export-deployments/src/common/fs.test.ts
```diff
@@ -28,10 +28,10 @@ describe('common/fs', () => {
 
         it('should return a left either if mkdirSync fails', () => {
             mkdirSyncMock.mockImplementation(() => {
-                throw 'borken'
+                throw 'broken'
             })
 
-            expect(mkdirSafe('some/dir/name')).toEqual(E.left(new Error('borken')))
```
janjakubnanista, line 34: Same here

packages/verify-contract/test/url.test.ts
```diff
@@ -2,7 +2,7 @@ import { tryCreateScanContractUrl, tryGetScanBrowserUrlFromScanUrl } from '@/com
 
 describe('url', () => {
     describe('tryGetScanBrowserUrlFromScanUrl', () => {
-        it.each([[''], ['borken url'], ['https://'], ['https:///'], ['scan.api']])(
```
janjakubnanista, line 5: Same here

packages/export-deployments/src/common/fs.test.ts
```diff
@@ -10,10 +10,10 @@ describe('common/fs', () => {
 
         it('should return a left either wrapped in error if readFileSync fails', () => {
             readFileSyncMock.mockImplementation(() => {
-                throw 'borken'
```
janjakubnanista, line 13: Same here

miles-six:
@janjakubnanista  Hey,sir,thanks for your response！
I'm not quite sure if what I'm doing is correct, but I would like to directly add the following modified code (I removed requirements.txt) to the end of .github/workflows/reusable-test.yaml. 

  # We'll collect the docker compose logs from all containers on failure
      - name: Store docker logs
        if: failure()
        uses: actions/upload-artifact@v3
        with:
          path: ./logs

# here is I would add
    codespell:
      runs-on: ubuntu-latest

      steps:
        - name: Checkout the repository
          uses: actions/checkout@v4

        - name: pip cache
          uses: actions/cache@v4
          with:
            path: ~/.cache/pip
            key: ${{ runner.os }}-pip
            restore-keys: ${{ runner.os }}-pip-

        - name: Install prerequisites
          run: sudo pip install codespell==2.2.5

        - name: Spell check
          run: codespell --config=./.codespell/.codespellrc

If this is incorrect, could you please guide me on what I should do?

miles-six:
@janjakubnanista  I have integrate codespell to existing [code check workflow](https://github.com/LayerZero-Labs/devtools/blob/main/.github/workflows/reusable-test.yaml#L22),and deleate requirements.txt

miles-six:
Hey @janjakubnanista sir
maybe I need to add some words to words.txt to skip , but it should after approving can start code test

.codespell/wordlist.txt
```diff
@@ -0,0 +1 @@
+crate
```
janjakubnanista, line 1: Please add _borken_ here as well, we actually use it :)

.github/workflows/reusable-test.yaml
```diff
@@ -113,3 +113,23 @@ jobs:
...
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/*.py') }}
+          restore-keys: ${{ runner.os }}-pip-
+
+      - name: Install Codespell
+        run: sudo pip install codespell==2.2.5
+
+      - name: Spell Check
+        run: codespell --config=./.codespell/.codespellrc
```
janjakubnanista, line 135: What I meant in my first was not to use the python version at all - use the node version so that we don't introduce new language to the stack please

janjakubnanista (changes requested):
Could you please use the node version of `cspell`? We don't want to be introducing python if it's not justified

.github/workflows/reusable-test.yaml
```diff
@@ -113,3 +113,21 @@ jobs:
...
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      - name: Install cspell
+        run: npm install -g cspell
```
janjakubnanista, line 130: You can replace this part with:

```yaml
      - name: Setup environment
        uses: ./.github/workflows/actions/setup-environment

      - name: Install dependencies
        uses: ./.github/workflows/actions/install-dependencies
```

Then add the `cspell` dependency as `devDependency` to the root `package.json` (either add it manually and run `pnpm install` or run `pnpm add -W -D cspell`). Don't forget to run `nvm use` before installing anything to make sure your node version matches the `.nvmrc`!

.cspell/cspell.json
```diff
@@ -0,0 +1,17 @@
+// cSpell Settings
+{
+  // Version of the setting file.  Always 0.2
+  "version": "0.2",
+  // language - current active spelling language
+  "language": "en",
+  // words - list of words to be always considered correct
```
janjakubnanista, line 7: I think JSON does not support comments, please remove these

.github/workflows/reusable-test.yaml
```diff
@@ -113,3 +113,21 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           path: ./logs
+
+  codespell:
+    runs-on: ubuntu-latest
```
janjakubnanista, line 118: Please also add `name: Check spelling` to make this job look pretty in the github actions UI

miles-six:
Hey @janjakubnanista  To avoid affecting the repository, I have set all the files in the ./ directory to be ignored. Additionally, for aesthetics and better management, I have moved the cspell settings to the package.json file.

miles-six:
@janjakubnanista If you need any configuration, please let me know at any time

janjakubnanista (changes requested):
There are two more issues with the setup - if you could fix those please.

.github/workflows/reusable-test.yaml
```diff
@@ -113,3 +113,16 @@ jobs:
...
+    steps:
+      - name: Setup environment
+        uses: ./.github/workflows/actions/setup-environment
+
+      - name: Install dependencies
+        uses: ./.github/workflows/actions/install-dependencies
+
+      - name: Spell Check
+        run: cspell "**/*" --config ./.cspell/cspell.json
```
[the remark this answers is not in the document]
miles-six: yes, cspell CLI -  this command exit with error code not correct

miles-six:
After several hours of effort, I managed to reduce the number of Unknown words from 2000 to 40. 
1.Although these remaining 40 may appear to be typos, they are actually part of a specific function, so I did not directly modify them. I'm unsure whether I should add these words to the ignore list or add the file to the ignore list for this particular case.

2.In the cspell.json file, I added some filtering words, including proper nouns.

3.The filtered paths can also be seen in the cspell.json file.
image

janjakubnanista:
@miles-six looks good, thanks! Could you just resolve the merge conflict in `pnpm-lock.yaml` (running `pnpm i` from root after hitting the conflict should do)? After that we can run the workflows and see what comes out

miles-six:
@janjakubnanista  I already resolved the merge conflict in pnpm-lock.yaml and runn pnpm i from root

.github/workflows/reusable-test.yaml
```diff
@@ -113,3 +113,16 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           path: ./logs
+
+  codespell:
+    name: Check spelling
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup environment
+        uses: ./.github/workflows/actions/setup-environment
```
janjakubnanista, line 122: [The workflow has unfortunately failed](https://github.com/LayerZero-Labs/devtools/actions/runs/8430221261/job/23116118945?pr=479) - the problem is that this job is missing the first step:

```yaml
steps:
  - name: Checkout repo
     uses: actions/checkout@v4
```

Once you put that one in you should be good and green!

janjakubnanista (changes requested):
Small thing to make the workflow pass - check my comment above. Thanks!

miles-six:
@janjakubnanista  I'm not sure about the exact cause of the error in the previous commit. If possible, I would appreciate it if you could take over and make the necessary modifications directly.

janjakubnanista:
> @janjakubnanista I'm not sure about the exact cause of the error in the previous commit. If possible, I would appreciate it if you could take over and make the necessary modifications directly.

Oh you're just an inch away @miles-six! The problem is that `cspell` command is not available globally. You can just simply replace `cspell` with `pnpm cspell` in your workflow and the problem will be gone.

miles-six:
@janjakubnanista hope you can help me sincerely

PR #500, opened 2024-03-19
🧹 Add default verify url for blast network

TRileySchwarz:
aded a default verify contract url for: 
blast


ryandgoulding (approved):
We also maintain a list in README.md, should we add it there

https://github.com/LayerZero-Labs/devtools/tree/main/packages/verify-contract

may want to get rid of it altogether too.

PR #505, opened 2024-03-21
chore:  add recommended commas to silence annoying IDE warnings

ryandgoulding:


ryandgoulding:
We also maintain a list in README.md, should we add it there

https://github.com/LayerZero-Labs/devtools/tree/main/packages/verify-contract

may want to get rid of it altogether too.

lzBryan (approved):

PR #513, opened 2024-03-25
🗞️ OwnableMixin + OApp = 🆒 [1/N]

janjakubnanista:
### In this PR

- Adding `OwnableMixin`, a collection of methods that can be mixed in to a class instance / class prototype to add `Ownable` SDk functionality
- Adding `createOwnableFactory` to wrap an arbitrary `OmniSDK` with `OwnableMixin`
- Adding `configureOwnable` function to configure ownership

At this point, the amount of very convoluted code that makes TypeScript happy about mixins does not justify the extra three literal methods on `OApp` that proxy the calls to `OwnableMixin`. Less repetitive solutions can be achieved with class decorators but the support is still experimental and could break TypeScript for lots of client projects - see here https://www.typescriptlang.org/docs/handbook/mixins.html.

On the other hand, having `OwnableMixin` out of the `OApp` prototype chain allows us to mixt it in with any SDK, not just `OApp` ones.

### DIFF
--- a/packages/ua-devtools-evm/src/ownable/mixin.ts
+++ b/packages/ua-devtools-evm/src/ownable/mixin.ts
@@ -1,0 +1,31 @@
+import { OmniAddress, OmniTransaction, areBytes32Equal, mapError } from '@layerzerolabs/devtools'
+import { OmniSDK } from '@layerzerolabs/devtools-evm'
+import type { IOwnable } from '@layerzerolabs/ua-devtools'
+
+export const OwnableMixin: IOwnable = {
+    async getOwner(this: OmniSDK): Promise<OmniAddress> {
+        this.logger.debug(`Getting owner`)
+
+        const owner = await mapError(
+            async () => this.contract.contract.owner(),
+            (error) => new Error(`Failed to get owner for OApp ${this.label}: ${error}`)
+        )
+
+        return this.logger.debug(`Got owner: ${owner}`), owner
+    },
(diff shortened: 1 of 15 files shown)


packages/ua-devtools-evm/src/ownable/mixin.ts
```diff
@@ -0,0 +1,31 @@
...
+export const OwnableMixin: IOwnable = {
+    async getOwner(this: OmniSDK): Promise<OmniAddress> {
+        this.logger.debug(`Getting owner`)
+
+        const owner = await mapError(
+            async () => this.contract.contract.owner(),
+            (error) => new Error(`Failed to get owner for OApp ${this.label}: ${error}`)
+        )
+
+        return this.logger.debug(`Got owner: ${owner}`), owner
```
ryandgoulding, line 14: For my own education, does this syntax just mean "return owner after I log this debug thingy"?

ryandgoulding (approved):

packages/ua-devtools-evm/src/ownable/mixin.ts
janjakubnanista: Yeah exactly, comma just means execute these expressions in sequence (and return the last one in this particular case). Works in many places, e.g.

```typescript
# Loops
for (let i = 0, j = 0; i++, j++; i < 10) {}

# Variables (much discouraged but very prevalent 20 years ago)
const I = 0, j = 0, k = 0;

# The return example above
return doSomething(), doSomethingElse(), getWhatNeedsToBeGot()
```

PR #533, opened 2024-04-02
🪲 Fix publish action, the saga continues

janjakubnanista:
### In this PR

- Instead of setting the safe directory, transfer ownership of the workspace directory to the root user (that's the user running the action)
- Fix a problem where environment variables are not expanded when passed as values for `env`

I am not proud but well that's about how much time I have to fix this.


lzBryan (approved):
grrr

PR #546, opened 2024-04-08
🧹 Improve error reporting for UIntBigIntSchema

janjakubnanista:
### In this PR

- When working on some projects with @TRileySchwarz , it transpired that forgetting a config value that uses `UIntBigIntSchema` results in `SyntaxError` errors that are completely undebuggable. These errors are not being caught by `zod` and carry no information about where they happened (which config property). To remedy this a more sturdy `zod` schema was introduced so that any `BigInt` constructor calls are wrapped with `try/catch`


sirarthurmoney (approved):
📜

lzBryan (approved):

PR #554, opened 2024-04-09
🗞️ Add --stage CLI argument to lz:deploy CLI task

janjakubnanista:
### In this PR

- Add `--stage {mainnet,testnet,sandbox}` CLI argument to `lz:deploy` so that we can deploy a range of networks without passing the whole list
  - This argument is ignored when `--networks` are also passed
  - In interactive mode, this argument will only affect the preselected networks and users can still add/remove networks to deploy
  - In CI mode this argument will act as a hard filter


packages/devtools-evm-hardhat/src/tasks/deploy.ts
```diff
@@ -85,9 +87,20 @@ const action: ActionType<TaskArgs> = async (
         logger.warn(`Failed to compile the project: ${error}`)
     }
 
+    // We'll warn the user that sending both stage & networks is not supported
+    // and that the --stage will be ignored
+    if (networksArgument != null && stage != null) {
+        logger.warn(`--stage ${stage} will be ignored since --networks argument has been passed`)
```
lzBryan, line 93: It would be safer to just throw here.

janjakubnanista: It would

janjakubnanista: It done

lzBryan (approved):

PR #558, opened 2024-04-10
🗞️ Retry functionality for SDKs

janjakubnanista:
### In this PR

- Adding `AsyncRetriable` decorator factory that can be applied to an `async` method and produces an `async` method that will retry its execution if it rejects
- This functionality is for now feature-flagged under `LZ_EXPERIMENTAL_ENABLE_RETRY` environment variable
- Since we compile `jest` test with `swc`, an additional `.swcrc` needed to be added for jest to be able to handle decorators

### DIFF
--- a/packages/devtools/src/common/retry.ts
+++ b/packages/devtools/src/common/retry.ts
@@ -1,5 +1,15 @@
+import { createModuleLogger, printJson } from '@layerzerolabs/io-devtools'
 import assert from 'assert'
 import { backOff } from 'exponential-backoff'
+
+export type OnRetry<TInstance, TArgs extends unknown[] = unknown[]> = (
+    attempt: number,
+    numAttempts: number,
+    error: unknown,
+    target: TInstance,
+    method: string,
+    args: TArgs
+) => boolean | void | undefined
 
 export interface RetriableConfig<TInstance = unknown> {
     /**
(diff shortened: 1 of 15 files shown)


packages/devtools/src/common/retry.ts
```diff
@@ -0,0 +1,104 @@
...
+     */
+    enabled?: boolean
+    /**
+     * The maximum delay, in milliseconds, between two consecutive attempts.
+     *
+     * @default Infinity
+     */
+    maxDelay?: number
+    /**
+     * Number of time the method call will be retried. The default is 3
```
lzBryan, line 26: Number of timeS

packages/devtools/src/common/retry.ts
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @default 3
+     */
+    numAttempts?: number
+    /**
+     * Callback called on every failed attempt.
```
lzBryan, line 32: Can you document whether this happens right after a failure or right before a retry?

packages/devtools/src/common/retry.ts
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @param {number} attempt
```
lzBryan, line 34: An attempt at documenting this was made

packages/devtools/src/common/retry.ts
janjakubnanista: I sure cannot, where would I get this information? Do you have a number I could call?

packages/devtools/src/common/retry.ts
janjakubnanista: Made them nicer

packages/devtools/src/common/retry.ts
lzBryan: What is this, the 90s? There is a chatbot that doesnt know either, but would love to help.

packages/devtools/test/common/retry.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+                expect(mock).toHaveBeenCalledTimes(3)
+                expect(mock).toHaveBeenNthCalledWith(1, 'y')
+                expect(mock).toHaveBeenNthCalledWith(2, 'y')
+                expect(mock).toHaveBeenNthCalledWith(3, 'y')
+            })
+        })
+
+        describe('when LZ_EXPERIMENTAL_ENABLE_RETRY is on', () => {
+            beforeEach(() => {
```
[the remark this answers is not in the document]
janjakubnanista: Added those

lzBryan (approved):

PR #603, opened 2024-05-08
🗞️ Add --stage CLI argument to the simulation task

janjakubnanista:
### In this PR

- Just like the `lz:deploy` task, `lz:test:simulation:start` task now accepts a `stage` CLI argument

### DIFF
--- a/packages/devtools-evm-hardhat/src/tasks/simulation/start.ts
+++ b/packages/devtools-evm-hardhat/src/tasks/simulation/start.ts
@@ -48,7 +48,7 @@ const action: ActionType<SimulationStartTaskArgs> = async (
     const isOnStage = stage == null ? () => true : (eid: EndpointId) => endpointIdToStage(eid) === stage
 
     // Let's grab the networks that will be included in the simulation
-    const networks: string[] = networksArgument
+    const networks = networksArgument
         ? // Here we need to check whether the networks have been defined in hardhat config
           assertDefinedNetworks(networksArgument)
         : //  But here we are taking them from hardhat config so no assertion is necessary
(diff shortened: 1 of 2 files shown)


packages/devtools-evm-hardhat/src/tasks/simulation/start.ts
```diff
@@ -32,12 +34,24 @@ const action: ActionType<SimulationStartTaskArgs> = async (
...
+    const isOnStage = stage == null ? () => true : (eid: EndpointId) => endpointIdToStage(eid) === stage
+
     // Let's grab the networks that will be included in the simulation
     const networks = networksArgument
         ? // Here we need to check whether the networks have been defined in hardhat config
           assertDefinedNetworks(networksArgument)
         : //  But here we are taking them from hardhat config so no assertion is necessary
-          Object.entries(getEidsByNetworkName(hre)).flatMap(([networkName, eid]) => (eid == null ? [] : [networkName]))
+          Object.entries(getEidsByNetworkName()).flatMap(([networkName, eid]) =>
+              eid != null && isOnStage(eid) ? [networkName] : []
```
lzBryan, line 53: would this be nicer if we used filter()?

janjakubnanista: You would need `filter` and `map` with an additional type declaration, i.e. `flatMap`:

```typescript
Object.entries(getEidsByNetworkName())
  .filter((value): value is [string, EndpointId] => eid != null && isOnStage(eid))
  .map(([networkName]) => networkName)
```

lzBryan (approved):

PR #612, opened 2024-05-13
🗞️ decode-lz-options

ryandgoulding:


packages/decode-lz-options/src/components/outputOptions.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const OutputOptions: React.FC<{ config: OutputOptionsType }> = ({
+  config,
+}) => {
+  return (
+    <Box flexDirection="column">
+      <Text>
+        <Text bold>Result:</Text>
+        {"\n"}
+        <Text color="green">{printJson(config)}</Text>
+      </Text>
```
janjakubnanista, line 26: Tiny thing but you can drop the wrapper `Text` and the `{"\n"}` and you should get the same result (maybe add a `gap={1}` on the `Box` to get an empty line.

janjakubnanista (dismissed):

packages/decode-lz-options/src/components/outputOptions.tsx
ryandgoulding: `gap` is not available;  I used `padding={1}` instead.

janjakubnanista (approved):

PR #615, opened 2024-05-22
🗞️ Allow expected reverts in verifyAndExecutePackets and fix forge coverage in TestHelper

DanL0:
- adds support for two vm.expectRevert signatures (bytes4 and bytes memory)
- renames verifyPackets to verifyAndExecutePackets for clarity
- adds tests to "@layerzerolabs/test-devtools-evm-foundry"
- use structs in TestHelper to make forge coverage working, fixes: https://github.com/LayerZero-Labs/LayerZero-v2/issues/42 (related PR: https://github.com/LayerZero-Labs/LayerZero-v2/pull/92)

Depends on:
- [ ] https://github.com/LayerZero-Labs/devtools/pull/791


St0rmBr3w:
Should we add a revert argument for checking the composed delivery?

TRileySchwarz:
This is also missing fixing the tests, or perhaps this is just PoC for now

DanL0:
seems vm.expectRevert Foundry doesn't behave like it's documented in Vm.sol, related issue: https://github.com/foundry-rs/foundry/issues/7629

Reported to Foundry team

packages/test-devtools-evm-foundry/contracts/TestHelperOz5.sol
```diff
@@ -310,30 +317,97 @@ contract TestHelperOz5 is Test, OptionsHelper {
...
+     * @param _expectedComposeRevertData4 Expected revert data for lzCompose in bytes4.
      */
-    function verifyPackets(uint32 _dstEid, bytes32 _dstAddress, uint256 _packetAmount, address _composer) public {
+    function verifyAndExecutePackets(
+        uint32 _dstEid,
+        bytes32 _dstAddress,
+        uint256 _packetAmount,
+        address _composer,
+        bytes memory _expectedReceiveRevertData,
+        bytes memory _expectedComposeRevertData,
```
TRileySchwarz, line 407: I dont know if I really agree on imbedding the expected revert data into the function like this, is there a functional reason why we dont just rely on the bubbled up revert and catch it with the vm.expectRevert inside of the test files implementation of this contract?

Otherwise it's unconventional compared to how you traditionally would read a testfile using vm.

DanL0: We could probably do that if we expose more granular functions from the TestHelper, eg. split verification and execution of packets. Right now vm.expectRevert only allows to catch revert in next function call or something and it doesnt catch reverts properly

TRileySchwarz:
We should split this PR into seperate smaller ones:

One for refactoring the endpoint setup to be more verbose. Excluding the "executeAndVerify" portion

DanL0:
> We should split this PR into seperate smaller ones:
> 
> One for refactoring the endpoint setup to be more verbose. Excluding the "executeAndVerify" portion

Agree. I'll do it when I have time or if anyone else has a time feel free to do it.

DanL0:
> We should split this PR into seperate smaller ones:
> 
> One for refactoring the endpoint setup to be more verbose. Excluding the "executeAndVerify" portion

https://github.com/LayerZero-Labs/devtools/pull/791

packages/test-devtools-evm-foundry/contracts/TestHelperOz5.sol
ryandgoulding: Agreed with @DanL0 , you can't just expectRevert() to be a catch-all for `verifyPackets()`

PR #619, opened 2024-05-29
🪲 Use `getNextNonce` when proposing transactions to safe

janjakubnanista:
### In this PR

- Add `nonce` coming from `getNextNonce` when proposing transactions to gnosis

### DIFF
--- a/packages/devtools-evm/src/signer/sdk.ts
+++ b/packages/devtools-evm/src/signer/sdk.ts
@@ -134,23 +134,20 @@ export class GnosisOmniSignerEVM<TSafeConfig extends ConnectSafeConfig> extends 
     }
 
     async #initSafe() {
-        if (this.safeConfig && (!this.safeSdk || !this.apiKit)) {
+        if (!this.safeSdk || !this.apiKit) {
             const ethAdapter = new EthersAdapter({
                 ethers,
                 signerOrProvider: this.signer,
             })
             this.apiKit = new SafeApiKit({ txServiceUrl: this.safeUrl, ethAdapter })
 
-            const contractNetworks = this.safeConfig.contractNetworks
             this.safeSdk = await Safe.create({
                 ethAdapter,
                 safeAddress: this.safeConfig.safeAddress!,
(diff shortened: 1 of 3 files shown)


lzBryan (approved):

packages/devtools-evm/src/signer/sdk.ts
```diff
@@ -130,23 +134,20 @@ export class GnosisOmniSignerEVM<TSafeConfig extends ConnectSafeConfig> extends
...
                 ethAdapter,
                 safeAddress: this.safeConfig.safeAddress!,
-                ...(!!contractNetworks && { contractNetworks }),
+                contractNetworks: this.safeConfig.contractNetworks,
             })
         }
-        if (!this.safeSdk || !this.apiKit) {
-            throw new Error('Safe SDK not initialized')
-        }
+
```
TRileySchwarz, line 150: log msg for already init?

TRileySchwarz: else case~

TRileySchwarz: might clog the debug if theres hundreds of tx though, so maybe not

TRileySchwarz:
This only addresses the nonce concept, not the multisender address for now correct?

packages/devtools-evm/src/signer/sdk.ts
janjakubnanista: that's gonna log on every transaction, this method is being called for every `signAndSend`

janjakubnanista:
> This only addresses the nonce concept, not the multisender address for now correct?

Oh that's already supported - we just need to set `contractNetworks` in `safeConfig` in hardhat config

packages/devtools-evm/src/signer/sdk.ts
TRileySchwarz: yeah realized that, nvm

TRileySchwarz (approved):

PR #632, opened 2024-06-12
🧹 Refactoring configureOAppPeers

ravinagill15:
## In this PR:
- Refactored `configureOAppPeers` in `ua-devtools > src > oapp > config.ts` to use `createConfigureEdges` pattern and added more logging

### DIFF
--- a/packages/ua-devtools/src/oapp/config.ts
+++ b/packages/ua-devtools/src/oapp/config.ts
@@ -64,6 +64,8 @@ export const configureOAppPeers: OAppConfigurator = withOAppLogger(
             async ({ vector: { from, to } }, sdk): Promise<OmniTransaction[]> => {
                 const logger = createOAppLogger()
 
+                logger.verbose(`Checking connection ${formatOmniVector({ from, to })}`)
+
                 const hasPeer = await sdk.hasPeer(to.eid, to.address)
 
                 logger.verbose(`Checked connection ${formatOmniVector({ from, to })}: ${printBoolean(hasPeer)}`)
@@ -75,37 +77,18 @@ export const configureOAppPeers: OAppConfigurator = withOAppLogger(
                 return [await sdk.setPeer(to.eid, to.address)]
             },
             {
-                onStart: (
-                    logger,
-                    [
-                        {
-                            vector: { from, to },
-                        },
-                    ]
-                ) => logger.verbose(`Checking connection for ${formatOmniVector({ from, to })}`),
-                onSuccess: (
-                    logger,
-                    [
-                        {
-                            vector: { from, to },
-                        },
-                    ]
-                ) => logger.verbose(`${printBoolean(true)} Created connection for ${formatOmniVector({ from, to })}`),
-                onError: (
-                    logger,
-                    [
-                        {
-                            vector: { from, to },
-                        },
-                    ],
-                    error
-                ) => logger.error(`Failed to create connection for ${formatOmniVector({ from, to })}: ${error}`),
+                onStart: (logger, [{ vector }]) =>
+                    logger.verbose(`Checking OApp peers for ${formatOmniVector(vector)}`),
+                onSuccess: (logger, [{ vector }]) =>
+                    logger.verbose(`${printBoolean(true)} Checked OApp peers for ${formatOmniVector(vector)}`),
+                onError: (logger, [{ vector }], error) =>
+                    logger.error(`Failed to check OApp peers for ${formatOmniVector(vector)}: ${error}`),
             }
         )
     ),
     {
         onStart: (logger) => logger.verbose(`Checking OApp peers configuration`),
-        onSuccess: (logger) => logger.verbose(`${printBoolean(true)} Created OApp peers`),
+        onSuccess: (logger) => logger.verbose(`${printBoolean(true)} Checked OApp peers configuration`),
     }
 )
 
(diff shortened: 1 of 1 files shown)


packages/ua-devtools/src/oapp/config.ts
```diff
@@ -9,6 +9,7 @@ import {
     formatOmniPoint,
     createConfigureMultiple,
     createConfigureNodes,
+    createConfigureEdges,
 } from '@layerzerolabs/devtools'
```
ravinagill15, line 13: **Local development issue**: When in the `ua-devtools` directory, this package (and a few others in the imports section) is unrecognized in this config file if I install `node_modules` with `pnpm i`. 

If I then delete `node_modules` in `ua-devtools` and use `npm i` to install, they are recognized. I run into this issue with `pnpm i` even after calling `nvm use`.

This _might_ be related to the issue resolved by [this PR](https://github.com/LayerZero-Labs/devtools/pull/631) where we needed to signal to the `Dockerfile` to use the `pnpm` version specified in the `package.json`...

```
ravina@Ravinas-MacBook-Pro ua-devtools % npm -v
9.5.1
ravina@Ravinas-MacBook-Pro ua-devtools % pnpm -v
8.14.0
```

packages/ua-devtools/src/oapp/config.ts
```diff
@@ -75,10 +73,41 @@ export const configureOAppPeers: OAppConfigurator = async (graph, createSdk) =>
...
+            },
+            {
+                onStart: (
+                    logger,
+                    [
+                        {
+                            vector: { from, to },
+                        },
+                    ]
+                ) => logger.verbose(`Checking connection for ${formatOmniVector({ from, to })}`),
```
ravinagill15, line 85: Where would these actually be printed? Is it in the terminal when wiring 2 contracts together using `npx hardhat lz:oapp:wire --oapp-config layerzero.config.ts` ? If so, I don't see any of these logs (even before I made my changes) -- perhaps there is a flag for a more verbose logging level similar to `forge`'s `-vvvv`?

packages/ua-devtools/src/oapp/config.ts
```diff
@@ -57,15 +58,12 @@ export const configureOAppDelegates: OAppConfigurator = withOAppLogger(
     }
 )
 
-export const configureOAppPeers: OAppConfigurator = async (graph, createSdk) => {
-    const logger = createModuleLogger('OApp')
-
-    return flattenTransactions(
```
ravinagill15, line 63: Unsure if I still need to `flattenTransactions` if using `createConfigureEdges`...will double check stargate repo to see how it done there

packages/ua-devtools/src/oapp/config.ts
```diff
@@ -75,10 +73,41 @@ export const configureOAppPeers: OAppConfigurator = async (graph, createSdk) =>
...
                 logger.verbose(`Creating a connection ${formatOmniVector({ from, to })}`)
                 return [await sdk.setPeer(to.eid, to.address)]
-            })
+            },
+            {
+                onStart: (
+                    logger,
+                    [
+                        {
+                            vector: { from, to },
```
janjakubnanista, line 82: Here you don't need to destructure and restructure - just do:

```typescript
onStart: (logger, [{ vector }]) => logger.verbose(`Checking connection ${formatOmniVector(vector)}`),
```

packages/ua-devtools/src/oapp/config.ts
janjakubnanista: The log levels for our hardhat tasks are specified using `--log-level` CLI flag so in your case:

```bash
npx hardhat lz:oapp:wire --oapp-config layerzero.config.ts --log-level verbose
```

If you do

```bash
npx hardhat lz:oapp:wire --help
```

it should give you all the possible values of `-log-level`

packages/ua-devtools/src/oapp/config.ts
```diff
@@ -75,10 +73,41 @@ export const configureOAppPeers: OAppConfigurator = async (graph, createSdk) =>
...
+                        },
+                    ]
+                ) => logger.verbose(`Checking connection for ${formatOmniVector({ from, to })}`),
+                onSuccess: (
+                    logger,
+                    [
+                        {
+                            vector: { from, to },
```
janjakubnanista, line 90: Same as above

packages/ua-devtools/src/oapp/config.ts
```diff
@@ -75,10 +73,41 @@ export const configureOAppPeers: OAppConfigurator = async (graph, createSdk) =>
...
+                        },
+                    ]
+                ) => logger.verbose(`${printBoolean(true)} Created connection for ${formatOmniVector({ from, to })}`),
+                onError: (
+                    logger,
+                    [
+                        {
+                            vector: { from, to },
```
janjakubnanista, line 98: Same as above

janjakubnanista (changes requested):

packages/ua-devtools/src/oapp/config.ts
ravinagill15: Makes sense !

packages/ua-devtools/src/oapp/config.ts
ravinagill15: Oo this is exactly what I was looking for, thank you!

janjakubnanista (approved):

janjakubnanista (approved):

PR #641, opened 2024-06-13
🧹 Update README.md

justOneLad:


janjakubnanista:
Although there is nothing wrong with the update, it also provides no value - as demonstrated by the fact that the `git revert` of this commit could be merged right after without any issues whatsoever.

PR #651, opened 2024-06-14
🧹 Allow users to specify whether the simulation accounts should be overwritten

janjakubnanista:
### In this PR

- To make #642 not affect the existing behavior, it has been adjusted with a flag that turns the account overwrites on/off
- The default value is `true` (which matches the previous behavior), setting it to an explicit `false` will turn the overwrites off


ravinagill15 (commented):
Should we also add a changeset? Other than that the changes LGTM 🚀

ravinagill15 (approved):
🚀

PR #660, opened 2024-06-21
👷➕ Added example solana OFT

St0rmBr3w:
Added example Solana OFT. Example scripts will not work without populating a valid private key / program IDs / accounts inputted.

@janjakubnanista feel free to modify as needed.

The workflow should be:

- Deploy the OFT Program ID using anchor and solana CLI
- Mint SPL Token and OFT Config
- Configure OFT Config
- Test transfer

Some To-Dos:

- convert users PRIVATE_KEY from bs58 and generate a new KeyPair file locally in their repo for deployment & configuration step.
- determine how to incorporate with existing npx framework (lz:deploy, lz:oapp:wire, etc.)
- Add OFT Adapter support (currently just NativeOFT)


TRileySchwarz:
We will split this up into smaller PR's, but can keep this for reference

PR #662, opened 2024-06-30
🧹onft remove OAppPreCrimeSimulator

ryandgoulding:
Do not support PreCrime in the default ONFT721 implementation.


ryandgoulding:
rebased

ryandgoulding:
The design choice was made to keep this.  Can always be revived if we change our mind.

PR #675, opened 2024-07-03
🧹 Add devtools configurator for OFTWrapper.setCallerBpsCap(...)

ravinagill15:
## In this PR:
- Adds getter and setter for `OFTWrapper.callerBpsCap` in `packages/ua-devtools-evm/src/oapp/sdk.ts`
- Calling above getter/setter in configurator
- Also have an example of how `callerBpsCap` can be configured in `layerzero.config.ts`

### DIFF
--- a/examples/oapp/layerzero.config.ts
+++ b/examples/oapp/layerzero.config.ts
@@ -21,9 +21,14 @@ const config: OAppOmniGraphHardhat = {
     contracts: [
         {
             contract: fujiContract,
-            config: {
-                callerBpsCap: BigInt(300),
-            },
+            /**
+             * This config object is optional.
+             * The callerBpsCap refers to the maximum fee (in basis points) that the contract can charge.
+             */
+
+            // config: {
+            //     callerBpsCap: BigInt(300),
+            // },
         },
         {
             contract: sepoliaContract,
--- a/examples/oapp/layerzero.config.ts
+++ b/examples/oapp/layerzero.config.ts
@@ -22,7 +22,7 @@ const config: OAppOmniGraphHardhat = {
         {
             contract: fujiContract,
             config: {
-                callerBpsCap: BigInt(1000),
+                callerBpsCap: BigInt(300),
             },
         },
         {
--- a/examples/oapp/layerzero.config.ts
+++ b/examples/oapp/layerzero.config.ts
@@ -22,8 +22,7 @@ const config: OAppOmniGraphHardhat = {
         {
             contract: fujiContract,
             config: {
-                callerBpsCap: 1000, // this is example of how callerBpsCap can be configured when calling lz:oapp:wire
-                // QUESTION: is this accurate + should this be in edges or nodes?
+                callerBpsCap: BigInt(1000),
             },
         },
         {
--- a/examples/oapp/layerzero.config.ts
+++ b/examples/oapp/layerzero.config.ts
@@ -21,6 +21,10 @@ const config: OAppOmniGraphHardhat = {
     contracts: [
         {
             contract: fujiContract,
+            config: {
+                callerBpsCap: 1000, // this is example of how callerBpsCap can be configured when calling lz:oapp:wire
+                // QUESTION: is this accurate + should this be in edges or nodes?
+            },
         },
         {
             contract: sepoliaContract,
--- a/packages/ua-devtools-evm/src/oapp/sdk.ts
+++ b/packages/ua-devtools-evm/src/oapp/sdk.ts
@@ -139,7 +139,7 @@ export class OApp extends Ownable implements IOApp {
     }
 
     async setCallerBpsCap(callerBpsCap: bigint): Promise<OmniTransaction | undefined> {
-        if (this.contract.contract.interface.functions['setCallerBpsCap'] == null) {
+        if (this.contract.contract.interface.functions['setCallerBpsCap(uint256)'] == null) {
             return (
                 this.logger.warn(
                     `Cannot set callerBpsCap for ${this.label}: setCallerBpsCap function is not supported`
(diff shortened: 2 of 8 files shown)


examples/oapp/layerzero.config.ts
```diff
@@ -21,6 +21,10 @@ const config: OAppOmniGraphHardhat = {
     contracts: [
         {
             contract: fujiContract,
+            config: {
+                callerBpsCap: 1000, // this is example of how callerBpsCap can be configured when calling lz:oapp:wire
+                // QUESTION: is this accurate + should this be in edges or nodes?
+            },
```
ravinagill15, line 27: I will push tests in the next commit, but want to confirm if `callerBpsCap` should be specified at the node level or the edge level in the `connections` further down in this file. I placed it at the node level because the [pattern for configureOftBps](https://github.com/stargate-protocol/stargate-v2/blob/825d447ead52725cc128751db8d58cf823899865/packages/stg-evm-v2/devtools/src/oft-wrapper/config.ts#L16) configures it at the node level

ravinagill15:
Looking into test failures!

packages/ua-devtools-evm/src/oapp/sdk.ts
```diff
@@ -138,6 +138,22 @@ export class OApp extends Ownable implements IOApp {
...
+            description: `Setting caller BPS cap to ${callerBpsCap}`,
+        }
+    }
+
+    @AsyncRetriable()
+    async getCallerBpsCap(): Promise<bigint> {
+        const callerBpsCap = await this.contract.contract.callerBpsCap()
+
+        return BigNumberishBigIntSchema.parse(callerBpsCap)
+    }
```
ravinagill15, line 155: As mentioned in another comment on this PR, not all contracts will have these functions (only `OFTWrapper` or its children will) so perhaps we should wrap the logic within these functions in a try-catch?

For testing purposes in this PR, I added these getters/setters to `DefaultOApp`

tests/ua-devtools-evm-hardhat-test/contracts/DefaultOApp.sol
```diff
@@ -6,7 +6,17 @@ import { OApp, Origin } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/OApp.
 import { OAppOptionsType3 } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/libs/OAppOptionsType3.sol";
 
 contract DefaultOApp is OApp, OAppOptionsType3 {
+    uint256 public callerBpsCap; // TODO can move to a different contract for testing
```
ravinagill15, line 9: Is this worth doing or should I leave it within `DefaultOApp.sol`?

ravinagill15:
Added `do not merge` label until above questions are addressed and I have cleaned up the comments

packages/ua-devtools-evm/src/oapp/sdk.ts
janjakubnanista: I think we have an example of handling this is stargate v2 - in asset sdk i believe, look for lpToken getter

tests/ua-devtools-evm-hardhat-test/contracts/DefaultOApp.sol
```diff
@@ -6,7 +6,17 @@ import { OApp, Origin } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/OApp.
...
     constructor(address _endpoint, address _delegate) OApp(_endpoint, _delegate) Ownable(_delegate) {}
 
     function _lzReceive(Origin calldata, bytes32, bytes calldata, address, bytes calldata) internal virtual override {}
+
+    /**
+        Copied from https://github.com/stargate-protocol/stargate-v2/blob/3556e333b197b4f3706c68aa2b5cb4060c9c3812/packages/stg-evm-v2/src/peripheral/oft-wrapper/OFTWrapper.sol#L38-L42
+        (except validations and event emitting)
+    */
+    function setCallerBpsCap(uint256 _callerBpsCap) external onlyOwner {
+        callerBpsCap = _callerBpsCap;
```
janjakubnanista, line 20: Always good to emit events in setters, maybe emit one here

janjakubnanista (changes requested):

packages/ua-devtools-evm/src/oapp/sdk.ts
ravinagill15: ah gotcha, this is exactly what I was looking for thanks!

examples/oapp/layerzero.config.ts
```diff
@@ -21,6 +21,9 @@ const config: OAppOmniGraphHardhat = {
     contracts: [
         {
             contract: fujiContract,
+            config: {
+                callerBpsCap: BigInt(1000),
+            },
```
ravinagill15, line 26: Should callerBpsCap be specified at the node level or the edge level in the connections further down in layerzero.config.ts? I placed it at the node level because the [pattern for configureOftBps](https://github.com/stargate-protocol/stargate-v2/blob/825d447ead52725cc128751db8d58cf823899865/packages/stg-evm-v2/devtools/src/oft-wrapper/config.ts#L16) configures it at the node level

ravinagill15:
One of the tests is a false positive -- looking into it now!

examples/oapp/layerzero.config.ts
ravinagill15: Tested this out by deploying and wiring real contracts, then fetching the value of `callerBpsCap` to ensure it matches the value specified in the `layerzero.config.ts` -- works as expected

examples/oapp/layerzero.config.ts
```diff
@@ -21,6 +21,9 @@ const config: OAppOmniGraphHardhat = {
     contracts: [
         {
             contract: fujiContract,
+            config: {
+                callerBpsCap: BigInt(1000),
```
ryandgoulding, line 25: Per discussions w/ Riley, I would recommend you set a default of `300`.

ravinagill15: Updated!

janjakubnanista (approved):

PR #676, opened 2024-07-09
🗞️ move OFT into devtools

ryandgoulding:


packages/oft-evm/artifacts/contracts/OFT.sol/OFT.dbg.json
```diff
@@ -0,0 +1,4 @@
+{
```
janjakubnanista, line 1: I think these dbg files should be ignored, the rest of the artifacts is fine to keep BUT what we need to do to stay consistent & compatible with devtools is to move `artifacts/contracts` to just `artifacts` and export that folder in `package.json`.

There are two ways I can think of:

- Update `hardhat.config.ts` and change the artifacts folder in the `paths` config to something else (let's say `fartifacts`), then on build copy `fartifacts/contracts` to `artifacts`
- Get rid of hardhat completely and do the same copying but form the `out` folder of foundry

The copying can be done either by low-level (but system-specific) `rm -rf && cp -R` or then by adding a build step and a tsup config, both have pros and cons

packages/oft-evm/contracts/OFTPrecrime.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
```
janjakubnanista, line 3: Do we want this one if it's all commented out?

packages/oft-evm/hardhat.config.ts
```diff
@@ -0,0 +1,96 @@
+// Get the environment configuration from .env file
+//
+// To make use of automatic environment setup:
+// - Duplicate .env.example file and name it .env
+// - Fill in the environment variables
```
janjakubnanista, line 5: I guess all of this and the `dotenv` should be removed

packages/oft-evm/test/mocks/SendUln302Mock.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
```
janjakubnanista, line 2: I think this one we already have in `packages/test-devtools-evm-foundry/contracts/mocks/SendUln302Mock.sol` so you can add an NPM dependency and import

packages/oft-evm/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "@layerzerolabs/typescript-config-next/library.json",
+
+  "files": ["hardhat.config.ts"], // this root tsconfig just exists to compose sub-projects
```
janjakubnanista, line 4: Comments are not valid JSON

ryandgoulding: what i get for using COMMAND+C

packages/oft-evm/test/mocks/SendUln302Mock.sol
ryandgoulding: Wish I could;  that one only supports OpenZeppelin v5.  OFT uses OpenZeppelin V4.

packages/oft-evm/artifacts/contracts/OFT.sol/OFT.dbg.json
ryandgoulding: Not sure I follow.  What about just eliminating hardhat, then changing foundry.toml so `out=artifacts`?

janjakubnanista: Yeah we can do that too. The trick for that solution is not to pack all the artifacts with the package since foundry will include all the test contracts there too:

Screenshot 2024-07-10 at 1 51 14 PM

So you need to handpick the artifacts you want included in `package.json` (in this case it should be just the `OFT` though)

ryandgoulding: done.

packages/oft-evm/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "@layerzerolabs/oft-evm",
+  "version": "0.0.1",
+  "license": "MIT",
+  "exports": {
```
janjakubnanista, line 5: Oh and if you could pls add the `repository` and `description` fields (even `keywords` if you're feeling adventurous)

TRileySchwarz (approved):

packages/oft-evm/README.md
```diff
@@ -0,0 +1 @@
+# OFT EVM
```
janjakubnanista, line 1: Could you copy the standard README from one of the other packages? Consistency is key :D

janjakubnanista (changes requested):
Just that README change and is ready, thanks!

janjakubnanista (approved):

PR #677, opened 2024-07-09
🧹 Add a test for an external contract in connections

janjakubnanista:
### In this PR

- Add a test case for the wire task that involves an external contract set as a `to` for a connection


ravinagill15 (approved):
LGTM 🚀

PR #681, opened 2024-07-09
🪲 Fix problem with missing forge tests in OApp example; Fix cache paths for examples

janjakubnanista:
### In this PR

- Add a dummy forge test for OApp example
- Fix cache paths for examples so that hardhat and forge don't overwrite each other's cache

### DIFF
--- a/examples/oapp/test/foundry/MyOApp.t.sol
+++ b/examples/oapp/test/foundry/MyOApp.t.sol
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+// MyOApp imports
+import { MyOApp } from "../../contracts/MyOApp.sol";
+
+// OApp imports
+import { IOAppOptionsType3, EnforcedOptionParam } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/libs/OAppOptionsType3.sol";
+import { OptionsBuilder } from "@layerzerolabs/lz-evm-oapp-v2/contracts/oapp/libs/OptionsBuilder.sol";
+
+// OZ imports
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+
+// Forge imports
+import "forge-std/console.sol";
+
+// DevTools imports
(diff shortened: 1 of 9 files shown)


TRileySchwarz (approved):

examples/oapp/test/foundry/MyOApp.t.sol
```diff
@@ -0,0 +1,57 @@
...
+// Forge imports
+import "forge-std/console.sol";
+
+// DevTools imports
+import { TestHelperOz5 } from "@layerzerolabs/test-devtools-evm-foundry/contracts/TestHelperOz5.sol";
+
+contract MyOAppTest is TestHelperOz5 {
+    using OptionsBuilder for bytes;
+
+    uint32 aEid = 1;
```
ryandgoulding, line 23: NIT:  add an access modifier (i.e., `internal`).

TRileySchwarz (approved):

examples/onft721/package.json
```diff
@@ -11,7 +11,7 @@
     "lint:fix": "eslint --fix '**/*.{js,ts,json}' && prettier --write . && solhint 'contracts/**/*.sol' --fix --noPrompt",
     "lint:js": "eslint '**/*.{js,ts,json}' && prettier --check .",
     "lint:sol": "solhint 'contracts/**/*.sol'",
-    "test": "$npm_execpath run test:forge && $npm_execpath run test:hardhat",
+    "test": "echo 'No tests yet'",
```
ryandgoulding, line 14: yeah i saw this fail locally... was the cache causing CI to pass?

ryandgoulding (approved):

examples/onft721/package.json
janjakubnanista: yeah the last time we updated the base image was 8 or so months ago and apparently since then `forge test` started failing if it didn't find any tests

examples/oapp/test/foundry/MyOApp.t.sol
janjakubnanista: I am just gonna merge this one as is if that's okay - it's blocking a bunch of PRs from getting through and I don't wanna be waiting for another pipeline to go through. Will fix in the next one, I still want to drop the `turbo.json` files we used to work around the `solc` `text file busy`

PR #701, opened 2024-07-16
🗞️ Fee interface and implementation

ryandgoulding:


packages/oft-evm/contracts/OFTFee.sol
```diff
@@ -0,0 +1,113 @@
...
+     * @return amountReceivedLD The amount received in local decimals on the remote.
+     */
+    function _debit(
+        address _from,
+        uint256 _amountLD,
+        uint256 _minAmountLD,
+        uint32 _dstEid
+    ) internal virtual override returns (uint256 amountSentLD, uint256 amountReceivedLD) {
+        (amountSentLD, amountReceivedLD) = _debitView(_amountLD, _minAmountLD, _dstEid);
+        _transferFrom(_from, address(this), amountSentLD - amountReceivedLD);
```
TRileySchwarz, line 56: Should do this conditionally, in case there isnt fees applied, ie. ensure > 0

packages/oft-evm/contracts/OFTFee.sol
```diff
@@ -0,0 +1,113 @@
...
+        address _delegate
+    ) OFT(_name, _symbol, _lzEndpoint, _delegate) {}
+
+    /**
+     * @notice Indicates whether the OFT contract requires approval of the 'token()' to send.
+     * @return requiresApproval Needs approval of the underlying token implementation.
+     *
+     * @dev In the case of OFT where the contract IS the token, approval is NOT required.
+     */
+    function approvalRequired() external pure virtual override returns (bool) {
```
TRileySchwarz, line 36: Shouldnt need this, should come with the oft inheretence?

packages/oft-evm/contracts/OFTFee.sol
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * @dev Credits tokens to the specified address.
+     * @param _to The address to credit the tokens to.
+     * @param _amountLD The amount of tokens to credit in local decimals.
+     * @dev _srcEid The source chain ID.
+     * @return amountReceivedLD The amount of tokens ACTUALLY received in local decimals.
+     */
+    function _credit(
```
TRileySchwarz, line 67: Dont think we need credit overriden?

packages/oft-evm/contracts/OFTFee.sol
```diff
@@ -0,0 +1,113 @@
...
+            revert SlippageExceeded(amountReceivedLD, _minAmountLD);
+        }
+        amountSentLD = (amountReceivedLD * BPS_DENOMINATOR) / (BPS_DENOMINATOR - feeBps);
+    }
+
+    function _feeBalance() internal virtual override returns (uint256) {
+        return IERC20(this).balanceOf(address(this));
+    }
+
+    function _transferFrom(address _from, address _to, uint256 _amount) internal virtual override returns (uint256) {
```
TRileySchwarz, line 107: There is no transfer from, so this helper is unnecessary

packages/oft-evm/contracts/OFTFee.sol
TRileySchwarz, line 56: This should be a setter, which defaults to address(this), but can be overriden by the owner to send direct to their fee colelctor

ryandgoulding:
Giving thought to how we assess fees, do not merge for now.

packages/oft-evm/contracts/OFTFee.sol
ryandgoulding: I disagree;  let me explain my thinking.  This implementation would lead to two different usage patterns for distributing fees:

1) Collect and lock fees in `address(this)`.  In the case of `OFTAdapter`, this obviously comingles fees and locked assets.  In this case `OFTFeeAdapter.feeBalance` is used to tally fees.  Later, the owner may `withdrawFees()`.
2) Provide a hook to set a receiver, and immediately distribute fees to that receiver.

I would argue for just the former.  My reason is that, especially with the case of AdapterOFT, funds are just moved once so it is cheaper (i.e., only locked in the OFTFeeAdapter contract), and fees are distinguished from locked assets using the `feeBalance`.  For the latter, we would need to implement complicated logic to check who the receiver is, and only add fees to `OFTFeeAdapter.feeBalance` if the receiver is not the contract itself (i.e., the funds never accrue because they are always immediately distributed).

packages/oft-evm/contracts/OFTFee.sol
ryandgoulding: I believe it is necessary.  `Fee` is abstract and could handle anything from ERC-20 tokens to native tokens (in fact, I use native in the unit tests as it is easier to implement).  As such, how the funds are moved is implementation dependent.  We are not providing a `token` argument to `Fee` constructor, so it has no knowledge of what it is actually collecting.  It just knows that it is collecting something, and eventually someone is going to `withdrawFees()`.  It has no idea how to perform `withdrawFees()`, it just knows that the child class will provide the appropriate functionality to do so.

Additionally, `OFTFeeAdapter` can reuse `_transferFrom()` for both `debit()` and `withdrawFees()`;  the steps are identical.

I am not against simplifying this, but I would argue that the status quo is not overly complicated either.

packages/oft-evm/contracts/OFTFee.sol
TRileySchwarz: Yes, but it is msg sender, thus they will always be being called from an approved spender, and is guarded via the slippage input

packages/oft-evm/contracts/OFTFee.sol
TRileySchwarz: Yeah I understand the point, let me take another look and see if theres a way to do this differently

packages/oft-evm/contracts/OFTFeeAdapter.sol
```diff
@@ -0,0 +1,77 @@
...
+
+abstract contract OFTFeeAdapter is OFTAdapter, Fee {
+    using SafeERC20 for IERC20;
+
+    uint256 public feeBalance;
+
+    constructor(address _token, address _lzEndpoint, address _delegate) OFTAdapter(_token, _lzEndpoint, _delegate) {}
+
+    /**
+     * @dev Burns tokens from the sender's specified balance, ie. pull method.
```
TRileySchwarz, line 18: This doesnt burn tokens, it locks them

packages/oft-evm/contracts/OFTFee.sol
ryandgoulding: This was done because if approval is not granted (i.e., https://github.com/LayerZero-Labs/devtools/blob/feat/oft-with-fee/packages/oft-evm/test/OFTFee.t.sol#L282) then it will revert:

```
revert: ERC20: insufficient allowance
```

TRileySchwarz (approved):

PR #703, opened 2024-07-17
🧹 Use network setup utility in @layerzerolabs/export-deployments-test

janjakubnanista:
### In this PR

- Small fix found while working on #700 - using the network setup utility in one of the test packages


ravinagill15 (commented):
Looks good but I think we need a changeset?

janjakubnanista:
> Looks good but I think we need a changeset?

Oh no, this is not a published package, this is just a test package.

ravinagill15 (approved):

PR #715, opened 2024-07-17
Riley/oft fee refactor

TRileySchwarz:


packages/oft-evm/contracts/OFTFeeAdapter.sol
```diff
@@ -12,59 +12,57 @@ abstract contract OFTFeeAdapter is OFTAdapter, Fee {
...
-        innerToken.safeTransferFrom(_from, address(this), amountSentLD);
-        feeBalance += (amountSentLD - amountReceivedLD);
+    constructor(
+        address _token,
+        address _lzEndpoint,
+        address _owner
+    ) OFTAdapter(_token, _lzEndpoint, _owner) {}
+
+    // @dev Fees accumulate inside of the contract to save gas, and then can be withdrawn by the owner.
+    function withdrawFees(address _to) external onlyOwner {
```
ryandgoulding, line 25: Should we allow the ability to whitelist a non-owner to withdrawFees here?  I.e., `onlyWithdrawoooooor`?  Just seems like a multisig will own this, which is painful to use to withdraw fees.

ryandgoulding (approved):

PR #717, opened 2024-07-17
🧹 Don't force a network to be defined in hardhat config when formatting transactions

janjakubnanista:
### In this PR

- When using `formatOmniTransaction` with a transaction whose network is not defined in `hardhat.config`, the function would throw. This is not great if we want to add e.g. Solana to our `layerzero.config`

### DIFF
--- a/packages/devtools-evm-hardhat/src/runtime.ts
+++ b/packages/devtools-evm-hardhat/src/runtime.ts
@@ -198,6 +198,34 @@ export const getEidForNetworkName = (
  * Gets a network name with its `eid` property matching
  * a particular `eid`
  *
+ * Throws if there are multiple networks defined with the same `eid`.
+ *
+ * Returns `undefined` if there is no network with given `eid`
+ *
+ * @param {EndpointId} eid
+ * @param {HardhatRuntimeEnvironment | undefined} [hre]
+ * @returns {string | undefined}
+ */
+export const getNetworkNameForEidMaybe = (
+    eid: EndpointId,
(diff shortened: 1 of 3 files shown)


packages/devtools-evm-hardhat/src/runtime.ts
```diff
@@ -218,8 +219,28 @@ export const getNetworkNameForEid = (
...
+ */
+export const getNetworkNameForEid = (
+    eid: EndpointId,
+    hre: HardhatRuntimeEnvironment = getDefaultRuntimeEnvironment()
+): string => {
+    const networkName = getNetworkNameForEidMaybe(eid, hre)
+
     // Here we error out if there are no networks with this eid
-    assert(false, `Could not find a network for eid ${eid} (${formatEid(eid)})`)
+    return assert(networkName != null, `Could not find a network for eid ${eid} (${formatEid(eid)})`), networkName
```
ravinagill15, line 243: hmm would network name ever be `null`? Seems like it would only ever be `string | undefined`

ravinagill15 (approved):
LGTM 🚀

packages/devtools-evm-hardhat/src/runtime.ts
janjakubnanista: `a == null` / `a != null` is a very safe way of checking whether a value is nullish (i.e. `null` or `undefined`) - it weeds out all the values that would produce problems if accessed using a `.` operator (i.e. `null.length`) and also works for things like `0` or `false`

PR #718, opened 2024-07-18
🧹 use @layerzerolabs/oft-evm for OFT instead of @layerzerolabs/lz-evm-oapp-v2

ryandgoulding:


packages/oft-evm/package.json
```diff
@@ -76,8 +76,5 @@
     "@openzeppelin/contracts-upgradeable": "^4.8.1 || ^5.0.0",
     "hardhat-deploy": "^0.11.44",
     "solidity-bytes-utils": "^0.8.0"
-  },
-  "publishConfig": {
-    "access": "private"
```
janjakubnanista, line 81: Generally I would prefer a two step process - first make this package public and publish, only then add it to the example because there will be a period of time when the example will be live in the repo but the package will not be public.

Keep it as is for now just publish as soon as possible pls

janjakubnanista (approved):

PR #728, opened 2024-07-18
🗞️ Update urls for testnets

TRileySchwarz:
Add support for default scan verify urls:
arbitrum sepolia
optimism sepolia
sepolia


ryandgoulding:
Probably affected by the huge outages.  I would recommend a small change and force pushing to re-run the jobs...

janjakubnanista (approved):

PR #745, opened 2024-07-23
🪲  allow feeBps == BPS_DENOMINATOR

ryandgoulding:
Although not a common setting, it could be useful to allow 100% fees.


janjakubnanista (changes requested):
Just a changeset pls

janjakubnanista (approved):

PR #756, opened 2024-07-24
🗞️ Solana: Enforced options on OFT SDK [9/N]

janjakubnanista:
### In this PR

- Adding getters & setters for enforced options on Solana OFT SDK


packages/ua-devtools-solana/src/oft/sdk.ts
```diff
@@ -124,17 +127,110 @@ export class OFT extends OmniSDK implements IOApp {
...
     async getEnforcedOptions(eid: EndpointId, msgType: number): Promise<Bytes> {
+        // First we check that we can understand the message type
+        this.assertMsgType(msgType)
+
         const eidLabel = `eid ${eid} (${formatEid(eid)})`
+        this.logger.verbose(`Getting enforced options for ${eidLabel} and message type ${msgType}`)
+
+        try {
+            const options = await OftTools.getEnforcedOptions(this.connection, this.configAccount, eid, this.publicKey)
+            const optionsForMsgType = msgType === MSG_TYPE_SEND ? options.send : options.sendAndCall
```
ryandgoulding, line 138: I know you know this, but just wanted to point out that if we ever add further message types in the default implementation, we will need to update this.

packages/ua-devtools-solana/src/oft/sdk.ts
```diff
@@ -124,17 +127,110 @@ export class OFT extends OmniSDK implements IOApp {
...
 
-        this.logger.debug(`Getting enforced options for ${eidLabel} and message type ${msgType}`)
+            return toHex(optionsForMsgType)
+        } catch (error) {
+            if (String(error).match(/Unable to find EnforcedOptions account/)) {
```
ryandgoulding, line 142: Not your fault as it seems this comes from elsewhere, but having to rely on error messages from an API seems like a recipe for disaster.  Eventually, we should aim for a better API in OftTools.

packages/ua-devtools-solana/src/oft/sdk.ts
janjakubnanista: I know, this just works with what I have from the monorepo and that function only allows these two, hence these checks

packages/ua-devtools-solana/src/oft/sdk.ts
janjakubnanista: I wouls love to scrape OftTools altogether, it's something we can completely move to devtools and save ourselves all this adapter logic

ryandgoulding (approved):

PR #811, opened 2024-08-02
🗞️ Solana: Add account initialization checkers for OFT [37/N]

janjakubnanista:
### In this PR

- Add account initialization status checkers to `OFT` and `EndpointV2` SDK. These are used in the initialization configurator to check whether we need to initialize `OApp` nonce, send & receive lib or config accounts

### DIFF
--- a/packages/devtools-solana/test/common/accounts.test.ts
+++ b/packages/devtools-solana/test/common/accounts.test.ts
@@ -1,0 +1,61 @@
+import { createGetAccountInfo } from '@/common/accounts'
+import { ConnectionFactory, createConnectionFactory, createRpcUrlFactory } from '@/connection'
+import { EndpointId } from '@layerzerolabs/lz-definitions'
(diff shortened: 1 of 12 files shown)


packages/devtools-solana/test/common/accounts.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    afterEach(() => {
+        getAccountInfoMock.mockRestore()
+    })
+
+    describe('createGetAccountInfo', () => {
+        it('should return account info if the account has been initialized', async () => {
+            const connection = await connectionFactory(oftConfig.eid)
+            const getAccountInfo = createGetAccountInfo(connection)
+
+            expect(await getAccountInfo(oftConfig.address)).not.toBeUndefined()
```
ravinagill15, line 41: Should we also validate the structure of the account info here?

ravinagill15 (approved):
LGTM 🚀

PR #818, opened 2024-08-07
🗞️ Solana: Add OFT example

janjakubnanista:
### In this PR

- Adding a new example, `oft-solana`
- Adding the example into the example selector in `create-lz-oapp`. For now it only appears if `LZ_ENABLE_EXPERIMENTAL_SOLANA_OFT_EXAMPLE` environment variable is set to a truthy value


examples/oft-solana/Anchor.toml
```diff
@@ -0,0 +1,19 @@
+[toolchain]
+anchor_version = "0.29.0"
+
+[features]
+seeds = false
+skip-lint = false
+
+[programs.localnet]
+oft = "CUinP3FqxxPGxhnmTSMDJM9rnKW6urhDugecXV13eZdK"
```
ryandgoulding, line 9: If I am not mistaken, don't we want folks to deploy their own program with their own account?

examples/oft-solana/Anchor.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[registry]
+url = "https://api.apr.dev"
+
+[provider]
+cluster = "Localnet"
+wallet = "./junk-id.json"
```
ryandgoulding, line 16: nice...

examples/oft-solana/junk-id.json
```diff
@@ -0,0 +1,6 @@
+[
```
ryandgoulding, line 1: Should we really be embedding this and not just generating it later?

examples/oft-solana/oft.jpg
ryandgoulding, line 1: why jpg and not svg or something with better compression?

examples/oft-solana/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "clean": "rm -rf .anchor target",
+    "compile": "anchor build",
+    "test": "anchor test"
+  },
+  "devDependencies": {
+    "@coral-xyz/anchor": "^0.29.0",
+    "@layerzerolabs/lz-solana-sdk-v2": "^2.3.31",
```
ryandgoulding, line 12: Can we use latest version (2.3.34)?

examples/oft-solana/programs/endpoint-mock/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
```
ryandgoulding, line 1: Why is the endpoint-mock bundled with the example code?  Wouldn't it make sense to make it a separate package so as to not bloat the example code?

examples/oft-solana/programs/oft/src/instructions/init_adapter_oft.rs
```diff
@@ -0,0 +1,66 @@
+use crate::*;
+use anchor_spl::token_interface::{Mint, TokenAccount, TokenInterface};
+
+#[derive(Accounts)]
+#[instruction(params: InitAdapterOftParams)]
+pub struct InitAdapterOft<'info> {
```
ryandgoulding, line 6: So you get the adapter and the oft all in one package?  In EVM, we do pure OFT or Adapter as two separate options.  I don't really care, just curious.

examples/oft-solana/programs/oft/src/instructions/lz_receive.rs
```diff
@@ -0,0 +1,167 @@
+use crate::*;
```
ryandgoulding, line 1: Again, and maybe I am missing something fundamental with solana, but wouldn't we want to put the source of the oft in `packages/` not `examples/`?

examples/oft-solana/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "types": ["jest"],
+    "lib": ["es2015"],
```
ryandgoulding, line 4: very old, why?

examples/oft-solana/Anchor.toml
St0rmBr3w: Solana is weird and so you create keypairs ahead of time and then deploy to them.

You can run `anchor keys sync` to sync to your keypairs for your wallet.

St0rmBr3w: > Each Anchor project has it's own program address when it's first built. Since you've just copied my code, the program address at the top of your lib.rs won't match yours. To generate a keypair, run anchor build, and then run anchor keys sync to set it in your Anchor.toml and in the id field of your lib.rs.

https://solana.com/developers/guides/getstarted/full-stack-solana-development

examples/oft-solana/programs/endpoint-mock/Cargo.toml
St0rmBr3w: You need the RegisterOApp method I believe, which is why it's included here.

examples/oft-solana/programs/oft/src/instructions/init_adapter_oft.rs
St0rmBr3w: You need both in one package because you mint accounts off the OFT Program, rather than deploying one or the other.

The OFT Program comes set with:

- OFT (NativeOFT)
- OFT Adapter
- Rate Limits
- Minting

to name a few.

examples/oft-solana/programs/oft/src/instructions/lz_receive.rs
St0rmBr3w: You need to deploy an OFT Program and mint accounts off of it, so you need to compile the anchor code and then deploy it directly via anchor.

After you've deployed your OFT Program, you can then mint the necessary accounts to interact with the OFT Program:

```
"accounts": [
    {
      "name": "SPL Token Mint Account",
      "publicKey": "Eh6XEPhSwoLv5wFApukmnaVSHQ6sAnoD9BmgmwQoN2sN"
    },
    {
      "name": "OFT Config Account",
      "publicKey": "D27Yp9iR2UAgD8BkBbWrHeK6MooSQu353NJm3SVmEvXw"
    },
    {
      "name": "OFT Program ID",
      "publicKey": "CKsfN2UvdF6BxWmt12576GN4EtLynA7MxSqEm6jrCqMh"
    }
  ]
```

examples/oft-solana/programs/endpoint-mock/Cargo.toml
mok-lz: Yes, this is used in unit testing, and I haven’t found a better way to handle this issue yet.

examples/oft-solana/Anchor.toml
janjakubnanista: Yeah so this is one of the tricky parts when it comes to having this project in a CI/CD environment. It goes something like this.

There are several references to program IDs (in `lib.rs` and `Anchor.toml`). They depend on keys generated, in our case, for `endpoint` and `oft`. These keys live in `target/deploy`.

If we ignore `target/deploy` (or if you delete it for that matter), next time you run e.g. `anchor test`, new keys will be generated. Tests will fail because the program IDs depend on these keys so you need to run `anchor keys sync`.

What I decided to do is to include three junk keys in the example - `junk-id.json`, `target/deploy/endpoint-keypair.json` and `target/deploy/oft-keypair.json`. This way if you check out devtools and run test, tests will pass.

The other option was way more annoying - generate new keys in `prepare` script and basically having a diff everytime you're making a PR.

Side note: I will also add these junk keys to the list of files to be deleted when you use `create-lz-oapp` so that users are not confused.

examples/oft-solana/programs/endpoint-mock/Cargo.toml
janjakubnanista: At this point, we have no process to publish crates but yeah it would be nice to have it packaged.

Possible solution is to move it to `packages` directory and have cargo install it using `git`, I'll investigate

examples/oft-solana/programs/oft/src/instructions/lz_receive.rs
janjakubnanista: The same as above @ryandgoulding - I will see if cargo could install from `git` using `path`.

examples/oft-solana/tsconfig.json
janjakubnanista: It's pretty much standard and more than enough, we're not using any features from `ES7` or above

examples/oft-solana/package.json
janjakubnanista: Done

examples/oft-solana/oft.jpg
janjakubnanista: ask my man @St0rmBr3w but I believe he pasted what he got

ryandgoulding (approved):

examples/oft-solana/oft.jpg
St0rmBr3w: We can replace, this is just the README hk shipped with, it will be dead soon @ryandgoulding

PR #833, opened 2024-08-09
🗞️ Support encoded constructor arguments in non-target verification

janjakubnanista:
### In this PR

- Constructor arguments (`--arguments`) can now be passed as a HEX-encoded string (with leading `0x`) to the `non-target` verification CLI


ravinagill15 (approved):
This was much needed! 🚀

PR #838, opened 2024-08-13
🗞️ Native OFT Adapter

ravinagill15:
## In this PR
- Implemented `NativeOFTAdapter` with tests

### DIFF
--- a/.changeset/shaggy-actors-call.md
+++ b/.changeset/shaggy-actors-call.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/oft-evm": minor
+---
+
+Added NativeOFTAdapter implementation
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -16,7 +16,7 @@ import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "
  */
 abstract contract NativeOFTAdapter is OFTCore {
 
-    error InsufficientMessageValue(uint256 provided, uint256 required);
+    error IncorrectMessageValue(uint256 provided, uint256 required);
     error CreditFailed();
 
     /**
@@ -68,11 +68,11 @@ abstract contract NativeOFTAdapter is OFTCore {
         MessagingFee calldata _fee,
         address _refundAddress
     ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
-        // @dev Ensure the native funds in msg.value are enough to cover the fees and amount to send (with dust removed).
+        // @dev Ensure the native funds in msg.value are exactly enough to cover the fees and amount to send (with dust removed).
         // @dev This will revert if the _sendParam.amountLD contains any dust
         uint256 requiredMsgValue = _fee.nativeFee + _removeDust(_sendParam.amountLD);
         if (msg.value != requiredMsgValue) {
-            revert InsufficientMessageValue(msg.value, requiredMsgValue);
+            revert IncorrectMessageValue(msg.value, requiredMsgValue);
         }
 
         return super.send(_sendParam, _fee, _refundAddress);
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.22;
 
 import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "./OFTCore.sol";
 
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -5,13 +5,12 @@ pragma solidity ^0.8.22;
 import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "./OFTCore.sol";
 
 /**
- * [AWAITING AUDIT]
  *
  * @title NativeOFTAdapter
  * @dev NativeOFTAdapter is a contract that adapts native currency to the OFT functionality.
  *
  * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
- * unless you make a NON-default implementation of OFT and needs to be done very carefully.
+ * unless you make a NON-default implementation of OFT, which needs to be done very carefully.
  * @dev WARNING: The default NativeOFTAdapter implementation assumes LOSSLESS transfers, ie. 1 native in, 1 native out.
  */
 abstract contract NativeOFTAdapter is OFTCore {
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -103,5 +103,7 @@ abstract contract NativeOFTAdapter is OFTCore {
     }
 
     // @dev Overridden to be empty as this assertion is done higher up on the send function.
-    function _payNative(uint256 _nativeFee) internal override returns (uint256 nativeFee) {}
+    function _payNative(uint256 _nativeFee) internal override returns (uint256 nativeFee) {
+        return _nativeFee;
+    }
 }
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -70,8 +70,9 @@ abstract contract NativeOFTAdapter is OFTCore {
     ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
         // @dev Ensure the native funds in msg.value are enough to cover the fees and amount to send (with dust removed).
         // TODO should we remove dust from _sendParam.amountLD here?
-        if (_fee.nativeFee + _removeDust(_sendParam.amountLD) != msg.value) {
-            revert NotEnoughNative(msg.value);
+        uint256 requiredMsgValue = _fee.nativeFee + _removeDust(_sendParam.amountLD);
+        if (msg.value != requiredMsgValue) {
+            revert InsufficientMessageValue(msg.value, requiredMsgValue);
         }
 
         return super.send(_sendParam, _fee, _refundAddress);
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -76,39 +76,10 @@ abstract contract NativeOFTAdapter is OFTCore {
         MessagingFee calldata _fee,
         address _refundAddress
     ) external payable override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
-        if (msg.value < _sendParam.amountLD) {
-            revert InsufficientMessageValue(msg.value, _sendParam.amountLD);
+        if (_fee.nativeFee + _sendParam.amountLD != msg.value) {
+            revert NotEnoughNative(msg.value);
         }
-        
-        MessagingFee memory feeWithExtraAmount = MessagingFee({
-            nativeFee: _fee.nativeFee,
-            lzTokenFee: _fee.lzTokenFee
-        });
 
-        uint256 remainingMsgValue = msg.value - _sendParam.amountLD;
-        feeWithExtraAmount.nativeFee = remainingMsgValue;
-
-        (uint256 amountSentLD, uint256 amountReceivedLD) = _debit(
-            msg.sender,
-            _sendParam.amountLD,
-            _sendParam.minAmountLD,
-            _sendParam.dstEid
-        );
-
-        // @dev Builds the options and OFT message to quote in the endpoint.
-        (bytes memory message, bytes memory options) = _buildMsgAndOptions(_sendParam, amountReceivedLD);
-
-        // @dev Sends the message to the LayerZero endpoint and returns the LayerZero msg receipt.
-        msgReceipt = _lzSend(_sendParam.dstEid, message, options, feeWithExtraAmount, _refundAddress);
-        
-        // @dev Formulate the OFT receipt.
-        oftReceipt = OFTReceipt(amountSentLD, amountReceivedLD);
-
-        emit OFTSent(msgReceipt.guid, _sendParam.dstEid, msg.sender, amountSentLD, amountReceivedLD);
-    }
-
-    function _payNative(uint256 _nativeFee) internal override returns (uint256 nativeFee) {
-        if (msg.value < _nativeFee) revert NotEnoughNative(msg.value);
-        return _nativeFee;
+        return super.send(_sendParam, _fee, _refundAddress);
     }
 }
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -5,7 +5,7 @@ pragma solidity ^0.8.20;
 import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "./OFTCore.sol";
 
 /**
- * @title NativeOFTAdapter Contract
+ * @title NativeOFTAdapter
  * @dev NativeOFTAdapter is a contract that adapts native currency to the OFT functionality.
  *
  * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
@@ -14,13 +14,10 @@ import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "
  */
 abstract contract NativeOFTAdapter is OFTCore {
 
-    address public constant NATIVE_TOKEN_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
-
     error InsufficientMessageValue(uint256 provided, uint256 required);
     error CreditFailed();
 
     /**
-     * @dev Constructor for the NativeOFTAdapter contract.
      * @param _localDecimals The decimals of the native on the local chain (this chain). 18 on ETH.
      * @param _lzEndpoint The LayerZero endpoint address.
      * @param _delegate The delegate capable of making OApp configurations inside of the endpoint.
@@ -36,12 +33,12 @@ abstract contract NativeOFTAdapter is OFTCore {
      * @return The address of the native token.
      */
     function token() public pure returns (address) {
-        return NATIVE_TOKEN_ADDRESS;
+        return address(0);
     }
 
     /**
      * @notice Indicates whether the OFT contract requires approval of the 'token()' to send.
-     * @return requiresApproval Needs approval of the underlying token implementation.
+     * @return bool indicating whether approval of underlying token implementation is required.
      *
      * @dev In the case of default NativeOFTAdapter, approval is not required.
      */
@@ -80,6 +77,7 @@ abstract contract NativeOFTAdapter is OFTCore {
 
     /**
      * @dev Locks native sent by the sender as msg.value
+     * @dev _from The address to debit.
      * @param _amountLD The amount of native to send in local decimals.
      * @param _minAmountLD The minimum amount to send in local decimals.
      * @param _dstEid The destination chain ID.
@@ -118,7 +116,11 @@ abstract contract NativeOFTAdapter is OFTCore {
         return _amountLD;
     }
 
-    // @dev Overridden to be empty as this assertion is done higher up on the overriden send() function.
+    /**
+     * @dev Overridden to be empty as this assertion is done higher up on the overriden send() function.
+     * @param _nativeFee The native fee to be paid.
+     * @return nativeFee The amount of native currency paid.
+     */
     function _payNative(uint256 _nativeFee) internal pure override returns (uint256 nativeFee) {
         return _nativeFee;
     }
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "./OFTCore.sol";
+
+/**
+ * @title NativeOFTAdapter Contract
+ * @dev NativeOFTAdapter is a contract that adapts native currency to the OFT functionality.
+ *
+ * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
+ * unless you make a NON-default implementation of OFT and needs to be done very carefully.
+ * @dev WARNING: The default NativeOFTAdapter implementation assumes LOSSLESS transfers, ie. 1 native in, 1 native out.
+ */
+abstract contract NativeOFTAdapter is OFTCore {
+
+    error DepositFailed();
+    error InsufficientMessageValue(uint256 provided, uint256 required);
+    error WithdrawalFailed();
+
+    /**
+     * @dev Constructor for the NativeOFTAdapter contract.
+     * @param _localDecimals The decimals of the native on the local chain (this chain). 18 on ETH.
+     * @param _lzEndpoint The LayerZero endpoint address.
+     * @param _delegate The delegate capable of making OApp configurations inside of the endpoint.
+     */
+    constructor(
+        uint8 _localDecimals,
+        address _lzEndpoint,
+        address _delegate
+    ) OFTCore(_localDecimals, _lzEndpoint, _delegate) {}
+
+    /**
+     * @dev Locks native sent by the sender as msg.value
+     * @param _from The address to debit from.
+     * @param _amountLD The amount of native to send in local decimals.
+     * @param _minAmountLD The minimum amount to send in local decimals.
+     * @param _dstEid The destination chain ID.
+     * @return amountSentLD The amount sent in local decimals.
+     * @return amountReceivedLD The amount received in local decimals on the remote.
+     */
+    function _debit(
+        address _from,
+        uint256 _amountLD,
+        uint256 _minAmountLD,
+        uint32 _dstEid
+    ) internal virtual override returns (uint256 amountSentLD, uint256 amountReceivedLD) {
+        // @dev Lock native funds by moving them into this contract from the caller.
+        (amountSentLD, amountReceivedLD) = _debitView(_amountLD, _minAmountLD, _dstEid);
+
+        (bool success, ) = payable(address(this)).call{value: amountSentLD}("");
+        if (!success) {
+            revert DepositFailed();
+        }
+    }
+
+    /**
+     * @dev Credits native to the specified address.
+     * @param _to The address to credit the native to.
+     * @param _amountLD The amount of native to credit.
+     * @dev _srcEid The source chain ID.
+     * @return amountReceivedLD The amount of native ACTUALLY received.
+     */
+    function _credit(
+        address _to,
+        uint256 _amountLD,
+        uint32 /*_srcEid*/
+    ) internal virtual override returns (uint256 amountReceivedLD) {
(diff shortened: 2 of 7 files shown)


packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,119 @@
...
+ * @title NativeOFTAdapter Contract
+ * @dev NativeOFTAdapter is a contract that adapts native currency to the OFT functionality.
+ *
+ * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
+ * unless you make a NON-default implementation of OFT and needs to be done very carefully.
+ * @dev WARNING: The default NativeOFTAdapter implementation assumes LOSSLESS transfers, ie. 1 native in, 1 native out.
+ */
+abstract contract NativeOFTAdapter is OFTCore {
+
+    error DepositFailed();
```
TRileySchwarz, line 17: debit failed

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,119 @@
...
+    error InsufficientMessageValue(uint256 provided, uint256 required);
+    error WithdrawalFailed();
```
TRileySchwarz, line 19: credit failed

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,119 @@
...
+    function _debit(
+        address _from,
+        uint256 _amountLD,
+        uint256 _minAmountLD,
+        uint32 _dstEid
+    ) internal virtual override returns (uint256 amountSentLD, uint256 amountReceivedLD) {
+        // @dev Lock native funds by moving them into this contract from the caller.
+        (amountSentLD, amountReceivedLD) = _debitView(_amountLD, _minAmountLD, _dstEid);
+
+        (bool success, ) = payable(address(this)).call{value: amountSentLD}("");
```
TRileySchwarz, line 51: This doesnt do anything, the msg.value automatically moves all the funds to the contract

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,114 @@
...
+        (bool success, ) = payable(_to).call{value: _amountLD}("");
+        if (!success) {
+            revert CreditFailed();
+        }
+
+        // @dev In the case of NON-default NativeOFTAdapter, the amountLD MIGHT not be == amountReceivedLD.
+        return _amountLD;
+    }
+
+    function send(
```
TRileySchwarz, line 74: function send(
    SendParam calldata _sendParam,
    MessagingFee calldata _fee,
    address _refundAddress
) external payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
    if (_fee.nativeFee + _sendParam.amountLD != msg.value) {
        revert NotEnoughNative(msg.value);
    }
    
    return super.send(_sendParam, _fee, _refundAddress);
}

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,114 @@
...
+        // @dev Sends the message to the LayerZero endpoint and returns the LayerZero msg receipt.
+        msgReceipt = _lzSend(_sendParam.dstEid, message, options, feeWithExtraAmount, _refundAddress);
+        
+        // @dev Formulate the OFT receipt.
+        oftReceipt = OFTReceipt(amountSentLD, amountReceivedLD);
+
+        emit OFTSent(msgReceipt.guid, _sendParam.dstEid, msg.sender, amountSentLD, amountReceivedLD);
+    }
+
+    function _payNative(uint256 _nativeFee) internal override returns (uint256 nativeFee) {
```
TRileySchwarz, line 110: override this to just do nothing

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,114 @@
...
+     * @param _amountLD The amount of native to credit.
+     * @dev _srcEid The source chain ID.
+     * @return amountReceivedLD The amount of native ACTUALLY received.
+     */
+    function _credit(
+        address _to,
+        uint256 _amountLD,
+        uint32 /*_srcEid*/
+    ) internal virtual override returns (uint256 amountReceivedLD) {
+        // @dev Unlock the tokens and transfer to the recipient.
```
TRileySchwarz, line 64: // @dev  Transfer tokens to the recipient.

packages/oft-evm/contracts/NativeOFTAdapter.sol
TRileySchwarz: I would also comment the reason for muting this out was that the assertion is now done on the send higher up in this contract

TRileySchwarz:
Also, need to add the token() and approvalRequired() functions

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: @TRileySchwarz Do we still want to remove dust from the `amountLD` before sending? If so, I don't think we can remove dust from the `amountLD` here since `_sendParam` is of type `calldata`. Creating a new `sendParam` where we set `amountLD` with dust removed also won't work because it would be in memory, which doesn't align with the function parameters we end up passing it to

Or perhaps you meant the dust should already be removed in the incoming `_sendparam`?

TRileySchwarz: we could in the assertion, mutate that param, but not adjust it in the super.send call. Alternatively I think we could just not do this and accumulate the dust

ravinagill15: > we could in the assertion, mutate that param, but not adjust it in the super.send call

I don't think we can mutate it since it's of type `calldata`, but I suppose we could remove the dust in the assertion:
```
if (_fee.nativeFee + _removeDust(_sendParam.amountLD) != msg.value) {
            revert NotEnoughNative(msg.value);
}
```

Alternatively, If we accumulate the dust, i'm guessing we need some kind of withdraw function for it?

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    function send(
+        SendParam calldata _sendParam,
+        MessagingFee calldata _fee,
+        address _refundAddress
+    ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
+        if (_fee.nativeFee + _sendParam.amountLD != msg.value) {
+            revert NotEnoughNative(msg.value);
+        }
```
ravinagill15, line 82: Found out while writing tests that `msg.value` must be the same as `_fee.nativeFee`, even though we are overriding `_payNative` to be empty. This is due to `EndpointV2Mock._assertMessagingFee` requiring `msg.value` to be equal to `_fee.nativeFee` [here](https://github.com/LayerZero-Labs/devtools/blob/c745cf8d0992e2f8df7d00388596fe1b9c4b160b/packages/test-devtools-evm-foundry/contracts/mocks/EndpointV2Mock.sol#L318-L323)

I will look for a way around this

ravinagill15: This actually may be due to the `msg.value` not being carried over to `EndpointV2Mock._assertMessagingFee`, still looking!

ravinagill15: Figured this out -- `_payNative` cannot be empty, it must just return the parameter

packages/oft-evm/contracts/NativeOFTAdapter.sol
TRileySchwarz: yeah sorry thats what i meant, do it like that. Technically does removeDust twice, but we can make this decision later

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,124 @@
...
+     *  - guid: The unique identifier for the sent message.
+     *  - nonce: The nonce of the sent message.
+     *  - fee: The LayerZero fee incurred for the message.
+     */
+    function send(
+        SendParam calldata _sendParam,
+        MessagingFee calldata _fee,
+        address _refundAddress
+    ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
+        if (_fee.nativeFee + _sendParam.amountLD != msg.value) {
```
TRileySchwarz, line 72: comment this in detail

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,124 @@
...
+     * @return amountSentLD The amount sent in local decimals.
+     * @return amountReceivedLD The amount received in local decimals on the remote.
+     */
+    function _debit(
+        address /*_from*/,
+        uint256 _amountLD,
+        uint256 _minAmountLD,
+        uint32 _dstEid
+    ) internal virtual override returns (uint256 amountSentLD, uint256 amountReceivedLD) {
+        // @dev Lock native funds by moving them into this contract from the caller.
```
TRileySchwarz, line 93: summarize these comments better

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,124 @@
...
+        (bool success, ) = payable(_to).call{value: _amountLD}("");
+        if (!success) {
+            revert CreditFailed();
+        }
+
+        // @dev In the case of NON-default NativeOFTAdapter, the amountLD MIGHT not be == amountReceivedLD.
+        return _amountLD;
+    }
+
+    // @dev Overridden to be empty as this assertion is done higher up on the send function.
```
TRileySchwarz, line 120: "...done higher up on the overriden send() function."

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,124 @@
...
+     *  - guid: The unique identifier for the sent message.
+     *  - nonce: The nonce of the sent message.
+     *  - fee: The LayerZero fee incurred for the message.
+     */
+    function send(
+        SendParam calldata _sendParam,
+        MessagingFee calldata _fee,
+        address _refundAddress
+    ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
+        if (_fee.nativeFee + _removeDust(_sendParam.amountLD) != msg.value) {
```
TRileySchwarz, line 72: comment this in detail, and add a second TODO comment about whether we want to keep the removeDust

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "./OFTCore.sol";
+
+/**
+ * @title NativeOFTAdapter Contract
```
ryandgoulding, line 8: I would drop `Contract` here.

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,125 @@
...
+ */
+abstract contract NativeOFTAdapter is OFTCore {
+
+    address public constant NATIVE_TOKEN_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
+
+    error InsufficientMessageValue(uint256 provided, uint256 required);
+    error CreditFailed();
+
+    /**
+     * @dev Constructor for the NativeOFTAdapter contract.
```
ryandgoulding, line 23: I would get rid of this line.  Don't need the @dev comment here as it doesn't add anything.

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,125 @@
...
+ * @title NativeOFTAdapter Contract
+ * @dev NativeOFTAdapter is a contract that adapts native currency to the OFT functionality.
+ *
+ * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
+ * unless you make a NON-default implementation of OFT and needs to be done very carefully.
+ * @dev WARNING: The default NativeOFTAdapter implementation assumes LOSSLESS transfers, ie. 1 native in, 1 native out.
+ */
+abstract contract NativeOFTAdapter is OFTCore {
+
+    address public constant NATIVE_TOKEN_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
ryandgoulding, line 17: I will defer to @TRileySchwarz here, but we pretty much adapted the non-standard convention of using `address(0)` instead of this.  I am not quite sure what we want to support going forward, but historically we used `address(0)` as it is chain agnostic.

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,125 @@
...
+     * @dev Returns the address of the native token
+     * @return The address of the native token.
+     */
+    function token() public pure returns (address) {
+        return NATIVE_TOKEN_ADDRESS;
+    }
+
+    /**
+     * @notice Indicates whether the OFT contract requires approval of the 'token()' to send.
+     * @return requiresApproval Needs approval of the underlying token implementation.
```
ryandgoulding, line 44: This documentation is off.  In this case, you are not using named return values.

packages/oft-evm/contracts/OFTCore.sol
```diff
@@ -174,7 +174,7 @@ abstract contract OFTCore is IOFT, OApp, OAppPreCrimeSimulator, OAppOptionsType3
         SendParam calldata _sendParam,
         MessagingFee calldata _fee,
         address _refundAddress
-    ) external payable virtual returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
+    ) public payable virtual returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
```
ryandgoulding, line 177: Why was this made public?

packages/oft-evm/test/OFT.t.sol
```diff
@@ -51,13 +58,18 @@ contract OFTTest is TestHelperOz5 {
     address public userA = makeAddr("userA");
     address public userB = makeAddr("userB");
     address public userC = makeAddr("userC");
+    address public userD = makeAddr("userD");
     uint256 public initialBalance = 100 ether;
+    uint256 public initialNativeBalance = 1000 ether;
+
+
+    address public constant NATIVE_TOKEN_ADDRESS = address(0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE);
```
ryandgoulding, line 66: Add a comment like:

```
// must match NativeOFTAdapter.sol
```

ryandgoulding (commented):
Great work!  A few small comments here and there, but structurally looks great.

packages/oft-evm/contracts/OFTCore.sol
ravinagill15: In order to call `super.send(...)` in `NativeOFTAdapter.send(...)` we had to change this from external to public. Do you foresee any issues with this change?

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: Ah I see, makes sense! @TRileySchwarz I will change this to return `address(0)` based on @ryandgoulding's recommendation above, please lmk if you disagree

packages/oft-evm/test/OFT.t.sol
ravinagill15: I will be removing this in favour of using `address(0)` as [discussed above](https://github.com/LayerZero-Labs/devtools/pull/838#discussion_r1716762372)

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: TODO has been added -- @TRileySchwarz @ryandgoulding what are your thoughts on whether we should remove dust from `amountLD` in the assertion here?

I think it makes sense to keep because it ensures the exact amount of `msg.value` is sent with no excess

.changeset/shaggy-actors-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@layerzerolabs/oft-evm": minor
```
ravinagill15, line 2: @TRileySchwarz @ryandgoulding @janjakubnanista Just want to confirm that this change would be considered minor?

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,127 @@
...
+     */
+    function send(
+        SendParam calldata _sendParam,
+        MessagingFee calldata _fee,
+        address _refundAddress
+    ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
+        // @dev Ensure the native funds in msg.value are enough to cover the fees and amount to send (with dust removed).
+        // TODO should we remove dust from _sendParam.amountLD here?
+        if (_fee.nativeFee + _removeDust(_sendParam.amountLD) != msg.value) {
+            revert NotEnoughNative(msg.value);
```
ravinagill15, line 72: Wondering if instead of reverting with `NotEnoughNative` we should revert with a custom error that also indicates how much native _should_ have been provided (i.e `_fee.nativeFee + _removeDust(_sendParam.amountLD)`)

.changeset/shaggy-actors-call.md
ravinagill15: Confirmed with @janjakubnanista that yes it would be a minor change

ravinagill15:
Looking into build failures!

ravinagill15:
Please do not merge until the NativeOFTAdapter contract has been audited

packages/oft-evm/contracts/NativeOFTAdapter.sol
TRileySchwarz: it technically is the same purpose, but I could be convinced to do a separate error. The issue is that the original assertion for the fee->msg.value doesn't happen now, so the underlying behaviour is inconsistent

packages/oft-evm/contracts/NativeOFTAdapter.sol
TRileySchwarz: Calculate the extra gas for doing this, that will determine the outcome imo

packages/oft-evm/contracts/OFTCore.sol
TRileySchwarz: This wont effect things down the line, or break interfaces etc. we should probably make this public imo

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: We only use an extra 194 gas when removing dust here vs without removing dust. Im thinking we should still keep it since its quite low?

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: Gotcha, yeah in that case I'll add a new custom error specifically for this use case. It will also help the user see what they should have passed for `msgValue` if they accidentally pass an incorrect value

ravinagill15: Actually, I think we should reuse `InsufficientMessageValue` error here as that already lets us indicate `provided` and `required` and fits this use case

packages/oft-evm/contracts/OFTCore.sol
ryandgoulding: It does make it slightly more expensive, but I see that it is needed.  I don't disagree.

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+     */
+    function _credit(
+        address _to,
+        uint256 _amountLD,
+        uint32 /*_srcEid*/
+    ) internal virtual override returns (uint256 amountReceivedLD) {
+        // @dev Transfer tokens to the recipient.
+        (bool success, ) = payable(_to).call{value: _amountLD}("");
+        if (!success) {
+            revert CreditFailed();
```
janjakubnanista, line 115: @TRileySchwarz not sure whether it's valuable to pass the `data` from the `call` to the error for debugging, whatcha think?

packages/oft-evm/contracts/OFTCore.sol
janjakubnanista: What about moving the body to `internal` function (`_send` or smth) and using that instead of modifying the interface?

packages/oft-evm/test/OFT.t.sol
```diff
@@ -373,6 +395,74 @@ contract OFTTest is TestHelperOz5 {
...
+        );
+
+        MessagingFee memory fee = dNativeOFTAdapter.quoteSend(sendParam, false);
+        uint256 msgValue = fee.nativeFee + sendParam.amountLD;
+
+        vm.prank(userD);
+        vm.expectRevert(
+            abi.encodeWithSelector(NativeOFTAdapter.InsufficientMessageValue.selector, fee.nativeFee, msgValue)
+        );
+        dNativeOFTAdapter.asNativeOFTAdapterMock().send{ value: fee.nativeFee}(sendParam, fee, userD);
```
janjakubnanista, line 457: Two things:

- I would add comments about what's being tested and why it should fail
- I would add the missing case for when the `msgValue` is higher than `fee.nativeFee + sendParam.amountLD` because that triggers a revert as well

packages/oft-evm/test/mocks/NativeOFTAdapterMock.sol
```diff
@@ -0,0 +1,27 @@
...
+        uint256 _amountToSendLD,
+        uint256 _minAmountToCreditLD,
+        uint32 _dstEid
+    ) public returns (uint256 amountDebitedLD, uint256 amountToCreditLD) {
+        return _debit(msg.sender, _amountToSendLD, _minAmountToCreditLD, _dstEid);
+    }
+
+    function credit(address _to, uint256 _amountToCreditLD, uint32 _srcEid) public returns (uint256 amountReceivedLD) {
+        return _credit(_to, _amountToCreditLD, _srcEid);
+    }
```
janjakubnanista, line 26: In general (so maybe in this case it's not applicable) testing internal functions is not the best idea. Internal functions can change as long as the external interface does not.

In these cases, I recommend testing these in the context of the external flow as they will never be run in isolation.

Check with @TRileySchwarz and @ryandgoulding though before changing anything, maybe there is some value to having them tested.

packages/oft-evm/contracts/OFTCore.sol
ravinagill15: I like that idea because it creates a consistent pattern between `OFTCore.send` and `NativeOFTAdapter.send` that can be referenced in similar future implementations.

I have a PR here that does exactly that: https://github.com/LayerZero-Labs/devtools/pull/877

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: That sounds like a good idea to me! I have a follow up PR here: https://github.com/LayerZero-Labs/devtools/pull/878

packages/oft-evm/test/OFT.t.sol
ravinagill15: > I would add comments about what's being tested and why it should fail

Will do!

> I would add the missing case for when the msgValue is higher than fee.nativeFee + sendParam.amountLD because that triggers a revert as well

Oo yes great catch! I'll write this in the next commit

packages/oft-evm/test/mocks/NativeOFTAdapterMock.sol
ravinagill15: I see what you mean, makes sense! I had originally done this to follow the [testing pattern of OFTAdapter](https://github.com/LayerZero-Labs/devtools/blob/555752ef24c787164cecd551aad363978da4d381/packages/oft-evm/test/OFT.t.sol#L329-L374), but if @TRileySchwarz and @ryandgoulding also agree, then I have a PR [here](https://github.com/LayerZero-Labs/devtools/pull/879) that removes testing internal functions. We already test the external flow in `test_native_oft_adapter_send()` so this PR just removes the internal function testing.

packages/oft-evm/test/OFT.t.sol
ravinagill15: I decided to make a separate PR for this to keep things cleaner since this PR has already been sent for audit as is

https://github.com/LayerZero-Labs/devtools/pull/880

packages/oft-evm/test/mocks/NativeOFTAdapterMock.sol
ryandgoulding: Just makes it easier to encapsulated functions in a more controlled manner.  Since the contracts are immutable, it is generally pretty safe as any implementation change is likely going to go through re-audit, redeployment, etc.  But, it is fine to remove them if they are tested elsewhere.  In general, agreed, it is a shortcut to testing the API as it will actually be used.

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.22;
+
+import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "./OFTCore.sol";
+
+/**
+ * [AWAITING AUDIT]
```
ryandgoulding, line 8: Should this be removed now?

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @title NativeOFTAdapter
+ * @dev NativeOFTAdapter is a contract that adapts native currency to the OFT functionality.
+ *
+ * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
+ * unless you make a NON-default implementation of OFT and needs to be done very carefully.
```
ryandgoulding, line 14: NIT: "which" needs to be done very carefully.

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -0,0 +1,130 @@
...
+        // @dev In the case of NON-default NativeOFTAdapter, the amountLD MIGHT not be == amountReceivedLD.
+        return _amountLD;
+    }
+
+    /**
+     * @dev Overridden to be empty as this assertion is done higher up on the overriden send() function.
+     * @param _nativeFee The native fee to be paid.
+     * @return nativeFee The amount of native currency paid.
+     */
+    function _payNative(uint256 _nativeFee) internal pure override returns (uint256 nativeFee) {
```
ryandgoulding, line 127: Why name the return parameter if you are not assigning it?

ryandgoulding (dismissed):

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: Great eye!

packages/oft-evm/contracts/NativeOFTAdapter.sol
ravinagill15: That's a great point that I didn't think of! I will keep in mind that return variables should only be named if being assigned. Considering this has already been audited, I am hesitant to change any code even though I removed the return parameter name locally and the tests still passed

Please let me know if you strongly recommend this change, if so I will look into whether the auditors are okay with this minor change 🫡

ryandgoulding: No it is fine, just wanted to make sure I wasn't overlooking anything.

janjakubnanista (approved):

PR #841, opened 2024-08-13
chore: update solana oft mock name

St0rmBr3w:


ryandgoulding:
Applies to an outdated version of OFT for Solana

PR #849, opened 2024-08-14
🗞️ Adding NativeOFTAdapter example to devtools

ravinagill15:
## In this PR
- Added NativeOFTAdapter example to devtools


ravinagill15:
Please do not merge until the `NativeOFTAdapter` contract within `feat/native-oft-adapter` has been audited

tests-user/tests/create-lz-oapp.bats
```diff
@@ -211,6 +222,17 @@ teardown() {
     yarn lint:fix
 }
 
+@test "should work with yarn & native-oft-adapter example in CI mode" {
+    local DESTINATION="$PROJECTS_DIRECTORY/yarn-oapp"
```
ravinagill15, line 226: Should this be `yarn-native-oft-adapter` instead? I copied this over from the `oft-adapter` test right above which also lists the destination as `yarn-oapp`. I can fix both if this is a typo

tests-user/tests/create-lz-oapp.bats
```diff
@@ -264,6 +286,17 @@ teardown() {
     npm run lint:fix
 }
 
+@test "should work with npm & native-oft-adapter example in CI mode" {
+    local DESTINATION="$PROJECTS_DIRECTORY/npm-oft"
```
ravinagill15, line 290: Same question as above here -- Should this be `npm-native-oft-adapter` instead? I copied this over from the oft-adapter test right above which also lists the destination as `npm-oft`. I can fix both if this is a typo

examples/native-oft-adapter/deploy/MyNativeOFTAdapter.ts
```diff
@@ -0,0 +1,52 @@
...
+    //     ...
+    //     eid: EndpointId.AVALANCHE_V2_TESTNET
+    //   }
+    // }
+    const endpointV2Deployment = await hre.deployments.get('EndpointV2')
+
+    const { address } = await deploy(contractName, {
+        from: deployer,
+        args: [
+            18, // TODO update this with native local decimals
```
ryandgoulding, line 39: This is where I really dislike deploy scripts, as they do not support mandatory arguments.  `18` is almost always correct, unless it isn't, which is rare but occasionally happens.  EVM has no API to determine gas token decimals that I am aware of (you would likely need to call a known WETH9 precompile).  I think this is the best we can do given the limitations, thoughts?

ravinagill15: Thanks for the feedback! I agree that hardcoding 18 isn't ideal, but given the lack of an API to determine gas token decimals, this _seems_ like the best option for now. Open to revisiting this if we find a better solution!

ryandgoulding (dismissed):

ryandgoulding (approved):

PR #856, opened 2024-08-15
🧹 Decoupling from hardhat deploy: Embedding EndpointV2 & Uln302 ABIs [1/N]

janjakubnanista:
### In this PR

- This PR changes the mechanism through which we encode/decode data for `EndpointV2` & `Uln302` SDKs. 
- The initial idea and pattern of SDK creation relied heavily on `hardhat-deploy` and in turn on `ethers` - the ABIs were loaded dynamically from deployments and wrapped in `Contract` instances. This dynamicity though is only an illusion - if the loaded ABIs didn't match the expected ones, the SDK would not be able to deal with the differences. On top of that, it makes the EVM SDKs rely too much on `hardhat-deploy` that is slowly getting pushed out and replaced by `hardhat ignition` or `forge` scripts.
- This PR embeds the core ABIs for `EndpointV2` and `Uln302` in a way that will later support a possible transition to `viem`. 
- The base classes and interfaces have not yet reflected this change - the EVM `OmniSDK` still does not expose `abi` or `provider` and the constructor has been left untouched, something to come in following PRs

The overall vision is:

- SDKs uses ABI and a provider to talk to the chain. While we're on ethers, this means creating a `Contract` instance
- This ABI is always defaulted to an appropriate ABI but can be overridden
- To maintain support for large use-cases such as Stargate, hardhat-deploy-based ABI factory will be provided


packages/omnicounter-devtools-evm/src/omnicounter/sdk.ts
```diff
@@ -1,18 +1,9 @@
...
 import { type OmniAddress, makeBytes32 } from '@layerzerolabs/devtools'
-import { OmniContract } from '@layerzerolabs/devtools-evm'
 
 export class OmniCounter extends OApp implements IOmniCounter {
-    public constructor(
-        public override contract: OmniContract,
-        protected override endpointV2Factory: EndpointV2Factory
-    ) {
-        super(contract, endpointV2Factory)
-    }
```
ryandgoulding, line 14: Got it, because we removed endpointV2Factory, we can just rely on OApp constructor.  Leaving this here for anyone else reviewing.

packages/omnicounter-devtools-evm/src/omnicounter/factory.ts
```diff
@@ -13,8 +11,5 @@ import { OmniContractFactory } from '@layerzerolabs/devtools-evm'
  * @param {EndpointV2Factory} [EndpointV2Factory]
```
ryandgoulding, line 11: Can remove `@param {EndpointV2Factory} [EndpointV2Factory]` now.

packages/protocol-devtools-evm/src/uln302/factory.ts
```diff
@@ -10,6 +11,5 @@ import { Uln302 } from './sdk'
  * @param {OmniContractFactory} contractFactory
```
ryandgoulding, line 11: Update docs to `providerFactory`

ryandgoulding (dismissed):
Just a few small nits about updating documentation.  Looks great!

ryandgoulding (approved):

PR #857, opened 2024-08-16
🧹 Set config example scripts

ravinagill15:
## In this PR
- Example solidity scripts to demonstrate how send and receive configs can be set

## Before running

1. Update `.env` with private key, rpc urls, and etherscan api key
2. Update foundry.toml` with rpc urls and etherscan api keys

## To run the `SendConfig.s.sol` script:

```
 source .env && forge script --chain sepolia forge-scripts/SendConfig.s.sol:SendConfig --rpc-url $SEPOLIA_RPC_URL --broadcast --verify -vvvv --sig "run(address, uint32)" 0xC51c580Eeb3844b4117C9B3f5e9Cc43f5B808A85 40231
```

## To run the `ReceiveConfig.s.sol` script:

```
source .env && forge script --chain sepolia forge-scripts/ReceiveConfig.s.sol:ReceiveConfig --rpc-url $SEPOLIA_RPC_URL --broadcast --verify -vvvv --sig "run(address, uint32)" 0xC51c580Eeb3844b4117C9B3f5e9Cc43f5B808A85 40231
```


scripts/forge-scripts/MyOFT.s.sol
ravinagill15, line 1: Unneeded, but forge compilation fails without this....will dig into how we can remove this

forge-config-scripts/scripts/ReceiveConfig.s.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
```
ryandgoulding, line 2: any reason you used this version?  Most OFT code uses 0.8.22

forge-config-scripts/scripts/ReceiveConfig.s.sol
```diff
@@ -0,0 +1,63 @@
...
+import { ExecutorConfig } from "@layerzerolabs/lz-evm-messagelib-v2/contracts/SendLibBase.sol";
+
+contract ReceiveConfig is Script {
+    uint32 public constant RECEIVE_CONFIG_TYPE = 2;
+
+    function run(address contractAddress, uint32 remoteEid) external {
+        uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
+
+        // ============================ Can deploy contract or fetch deployed contract ============================
+        // MyOFT myOFT = new MyOFT("MyOFT", "OFT", address(0x6EDCE65403992e310A62460808c4b910D972f10f), address(0x565786AbE5BA0f9D307AdfA681379F0788bEdEf7));
```
ryandgoulding, line 21: Not sure if we want to deploy as a part of this script.  I would argue to separate concerns.

forge-config-scripts/scripts/ReceiveConfig.s.sol
```diff
@@ -0,0 +1,63 @@
...
+        MyOFT myOFT = MyOFT(contractAddress);
+
+        ILayerZeroEndpointV2 endpoint = ILayerZeroEndpointV2(address(myOFT.endpoint()));
+        
+        (address receiveLibraryAddress, ) = endpoint.getReceiveLibrary(address(myOFT), remoteEid);
+
+        console.log("============================ PREPARING RECEIVE CONFIG ============================");
+        
+        UlnConfig memory receiveConfig = UlnConfig({
+                confirmations: 59,
```
ryandgoulding, line 32: Where is this number coming from?  We really ought to set it based on the defaults, not just a random magic number.

forge-config-scripts/scripts/ReceiveConfig.s.sol
```diff
@@ -0,0 +1,63 @@
...
+                requiredDVNs: new address[](0),
```
ryandgoulding, line 33: Is there risk to doing this?  Like, can setting nothing mean every message is valid?  I haven't looked at that code recently.

forge-config-scripts/scripts/ReceiveConfig.s.sol
```diff
@@ -0,0 +1,63 @@
...
+        endpoint.setConfig(address(myOFT), receiveLibraryAddress, setReceiveConfigParams);
+
+        vm.stopBroadcast();
+
+        console.log("============================ GETTING UPDATED RECEIVE CONFIG ============================");
+        
+        bytes memory updatedUlnConfigBytes = endpoint.getConfig(address(myOFT), receiveLibraryAddress, remoteEid, 2);
+        UlnConfig memory updatedUlnConfig = abi.decode(updatedUlnConfigBytes, (UlnConfig));
+        
+        console.log(" Updated Uln Config Confirmations: ");
```
ryandgoulding, line 60: extra leading white space

forge-config-scripts/scripts/SendConfig.s.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
```
ryandgoulding, line 2: same comment about version.

forge-config-scripts/scripts/SendConfig.s.sol
```diff
@@ -0,0 +1,78 @@
...
+
+        ILayerZeroEndpointV2 endpoint = ILayerZeroEndpointV2(address(myOFT.endpoint()));
+        
+        address sendLibraryAddress = endpoint.getSendLibrary(address(myOFT), remoteEid);
+
+        console.log("============================ PREPARING SEND CONFIG ============================");
+
+        ExecutorConfig memory executorConfig = ExecutorConfig({ maxMessageSize: 50, executor: 0x71d7a02cDD38BEa35E42b53fF4a42a37638a0066});
+        UlnConfig memory ulnConfig = UlnConfig({
+                                    confirmations: 99,
```
ryandgoulding, line 34: same comments.

ryandgoulding (changes requested):

ryandgoulding (commented):
Cool stuff!  Had some nitty comments, but the approach looks right.  Let me know if you have any questions!

forge-config-scripts/scripts/ReceiveConfig.s.sol
ravinagill15: Nice catch -- just a copy paste error!

forge-config-scripts/scripts/ReceiveConfig.s.sol
ravinagill15: Good point -- I will use fuzz testing instead!

forge-config-scripts/scripts/ReceiveConfig.s.sol
ravinagill15: Now using fuzz testing to set such configs

packages/ua-devtools-evm-foundry/src/ReceiveConfig.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
```
ravinagill15, line 1: Please double check licenses

ravinagill15:
Looking into cyclical dependencies!

ravinagill15:
Added `do not merge` label because we have not yet made a decision on whether we want to support foundry-only development. The reason is because this example cannot be used on non-evm chains, so if a client wants to move to solana, for example, they would need to learn a whole new set of tools.

ryandgoulding:
Is this something that is still actively being worked on / developed?  There is no pressure, I would just prefer to close this PR if this is not planned to be completed.

ravinagill15:
This PR has been referenced in the past for clients that prefer to only use foundry tooling, but I agree that it doesn't look like we will continue work on this PR. I have updated the [Tech Debt Jungle Preproduction](https://docs.google.com/document/d/1Dle9GEizvgfD3Ngemafy2IaJz5t8QCwH8tylBYDbA-Y/edit?usp=sharing) document (under devtools section) to consider revisiting this PR since we will be rewriting devtools. 

Will close this PR out for now!

cc: @ryandgoulding @bholcomb8 @St0rmBr3w

shankars99:
if we ever revive this and want to run automated testing (https://github.com/LayerZero-Labs/lz-foundry-utils) is the lz config equivalent that i cooked up 

https://github.com/LayerZero-Labs/maple-finance-syrup-integration/tree/main/script has some scripts for the diff operations 

using this is quite simple you just import it and go with things like 

```
uint32 aEid = getEid(chainName);
string bRPC = getRpcUrl(chainName);
address bEndpoint = getAddressLZ(chainName, "Contract Name");
```

https://github.com/LayerZero-Labs/lz-foundry-utils/blob/main/test/Json.t.sol has the complete usage

PR #863, opened 2024-08-19
🧹 Decouple UA SDKs from hardhat deploy

janjakubnanista:
### In this PR

- Decouple UA SDKs from `hardhat-deploy`. This PR will be split into several ones, it is here only to see the test runs

### DIFF
--- a/packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
+++ b/packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
@@ -3,7 +3,7 @@ import pMemoize from 'p-memoize'
 import { OmniContract } from '@layerzerolabs/devtools-evm'
 import { Contract } from '@ethersproject/contracts'
 import assert from 'assert'
-import { OmniContractFactoryHardhat, OmniDeployment } from './types'
+import { AbiFactoryHardhat, OmniContractFactoryHardhat, OmniDeployment } from './types'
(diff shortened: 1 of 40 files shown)


packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
```diff
@@ -18,6 +18,75 @@ export const omniDeploymentToContract = ({ eid, deployment }: OmniDeployment): O
...
+        if (contractName != null && address == null) {
+            logger.verbose(`Looking for contract ${contractName} in deployments`)
+
+            const deployment = await env.deployments.getOrNull(contractName)
+            assert(deployment != null, `Could not find a deployment for contract '${contractName}' on ${networkLabel}`)
+
+            return deployment.abi
+        }
+
+        // And if we only have the address, we need to go get it from deployments by address
```
St0rmBr3w, line 53: Doesn't this still enforce that you have the hardhat-deploy JSON artifact in the deployments folder?

Wouldn't a true decoupling mean we provide the ABI for the base OApp standard and just check that the address has these ABI methods?

janjakubnanista: Decoupling will not mean that we lose all the hardhat deploy functionality, we'll still provide it for people to use and they can opt in if they want. For example Stargate still benefits from hardhat-deploy functionality

PR #874, opened 2024-08-28
🧹 Validate RPC urls

ravinagill15:
## In this PR:
- https://layerzerolabs.atlassian.net/browse/APPS-84
- A task to validate all RPC URLs within the local `hardhat.config.ts` has been implemented within `devtools-evm-hardhat`
- A corresponding test file `validate.rpcs.test.ts` has also been created.


packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,134 @@
...
+        return false
+    } catch (error) {
+        logger.error(
+            `Validation failed for RPC URL: ${error instanceof Error ? error.message : 'An unknown error occurred'}`
+        )
+        return false
+    }
+}
+
+const validateWebSocketRpcUrl = async (rpcUrl: string, timeout: number): Promise<boolean> => {
```
ravinagill15, line 60: This currently passes _any_ responsive wss instead of only responsive wss **for rpcs**. I will fix this in a follow up commit!

tests/devtools-evm-hardhat-test/test/task/validate.rpcs.test.ts
```diff
@@ -0,0 +1,69 @@
+/// <reference types="jest-extended" />
+
+import hre from 'hardhat'
+import { join } from 'path'
+import { TASK_LZ_VALIDATE_RPCS } from '@layerzerolabs/devtools-evm-hardhat'
+import { spawnSync } from 'child_process'
+
+describe(`task ${TASK_LZ_VALIDATE_RPCS}`, () => {
+    describe('expectations', () => {
```
ravinagill15, line 9: Will add a test validating multiple RPCs now that I have removed the test RPCs

packages/devtools-evm-hardhat/src/constants/tasks.ts
```diff
@@ -10,4 +10,6 @@ export const TASK_LZ_TEST_SIMULATION_LOGS = 'lz:test:simulation:logs'
 
 export const TASK_LZ_TEST_SIMULATION_STOP = 'lz:test:simulation:stop'
 
-export const TASK_LZ_VALIDATE_SAFE_CONFIGS = 'lz:validate-safe-configs'
+export const TASK_LZ_VALIDATE_SAFE_CONFIGS = 'lz:validate-safe-configs' // TODO rename this to lz:validate:safe-configs in followup PR
+
+export const TASK_LZ_VALIDATE_RPCS = 'lz:validate-rpcs' // TODO rename this to lz:validate:rpcs in followup PR
```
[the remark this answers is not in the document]
ravinagill15: Follow up PR: https://github.com/LayerZero-Labs/devtools/pull/876

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,161 @@
...
+    logger.error(`Unsupported RPC protocol in network: ${networkName}`)
+
+    return false
+}
+
+const action: ActionType<TaskArguments> = async (taskArgs, hre) => {
+    printLogo()
+
+    const networks = hre.userConfig.networks || {}
+    const networkNames = Object.keys(networks)
```
janjakubnanista, line 135: This object might also contain `hardhat` and `localhost` which will in most cases not be available

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,161 @@
...
+        `========== Validating RPC URLs with ${taskArgs.timeout}ms timeout for networks: ${networkNames.join(', ')}`
+    )
+
+    for (const networkName of networkNames) {
+        const rpcUrl = networks[networkName]?.[RPC_URL_KEY]
+
+        if (rpcUrl && !(await validateRpcUrl(rpcUrl, taskArgs.timeout, networkName))) {
+            throw new Error(`${printBoolean(false)} RPC URL validation failed for network: ${networkName}`)
+        }
+    }
```
janjakubnanista, line 147: This will validate the URLs one by one - it might be worth using `Promise.all` instead to parallelize

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    logger.info(`${printBoolean(true)} All RPC URLs are valid!`)
+}
+task(
+    TASK_LZ_VALIDATE_RPCS,
+    'Validate RPC URLs in hardhat.config.ts. RPCs are only considered valid if they use the https or wss protocol and respond within the specified timeout.',
+    action
+).addParam(
+    'timeout',
+    `Maximum amount of time (in milliseconds) that the RPC URLs have to respond. If unspecified, default timeout will be used.`,
```
janjakubnanista, line 157: `If unspecified, default timeout of ${TIMEOUT}ms will be used.`

janjakubnanista (changes requested):
Just some small stuff to be fixed, the most annoying is the `hardhat`/`localhost` network thing because `hardhat` does not provide any utility to filter these out.

You can use `getNetworksByEid` to see which networks have `eid` defined because `hardhat` and `localhost` should never have it in the config. Or just filter them manually, see what works better

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
ravinagill15: Oh great catch! I will update the logic and cover this scenario with a test as well.

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
ravinagill15: Ah I see what you mean -- then the user will know all the bad URLs in one go instead of needing to re-run the task after each fix. I will look into this!

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
ravinagill15: I had previously added this but noticed that when we run the task with `--help`, any defaults are automatically printed beside each argument:

Screenshot 2024-08-30 at 10 24 57 AM

However, I understand where you are coming from and it is always good to be as clear as possible so I will add it to the description as well!

ravinagill15:
> You can use getNetworksByEid to see which networks have eid defined because hardhat and localhost should never have it in the config. Or just filter them manually, see what works better

If hardhat and localhost should never have `eid` defined, then I think it is best to filter out networks that do not have `eid` defined instead of manually filtering out networks with names `hardhat` or `localhost`. This is because the naming could potentially change, so checking `eid` is more robust imo. 

Will push this fix soon!

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        ws.onclose = () => {
+            cleanup(
+                false,
+                `Validation failed for ${rpcUrl} for network ${networkName}: WebSocket connection closed unexpectedly`
+            )
+        }
+    })
+}
```
janjakubnanista, line 119: You don't need to rewrite all this logic - it's all already wrapped in a `Provider`.

What you can do is just

```typescript
const providerFactory = createProviderFactory()
const provider = await providerFactory(eid)

// Now get a chain ID or a block number using the provider
```

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const validationPromises = networkNames.map(async (networkName) => {
+        const rpcUrl = networks[networkName]?.[RPC_URL_KEY]
+
+        if (rpcUrl && !(await validateRpcUrl(rpcUrl, taskArgs.timeout, networkName))) {
+            networksWithInvalidRPCs.push(networkName)
+        }
+    })
+
+    await Promise.all(validationPromises)
```
janjakubnanista, line 159:
```typescript
await Promise.all(Object.entries(eidByNetworkName).map(([networkName, eid]) => {
  if (eid == null) return
  
  const provider = await providerfactory(eid);
  
  // Now do your call, maybe with some timeout like so
  return Promise.race([
    provider.getBlock(),
    // We need to finally add this to some devtools package but for now
    new Promise<void>((, reject) => setTimeout(reject, 5000))
  ]).then(
    // Block call succeeded
    block => {}
    // Error was thrown, either a timeout or the block call failed
    () => {
      // Do your fail logic
      networksWithInvalidRPCs.push(networkName)
    }
}))
```

janjakubnanista (changes requested):

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
ravinagill15: Ah I see what you mean, this is much more simple so I will refactor HTTPS RPCs to be validated using the provider. 

However I noticed this does not work with websocker RPCs, so to validate websocker RPCs, I will use 

```
const wsProvider = new ethers.providers.WebSocketProvider('wss://ethereum-rpc.publicnode.com');
console.log("WSS PROVIDER: ", await wsProvider.getBlockNumber());
```

which works in a very similar way. Will push in next commit!

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
ravinagill15: Gotcha -- I will format the validation promises like this in the next commit, thanks!

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
```diff
@@ -0,0 +1,111 @@
...
+        logger.error(
+            `========== ${printBoolean(false)} RPC URL validation failed for network(s): ${networksWithInvalidRPCs.join(', ')}`
+        )
+    } else {
+        logger.info(`========== ${printBoolean(true)} All RPC URLs are valid!`)
+    }
+}
+task(
+    TASK_LZ_VALIDATE_RPCS,
+    'Validate RPC URLs in hardhat.config.ts. RPCs are only considered valid if they use the https or wss protocol and respond within the specified timeout.',
```
ravinagill15, line 103: I think we can have RPCs with other protocols, but chainlist seems to only support HTTPS and WSS

ravinagill15:
Looking into test failures!

packages/devtools-evm-hardhat/src/tasks/healthcheck/validate-rpcs.ts
ravinagill15: It seemed that `providerFactory` was not recognizing the networks in the provided config, so I used `ethers.JSONRpcProvider` for `https` urls as well

packages/devtools-evm-hardhat/package.json
```diff
@@ -66,6 +66,7 @@
     "@swc/core": "^1.4.0",
     "@swc/jest": "^0.2.36",
     "@types/jest": "^29.5.12",
+    "ethers": "^5.7.2",
```
janjakubnanista, line 69: Instead of `ethers` please use `@ethersproject/providers`, it's already in the dependencies

janjakubnanista (changes requested):
One tiny thing there

packages/devtools-evm-hardhat/package.json
ravinagill15: Ah great catch -- updated!

janjakubnanista (approved):

PR #877, opened 2024-08-30
implemented internal _send

ravinagill15:
## In this PR:
- Instead of changing `OFTCore.send` to be public so it can be called by `NativeOFTAdapter.send`, the body can be moved into an internal function so it can be called by both `OFTCore.send` and `NativeOFTAdapter.send`
- See conversation [here](https://github.com/LayerZero-Labs/devtools/pull/838#discussion_r1716766844)

### DIFF
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -75,7 +75,7 @@ abstract contract NativeOFTAdapter is OFTCore {
             revert InsufficientMessageValue(msg.value, requiredMsgValue);
         }
 
-        return super.send(_sendParam, _fee, _refundAddress);
+        return super._send(_sendParam, _fee, _refundAddress);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ravinagill15:
Will add changeset if this is merged after `feat/native-oft-adapte`

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -75,7 +75,7 @@ abstract contract NativeOFTAdapter is OFTCore {
             revert InsufficientMessageValue(msg.value, requiredMsgValue);
         }
 
-        return super.send(_sendParam, _fee, _refundAddress);
+        return super._send(_sendParam, _fee, _refundAddress);
```
ravinagill15, line 78: don't need `super`

ravinagill15: merge after fixed

TRileySchwarz (approved):

PR #878, opened 2024-08-30
Improve error messaging

ravinagill15:
## In this PR:
- Added call data to `CreditFailed` error to help with debugging 
- See conversation [here](https://github.com/LayerZero-Labs/devtools/pull/838#discussion_r1738895442)

### DIFF
--- a/packages/oft-evm/contracts/NativeOFTAdapter.sol
+++ b/packages/oft-evm/contracts/NativeOFTAdapter.sol
@@ -16,13 +16,8 @@ import { MessagingFee, MessagingReceipt, OFTCore, OFTReceipt, SendParam } from "
  */
 abstract contract NativeOFTAdapter is OFTCore {
 
-<<<<<<< HEAD
     error IncorrectMessageValue(uint256 provided, uint256 required);
-    error CreditFailed(address to, uint256 amountLD);
-=======
-    error InsufficientMessageValue(uint256 provided, uint256 required);
     error CreditFailed(address to, uint256 amountLD, bytes revertData);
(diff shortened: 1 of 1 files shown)


ravinagill15:
Will add changeset if this is merged after `feat/native-oft-adapter`

packages/oft-evm/contracts/NativeOFTAdapter.sol
```diff
@@ -112,7 +112,7 @@ abstract contract NativeOFTAdapter is OFTCore {
         // @dev Transfer tokens to the recipient.
         (bool success, ) = payable(_to).call{value: _amountLD}("");
         if (!success) {
-            revert CreditFailed();
+            revert CreditFailed(_to, _amountLD);
```
janjakubnanista, line 115: Oh in my comment I meant something like this:

```solidity
error CreditFailed(address to, bytes revertData);

// If the call fails, `data` will hold the revert reason
(bool success, bytes memory data) = payable(_to).call{value: _amountLD}("");
if (!success) revert CreditFailed(_to, data);
```

Thoughts and prayers please @TRileySchwarz

ravinagill15: Oh i see! That would definitely be helpful -- I will update this in the next commit 😄

ravinagill15: I kept the `to` and `amountLD` parameters bc I think they could be helpful as well and also added the `revertData` as a third parameter now

TRileySchwarz (approved):

PR #880, opened 2024-08-30
Updated tests

ravinagill15:
## In this PR:
- Added test case for when extra `msg.value` is provided (should revert)
- Updated `InsufficientMessageValue` error to `IncorrectMessageValue` since this error will be thrown even if there is extra `msg.value` provided
- Added comments to tests
- See conversation [here](https://github.com/LayerZero-Labs/devtools/pull/838#discussion_r1738903529)


ravinagill15:
Will add changeset if this is merged after `feat/native-oft-adapter`

PR #893, opened 2024-09-18
🪲broken solana tasks

ryandgoulding:


janjakubnanista (changes requested):
The limitation of the task overriding is that the newly added params cannot be added as required, that's why the build is failing

PR #894, opened 2024-09-18
🧹 Standalone CLI: Moving sign & send subtask to a generic function [1/N]

janjakubnanista:
### In this PR

- This is the first PR from a series focused on moving functionality from hardhat-specific packages into generic ones. Eventually, this functionality will support the standalone CLI under `@layerzerolabs/devtools-cli`
- A concept of _flow_ is introduced. Flows will be high-level functions that perform complex series of operations. Flows will also be composable just like the original hardhat tasks - e.g. the wire flow will use the sign and send flow to submit the transaction
- In this PR the sign & send subtask is gutted and its internals transplanted as a whole into `@layerzerolabs/devtools` as a flow. This flow is then used as the subtask itself - which gives us the great benefit of having the existing tests cover the functionality


ravinagill15 (approved):
Nice, it's awesome that you were able to implement flows in such a way that we can re-use existing tests! LGTM 🚀

ryandgoulding (approved):

PR #909, opened 2024-10-01
🧹 Move config loading into a flow

janjakubnanista:
### In this PR

- Continuing the transition towards a standalone CLI, this PR moves the config loading functionality into a flow function
  - The new flow function no longer loads and then transforms the graph. Instead, the transformation now needs to be encapsulated in the schema (in our case we use the `.transform` function on the schema to apply an asynchronous transformation)
- The existing functionality has not been removed (i.e. `createConfigLoader`), it has just been marked as deprecated and its tests have been reused for the new flow function
- `OmniGraphBuilderHardhat` has also been deprecated - as it should have been a long time ago - since what it does is not worth having an extra class there

### DIFF
--- a/packages/devtools/src/flows/config.load.ts
+++ b/packages/devtools/src/flows/config.load.ts
@@ -1,5 +1,5 @@
 import { createLogger, importDefault, isFile, isReadable, Logger, printZodErrors } from '@layerzerolabs/io-devtools'
-import type { ZodType, ZodTypeDef } from 'zod'
+import type { SafeParseReturnType, ZodType, ZodTypeDef } from 'zod'
 import type { OmniGraph } from '@/omnigraph'
 import { resolve } from 'path'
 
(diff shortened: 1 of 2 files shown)


.changeset/stupid-drinks-burn.md
```diff
@@ -0,0 +1,8 @@
+---
+"@layerzerolabs/ua-devtools-evm-hardhat": patch
+"@layerzerolabs/devtools-evm-hardhat": patch
+"@layerzerolabs/io-devtools": patch
+"@layerzerolabs/devtools": patch
+---
```
ravinagill15, line 6: I think we also need to include `@layerzerolabs/ua-devtools-evm-hardhat-test`

packages/devtools/src/flows/config.load.ts
```diff
@@ -0,0 +1,111 @@
...
+    //
+    // We do this so that we can prepend the error message with a more meaningful one
+    let configParseResult: SafeParseReturnType<unknown, TOmniGraph>
+    try {
+        configParseResult = await configSchema.safeParseAsync(rawConfigMaterialized)
+    } catch (error) {
+        throw new Error(`Config from file '${configPath}' is invalid: ${error}`)
+    }
+
+    if (configParseResult.success === false) {
```
ravinagill15, line 102: Should we write this as `!configParseResult.success` if the value can ever be `null`, `undefined`, or other falsy values?

ravinagill15 (commented):
Looks great!! Just a couple minor comments below

packages/devtools/src/flows/config.load.ts
janjakubnanista: Oh in this case it's always `true` or `false`. And TypeScript also has this quirk where `!` actually breaks (type narrowing)[https://www.typescriptlang.org/docs/handbook/2/narrowing.html].

```typescript
const a: boolean = false;

// If we check for true/false directly, type narrowing works 
if (a === false) return

// i.e. a will by typed as `true` now
const typeNarrowedA: true = a

// BUT if instead we check with `!`, type narrowing stops working
if (!a) return

// And this statement will not compile
const typeNarrowedA: true = a
```

.changeset/stupid-drinks-burn.md
janjakubnanista: That's a private package we don't publish - we can but don't need to add changesets since well they have no effect

ravinagill15 (approved):
🚀

PR #910, opened 2024-10-01
🧹 Move configuration execution into a flow

janjakubnanista:
### In this PR

- Move configuration execution subtask into a flow. This was the last subtask to be moved into a flow which means the wire task can now be moved to a flow and subsequently used by both the hardhat and the CLI implementations

### DIFF
--- a/.changeset/calm-spies-deny.md
+++ b/.changeset/calm-spies-deny.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/ua-devtools-evm-hardhat": patch
+---
+
+Move configuration execution into a flow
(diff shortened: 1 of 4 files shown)


.changeset/calm-spies-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@layerzerolabs/ua-devtools-evm-hardhat": patch
```
ravinagill15, line 2: I think we are missing devtools from this changeset

ravinagill15 (commented):
Looks good, just one minor comment!

.changeset/calm-spies-deny.md
janjakubnanista: Aaaand fixed

ravinagill15 (approved):

PR #914, opened 2024-10-08
🧹 Turn configLoadFlow into a factory

janjakubnanista:
### In this PR

- Turn `configLoadFlow` into `createConfigLoadFlow` so that some of the parameters are encapsulated in a closure. This is the last PR before the versioning PR can be unblocked and merged


ravinagill15 (approved):
:shipit:

PR #931, opened 2024-10-17
🪲 Use defaults for Solana RPC URLs

janjakubnanista:
### In this PR

- Use default Solana RPC URLs for tests for PRs coming from forks

### DIFF
--- a/packages/devtools-solana/test/common/accounts.test.ts
+++ b/packages/devtools-solana/test/common/accounts.test.ts
@@ -19,8 +19,8 @@ describe('common/accounts', () => {
     beforeAll(() => {
         connectionFactory = createConnectionFactory(
             createRpcUrlFactory({
-                [EndpointId.SOLANA_V2_MAINNET]: process.env.RPC_URL_SOLANA_MAINNET,
-                [EndpointId.SOLANA_V2_TESTNET]: process.env.RPC_URL_SOLANA_TESTNET,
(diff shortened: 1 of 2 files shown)


packages/devtools-solana/test/common/accounts.test.ts
```diff
@@ -19,8 +19,8 @@ describe('common/accounts', () => {
     beforeAll(() => {
         connectionFactory = createConnectionFactory(
             createRpcUrlFactory({
-                [EndpointId.SOLANA_V2_MAINNET]: process.env.RPC_URL_SOLANA_MAINNET,
-                [EndpointId.SOLANA_V2_TESTNET]: process.env.RPC_URL_SOLANA_TESTNET,
+                [EndpointId.SOLANA_V2_MAINNET]: process.env.RPC_URL_SOLANA_MAINNET || 'https://rpc.ankr.com/solana',
+                [EndpointId.SOLANA_V2_TESTNET]: process.env.RPC_URL_SOLANA_TESTNET || 'https://api.testnet.solana.com',
```
ryandgoulding, line 23: Wait wait wait, EndpointId.SOLANA_V2_TESTNET points to devnet.  I don't know who did that but we should probably honor the convention.

ryandgoulding (changes requested):

packages/devtools-solana/test/common/accounts.test.ts
janjakubnanista: Ah oke will change

ryandgoulding (approved):

PR #933, opened 2024-10-18
🗞️ support token2022 standard

ryandgoulding:
Also significantly cleans by factoring out common code, deleting unneeded code, and centralizing on UMI for transaction processing.


examples/oft-solana/tasks/evm/send.ts
```diff
@@ -38,17 +38,16 @@ const action: ActionType<TaskArguments> = async ({ dstEid, amount, to }, hre: Ha
...
-    console.dir(msgFee)
     const txResponse = await token.functions.send(sendParam, msgFee, signer.address, {
         value: msgFee.nativeFee,
         gasLimit: 500_000,
     })
     const txReceipt = await txResponse.wait()
     console.log(`send: ${amount} to ${to}: ${txReceipt.transactionHash}`)
+    console.log(
+        `Track cross-chain transfer here: ${getLayerZeroScanLink(txReceipt.transactionHash, dstEid == EndpointId.SOLANA_V2_TESTNET)}`
+    )
```
janjakubnanista, line 50: For later - we have the nicer loggers, we should replace these console.logs with those

examples/oft-solana/tasks/solana/createOFT.ts
```diff
@@ -1,147 +1,202 @@
...
-        )
-        console.log(`Accounts have been saved to ${outputDir}/OFT.json`)
-    })
+            console.log(`created SPL multisig @ ${mintAuthorityPublicKey.toBase58()}`)
+
+            const mint = isMABA
+                ? createNoopSigner(publicKey(mintStr))
+                : createSignerFromKeypair(umi, eddsa.generateKeypair())
+            const isTestnet = eid == EndpointId.SOLANA_V2_TESTNET
+            if (!isMABA) {
```
janjakubnanista, line 141: I'd rather see these if/elses being pulled into two separate tasks or task action implementations, the difference in the branch logic is quite large

janjakubnanista (approved):

examples/oft-solana/tasks/evm/send.ts
ryandgoulding: Good point. Will do in follow-up.

examples/oft-solana/tasks/solana/createOFT.ts
ryandgoulding: ditto

PR #937, opened 2024-10-21
🧹 bump "major" version (really minor) for Solana OFT

ryandgoulding:
Missed in the initial merge.  We don't publish this package anyway, so it shouldn't cause issues.

### DIFF
--- a/.changeset/curvy-vans-worry.md
+++ b/.changeset/curvy-vans-worry.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/oft-solana-example": patch
+---
+
+Bump for new Solana Implementation Version


Maliksb11:
ok

.changeset/curvy-vans-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@layerzerolabs/oft-solana-example": patch
```
ravinagill15, line 2: minor or patch? or doesn't matter?

ravinagill15 (approved):
LGTM 🚀

.changeset/curvy-vans-worry.md
ryandgoulding: I didn't even realize I had put `patch`, sorry you are right.

PR #940, opened 2024-10-22
🧹 Docker cache

janjakubnanista:
### In this PR

- Create cache artifacts in GHCR instead of the GHA cache. This means that the layers of all, including the intermediate, images will be uploaded to GHCR
- The cache will try to use the layers built for its branch will fallback to the `main` branch cache


ravinagill15 (approved):
🚀

PR #942, opened 2024-10-24
🪲 Fix an issue with outdated hardhat cache

janjakubnanista:
### In this PR

- Reload the whole hardhat deploy store if we get no deployments back. This fixes an issue currently present in Stargate
- Fixed a small issue with the test - a `spy` was initialized on the top of the test suite but was effectively erased in the `afterEach` hook

### DIFF
--- a/packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
+++ b/packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
@@ -63,15 +63,29 @@ export const createContractFactory = (environmentFactory = createGetHreByEid()):
             // with complete and partial ABIs
             //
             // To handle this case we'll merge the ABIs to make sure we have all the methods available
-            const deployments = await env.deployments.getDeploymentsFromAddress(address).catch(async () => {
-                // Hardhat deploy does not call its setup function when we call getDeploymentsFromAddress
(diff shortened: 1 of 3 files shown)


packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
```diff
@@ -63,15 +63,29 @@ export const createContractFactory = (environmentFactory = createGetHreByEid()):
...
+                                `Could not find a deployment for address '${address}' on ${networkLabel}`
+                            ),
+                            deployments
+                        )
+                )
+                .catch(async () => {
+                    // Hardhat deploy does not call its setup function when we call getDeploymentsFromAddress
+                    // so we need to force it to do so
+                    //
+                    // Since thee setup function is not available on the deployments extension, we need to trigger it indirectly
```
ravinagill15, line 84: small typo `thee` --> `the`

ravinagill15 (approved):
🚀

packages/devtools-evm-hardhat/src/omnigraph/coordinates.ts
janjakubnanista: I shall add this to my backlog since Stargate needs this A$AP

PR #944, opened 2024-10-24
🧹 Bump timeout on TON node tests

janjakubnanista:
### In this PR

- The TON local node tests depend on the quality of the machine that runs the node so we bump the timeout to remove any flakiness


ravinagill15 (approved):
:shipit:

PR #947, opened 2024-10-25
🪲 Avoid concurrent solana cache access in user tests

janjakubnanista:
### In this PR

- As of lately the concurrent nature of the user tests has been causing flaky user tests runs. This PR wraps the compilation step in an `flock` to avoid concurrent access of Solana's cache
- Another issue had to do with discrepancy between `anchor` versions available in the image and set in the Solana example. @St0rmBr3w is there a reason why I should stick to anchor `0.29.0` in the example? If there is I will downgrade the version we install in our base docker container


ryandgoulding (dismissed):

St0rmBr3w:
From my understanding, it's just been marked as a hard dependency by the HK team when deploying the old `oft-example`.

I know that for compiling at least, the solana version is a hard requirement otherwise the build results do not match as expected. Likely similar reasoning for why they capped the anchor version at `0.29` at least during project compilation.

cc: @HansonYip

janjakubnanista:
Closing since the parallel tests have been disabled

PR #957, opened 2024-10-28
🪲fix always compose_msg pattern

zlog-in:
When calling to `send` instruction without compose message, we have to use `composeMsg: undefined` rather than `composeMsg: Buffer.from('')`. 

The `Buffer.from('')` will be treated as `Some([])` rather than `None`, which results into the [`if` branch](https://github.com/LayerZero-Labs/devtools/blob/a396c205800f5b23a55ede98e67b9f56bf0d04ce/examples/oft-solana/programs/oft/src/msg_codec.rs#L13) for the compose message.

If we used `composeMsg: Buffer.from('')`, then there is always a suffix `0861db5f04e9bc6ff0c14b7e8b72c659b60e4a1a0bfc3029800a1987423df801` (encoded from `Some([])`) in the [message payload](https://testnet.layerzeroscan.com/tx/2FgTnRQPWq5HyhaEQzBAMi23uae3dKLAYE2pcQ1qh2a8sZ9DFLk28iZrtwJvAspyKAMiGNGJhw1fMNGjSNHCgC9h), and trigger a `ComposeSent` event and `LzComposeAlert` event on the destination chain.

### DIFF
--- a/.changeset/clean-stingrays-drop.md
+++ b/.changeset/clean-stingrays-drop.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/build-devtools": patch
+---
+
+Fix the default enabled value for declaration map plugin
--- a/.changeset/sharp-knives-jam.md
+++ b/.changeset/sharp-knives-jam.md
@@ -1,5 +1,0 @@
----
-"@layerzerolabs/oft-solana-example": patch
----
-
-Fix createOFT should allow 0 amount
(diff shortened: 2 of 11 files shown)


ryandgoulding:
@zlog-in any chance you can rebase?

zlog-in:
> @zlog-in any chance you can rebase?

is ok now for merge action?

ryandgoulding:
Rebased in #966

Issue #970, opened 2024-11-01
bug: lz:oapp:wire configures delegates first, which causes the rest of the transactions to fail


ryandgoulding:
**Describe the bug**
lz:oapp:wire will config delegates first, causing the rest of transactions to fail.

**To Reproduce**

1. Include delegate in layerzero.config.ts
2. Run `pnpm hardhat lz:oapp:wire --oapp-config <configFile>`

See failed transactions.

**Expected behavior**
Call `configureOAppDelegates` configurator last.

**Screenshots**
N/A

**Environment (please complete the following information):**

- OS: macOS
- OS Version Sonoma 14.7.1
- VM: node
- VM Version: All

**Additional context**
N/A

PR #971, opened 2024-11-01
🪲 call configureOAppDelegates configurator last

ryandgoulding:


ryandgoulding:
Resolves #970

janjakubnanista:
I think we might need to pull this logic out altogether into an ownership transfer task (or a separate task altogether) since some time ago we did move the opposite direction #687

janjakubnanista (approved):

Issue #978, opened 2024-11-06
Missing error signature in  ILayerZeroPriceFeed


0xdcota:
I appears that `PriceFeedMock.sol` in `LayerZero-Labs/devtools/packages/test-devtools-evm-foundry/contracts/mocks/PriceFeedMock.sol`

Is missing the error `LZ_PriceFeed_NotAnOPStack(uint32)` in the interface.

https://github.com/LayerZero-Labs/devtools/blob/36446d0eba516bed98748ea49009efc7b6dafec4/packages/test-devtools-evm-foundry/contracts/mocks/PriceFeedMock.sol#L184

Interface is defined in `LayerZero-Labs/LayerZero-v2/packages/layerzero-v2/evm/messagelib/contracts/interfaces/ILayerZeroPriceFeed.sol`

https://github.com/LayerZero-Labs/LayerZero-v2/blob/2d4d177d017d525bfe11892a98f2c511155d2160/packages/layerzero-v2/evm/messagelib/contracts/interfaces/ILayerZeroPriceFeed.sol#L37-L41

Or the dependencies are broken.

ryandgoulding:
I can't replicate this, and it seems as though it resolves just fine.  Please feel free to re-open if you continue to experience issues!

PR #981, opened 2024-11-07
🪲 Add missing error to PriceFeedMock

0xAkuti:
Adds missing error definition for the error `LZ_PriceFeed_NotAnOPStack` which is used in `PriceFeedMock._getL1LookupId()` and currently causing issues (https://github.com/LayerZero-Labs/devtools/issues/978).
```
Error: 
Compiler run failed:
Error (7576): Undeclared identifier.
   --> lib/devtools/packages/test-devtools-evm-foundry/contracts/mocks/PriceFeedMock.sol:184:16:
    |
184 |         revert LZ_PriceFeed_NotAnOPStack(l2Eid);
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^

```
The alternative would be to add it in `ILayerZeroPriceFeed` where the error `LZ_PriceFeed_UnknownL2Eid` which was previously used is defined, but since `LZ_PriceFeed_NotAnOPStack` seems to only be used in this mock and not in layerzero v2 adding it here seems to be the way to go.


0xAkuti:
Hi, any feedback on this? @janjakubnanista 
Would be nice to merge it since we currently use a forked repo with the fix and would like to use the official one.

janjakubnanista (changes requested):
Hey @0xAkuti! Could you please include a changeset? You just need to run `pnpm changeset` from the root of the repo and mark this package as `patch`, thanks!

If you're at it I would also update the branch on top of the newest `main` from devtools, we included some bug fixes that remedied flaky test runs

0xAkuti:
Hi @janjakubnanista, I added the changeset and updated to the current `main`

0xAkuti:
It was added last week in the layerzero-v2 repo (https://github.com/LayerZero-Labs/LayerZero-v2/pull/121) in `ILayerZeroPriceFeed` like I mentioned as the alternative approach in the top comment. So it's no longer needed here.

0xAkuti:
Resolved with https://github.com/LayerZero-Labs/LayerZero-v2/pull/121

PR #983, opened 2024-11-07
🗞️ solana config get

ryandgoulding:
This enables users to run:

```
pnpm hardhat lz:oapp:config:get --oapp-config ./layerzero.config.ts
```

For Solana OApps.  The concept of "default config" doesn't really exist for Solana, so `undefined` is used to represent these fields.

Future work would remove the `@layerzerolabs/ua-devtools-solana` dependency from `@layerzerolabs/ua-devtools-evm-hardhat` package.  However, since the task is already there, it is considered extraneous to this PR.


packages/ua-devtools-evm-hardhat/package.json
```diff
@@ -55,6 +55,7 @@
     "@layerzerolabs/protocol-devtools-evm": "~1.2.1",
     "@layerzerolabs/ua-devtools": "~1.0.5",
     "@layerzerolabs/ua-devtools-evm": "~3.0.1",
+    "@layerzerolabs/ua-devtools-solana": "~2.0.0",
```
janjakubnanista, line 58: This breaks the architecture and results in purely EVM projects to download all the Solana stuff

janjakubnanista (changes requested):

ryandgoulding:
replaced by #1003 which attains the same feature set without the architectural issues

PR #984, opened 2024-11-07
🧹 Update package versions

yargo13:
Updating package versions from `@layerzerolabs` to 3.0.12, also updating contract mocks accordingly


janjakubnanista (changes requested):
Please add a changeset - run `pnpm changeset` from the root of the project

janjakubnanista (approved):

PR #991, opened 2024-11-08
🧹  Updating OmniTransaction properties for logs

ravinagill15:
## In this PR:
- Added `metadata` optional property to `OmniTransaction` interface to support https://layerzerolabs.atlassian.net/browse/STAR2-1015
- Updated `formatOmniTransaction` to also print metadata
- Added optional `contractName` parameter for sdk constructors (see discussion [here](https://github.com/LayerZero-Labs/devtools/pull/991#discussion_r1838900890))

Example:
Screenshot 2024-11-08 at 10 38 49 AM

### DIFF
--- a/packages/devtools-evm/src/omnigraph/sdk.ts
+++ b/packages/devtools-evm/src/omnigraph/sdk.ts
@@ -44,6 +44,7 @@ export abstract class OmniSDK implements IOmniSDK {
 
     constructor(
         public readonly contract: OmniContract,
+        public readonly contractName: string = '',
         protected readonly logger: Logger = createModuleLogger(
             `EVM SDK ${new.target.name} @ ${formatOmniContract(contract)}`
         )
--- a/packages/devtools-evm/src/omnigraph/sdk.ts
+++ b/packages/devtools-evm/src/omnigraph/sdk.ts
@@ -44,7 +44,6 @@ export abstract class OmniSDK implements IOmniSDK {
 
     constructor(
         public readonly contract: OmniContract,
-        public readonly contractName: string = '',
         protected readonly logger: Logger = createModuleLogger(
             `EVM SDK ${new.target.name} @ ${formatOmniContract(contract)}`
         )
--- a/packages/devtools/src/transactions/format.ts
+++ b/packages/devtools/src/transactions/format.ts
@@ -8,6 +8,7 @@ export const formatOmniTransaction = (
     OmniAddress: transaction.point.address,
     OmniContract: transaction.point.contractName,
     'Function Name': transaction.functionName,
+    'Function Arguments': transaction.functionArgs,
     Description: transaction.description,
     Data: transaction.data,
     Value: transaction.value,
--- a/packages/devtools/src/transactions/format.ts
+++ b/packages/devtools/src/transactions/format.ts
@@ -3,9 +3,11 @@ import { formatEid } from '@/omnigraph/format'
 
 export const formatOmniTransaction = (
     transaction: OmniTransaction
-): Record<string, string | number | bigint | undefined> => ({
+): Record<string, string | number | bigint | undefined | null> => ({
     Endpoint: formatEid(transaction.point.eid),
     OmniAddress: transaction.point.address,
+    OmniContract: transaction.point.contractName,
+    'Function Name': transaction.functionName,
     Description: transaction.description,
     Data: transaction.data,
     Value: transaction.value,
--- a/packages/devtools/src/transactions/format.ts
+++ b/packages/devtools/src/transactions/format.ts
@@ -1,14 +1,14 @@
-import { OmniTransaction } from './types'
+import { OmniTransaction, OmniTransactionMetadata } from './types'
 import { formatEid } from '@/omnigraph/format'
 
 export const formatOmniTransaction = (
     transaction: OmniTransaction
-): Record<string, string | number | bigint | undefined | null> => ({
+): Record<string, string | number | bigint | undefined | null | OmniTransactionMetadata> => ({
     Endpoint: formatEid(transaction.point.eid),
     OmniAddress: transaction.point.address,
     OmniContract: transaction.point.contractName,
-    'Function Name': transaction.functionName,
-    'Function Arguments': transaction.functionArgs,
+    'Function Name': transaction.metadata?.functionName,
+    'Function Arguments': transaction.metadata?.functionArgs,
     Description: transaction.description,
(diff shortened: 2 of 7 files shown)


packages/devtools/src/transactions/format.ts
```diff
@@ -3,9 +3,11 @@ import { formatEid } from '@/omnigraph/format'
 
 export const formatOmniTransaction = (
     transaction: OmniTransaction
-): Record<string, string | number | bigint | undefined> => ({
+): Record<string, string | number | bigint | undefined | null> => ({
     Endpoint: formatEid(transaction.point.eid),
     OmniAddress: transaction.point.address,
+    OmniContract: transaction.point.contractName,
+    'Function Name': transaction.functionName,
```
ravinagill15, line 10: These are currently always printing empty, i will look into how to set these properties on the relevant transactions

packages/devtools/src/transactions/format.ts
```diff
@@ -3,9 +3,12 @@ import { formatEid } from '@/omnigraph/format'
 
 export const formatOmniTransaction = (
     transaction: OmniTransaction
-): Record<string, string | number | bigint | undefined> => ({
+): Record<string, string | number | bigint | undefined | null> => ({
     Endpoint: formatEid(transaction.point.eid),
     OmniAddress: transaction.point.address,
+    OmniContract: transaction.point.contractName,
```
ravinagill15, line 9: This is always `undefined` because the transaction is built with [devtools-evm createTransaction](https://github.com/LayerZero-Labs/devtools/blob/5b1586fbab8f6f39860849e31b2cb8274c7c483a/packages/devtools-evm/src/omnigraph/sdk.ts#L59-L68), which uses [omniContractToPoint](https://github.com/LayerZero-Labs/devtools/blob/5b1586fbab8f6f39860849e31b2cb8274c7c483a/packages/devtools-evm/src/omnigraph/coordinates.ts#L5-L8) to build the point.

However, as you can see, the `omniContractToPoint` function does not specify the `contractName` -- only the `eid` and `address`, which I think is the root of the issue.

I have been looking into how to retrieve the contract name, but had no luck so far. It seems i need to do a reverse lookup on the ens registrar contract or use an rpc provider to call `lookupAddress`, both of which I do not have access to in the `devtools-evm/src/omnigraph/coordinates` file for `omniContractToPoint`.

Is there another way I can get the contract name?

janjakubnanista: Oh yes [`omniContractToPoint`](https://github.com/LayerZero-Labs/devtools/blob/5b1586fbab8f6f39860849e31b2cb8274c7c483a/packages/devtools-evm/src/omnigraph/coordinates.ts#L5-L8)needs to be updated since the missing `contractName` property is a bug.

packages/devtools/src/transactions/format.ts
```diff
@@ -3,9 +3,12 @@ import { formatEid } from '@/omnigraph/format'
...
+    'Function Name': transaction.functionName,
+    'Function Arguments': transaction.functionArgs,
```
janjakubnanista, line 11: Rather than growing the `OmniTransaction` object, I'd create a new `OmniTransactionMetadata` interface like so:

```typescript
interface OmniTransactionMetadata {
  functionName?: string
  functionArgs?: unknown[]
}
```

Then you add a `metadata?: OmniTransactionMetadata` property to `OmniTransaction`. This way we keep the meta properties isolated.

janjakubnanista (changes requested):

packages/devtools/src/transactions/format.ts
ravinagill15: Ah great idea! On it 🫡

packages/devtools/src/transactions/format.ts
ravinagill15: I see, thank you for looking into this! @calebbanister please see Jan's explanation above.

While #869 is the long-term approach to fetch `contractName`, perhaps there is a quick "hacky" way to do it temporarily? It would really help when deploying/wiring stargate!

ravinagill15: @calebbanister mentioned we could:

> scrape the xxxxxxx-mainnet deployments folder and create a mapping of address -> contract name after splitting off the .json of the filenames

@janjakubnanista I can look into the above approach if you think there is no other "hacky way" to do it

janjakubnanista: Please no ad-hoc mappings no. The fastest hacky way is to look into the contract factories - the factories have access to the original `OmniPoint` that contains the `contractName`. You can add custom constructors to SDKs whose names you want to log and pass this `OmniPoint` or just the `contractName`.

packages/devtools-evm/src/omnigraph/sdk.ts
```diff
@@ -44,6 +44,7 @@ export abstract class OmniSDK implements IOmniSDK {
 
     constructor(
         public readonly contract: OmniContract,
+        public readonly contractName: string = '',
```
ravinagill15, line 47: @janjakubnanista adding this so that I can pass the contract name when calling the constructor [here](https://github.com/stargate-protocol/stargate-v2/blob/ff015c3b75968f155e62ff79d49a813fedc9819d/packages/stg-devtools-evm-hardhat-v2/src/asset/factory.ts#L17) for example

ravinagill15: ravina todo: test if `this.contractName` is accessible like `this.contract` is

ravinagill15:
Hmm I think the User Test failure might be legitimate considering it is failing even after rebasing with main -- looking into this!

tests-user/tests/create-lz-oapp.bats
```diff
@@ -247,6 +247,7 @@ teardown() {
 
     YARN_CACHE_FOLDER="/tmp/.yarn-cache-oft-solana" LZ_ENABLE_EXPERIMENTAL_SOLANA_OFT_EXAMPLE=1 npx --yes create-lz-oapp --ci --example oft-solana --destination $DESTINATION --package-manager yarn
     cd "$DESTINATION"
+    yarn install
```
ryandgoulding, line 250: Definitely don't want to do this.

packages/devtools-evm/src/omnigraph/sdk.ts
ravinagill15: remove this change

janjakubnanista (approved):

PR #1014, opened 2024-11-12
🗞️ Aptos v2 examples

AlexanderLiteplo:
adding aptos:, oapp, oft, oft-adapter to examples


AlexanderLiteplo:
Keeping this as a dev branch until cli complete

PR #1022, opened 2024-11-14
🪲 squads multisig support

ryandgoulding:
"admin" is derived correctly.

### DIFF
--- a/packages/ua-devtools-solana/src/oft/sdk.ts
+++ b/packages/ua-devtools-solana/src/oft/sdk.ts
@@ -414,7 +414,7 @@ export class OFT extends OmniSDK implements IOApp {
                 this._umiToWeb3Tx([
                     oft.initConfig(
                         {
-                            admin: await this._getOwnerSigner(),
+                            admin: await this._getAdmin(),
                             oftStore: this.umiPublicKey,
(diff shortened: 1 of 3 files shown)


packages/ua-devtools-solana/src/oft/sdk.ts
```diff
@@ -574,11 +574,8 @@ export class OFT extends OmniSDK implements IOApp {
         return web3Transaction
     }
 
-    protected async _getOwnerSigner(): Promise<Signer> {
+    protected async _getAdmin(): Promise<Signer> {
         const owner = await this.getOwner()
-        if (!owner) {
-            throw new Error(`No owner found for ${this.label}`)
-        }
         return createNoopSigner(publicKey(owner))
```
ravinagill15, line 579: Would `createNoopSigner(publicKey(owner))` throw an error if `owner` was falsey since we are removing the validation check?

ravinagill15 (approved):
Had a question, otherwise LGTM 🚀

packages/ua-devtools-solana/src/oft/sdk.ts
ryandgoulding: Good question.  It must exist in the contract.  If falsey is returned, we have bad info.

PR #1039, opened 2024-11-18
🗞️ Created `packages/oapp-alt-evm` and added `EndpointV2Alt` to TestHelper

shankars99:
OAppAlt is OApp but with alt-endpoint support. This contains a different OAppSender as the fee is paid in ERC20 instead of native token. 

TestHelper also did not have support for EndpointV2Alt. Added a function that spins up endpoints with erc20s as the payment token that does not affect the current implementations of TestHelper testing.

Added testing for OAppAlt.


.changeset/smart-plants-hide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@layerzerolabs/test-devtools-evm-foundry": major
+"@layerzerolabs/oapp-alt-evm": major
```
janjakubnanista, line 3: This will immediatelly bump the version to `1.0.0` which is not what we want I believe

janjakubnanista (changes requested):

.changeset/smart-plants-hide.md
shankars99: just added a minor version for devtools

packages/oapp-alt-evm/turbo.json
janjakubnanista, line 1: Please delete - this is a mistake that is being copied over and over, I am just creating a PR to solve this.

packages/oapp-alt-evm/package.json
```diff
@@ -0,0 +1,60 @@
...
+      "imports": "./artifacts/*.json"
+    }
+  },
+  "files": [
+    "artifacts/**/*",
+    "contracts/**/*",
+    "test/**/*"
+  ],
+  "scripts": {
+    "build": "$npm_execpath compile",
```
janjakubnanista, line 31: Please delete the `build` command: 

- `build` command builds JS artifacts and depends on `compile` (see `turbo.json`)
- `compile` command does any contract/program compilation

packages/oapp-alt-evm/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "clean": "rm -rf .turbo cache artifacts",
+    "compile": "$npm_execpath forge build",
+    "test": "$npm_execpath forge test"
```
janjakubnanista, line 34: Please remove the `$npm_execpath`, `forge` is a standalone binary

janjakubnanista (changes requested):

PR #1041, opened 2024-11-19
🗞️ output transaction data to JSON file

ryandgoulding:


packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
```diff
@@ -96,13 +102,18 @@ const action: ActionType<TaskArgs> = async (
         logger,
         // We use hardhat subtasks to provide the option to override certain behaviors on a more granular level
         executeConfig: ({ graph }) => hre.run(configureSubtask, { graph } satisfies SubtaskConfigureTaskArgs),
-        signAndSend: ({ transactions }) =>
-            hre.run(signAndSendSubtask, {
+        signAndSend: ({ transactions }) => {
+            if (outputFilename) {
+                logger.debug(`Writing transactions to ${outputFilename}`)
+                writeFileSync(outputFilename, JSON.stringify(transactions, null, 2))
```
ravinagill15, line 108: Is it worth putting this within a try-catch in case it times out or throws some other exception?

ravinagill15 (approved):
Very clean approach 🚀

packages/ua-devtools-evm-hardhat/src/tasks/oapp/wire/index.ts
ryandgoulding: Nah, I want the program to completely blow up if it can't write the file.

PR #1049, opened 2024-11-19
🧹 Use concurrently for parallel compilation

janjakubnanista:
### In this PR

- Use `concurrently` to parallelize the `compile` scripts in examples
- The shortcut `pnpm:compile:*` version of concurrently command cannot be used since examples might use different package managers

### DIFF
--- a/examples/oapp/package.json
+++ b/examples/oapp/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "clean": "rm -rf artifacts cache out",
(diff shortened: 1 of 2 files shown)


examples/oapp/package.json
```diff
@@ -43,6 +43,7 @@
     "@types/mocha": "^10.0.6",
     "@types/node": "~18.18.14",
     "chai": "^4.4.1",
+    "concurrently": "~9.1.0",
```
yargo13, line 46: You didn't update the compile on this file?

janjakubnanista: Done

PR #1067, opened 2024-11-21
chore: rearrange `create-lz-oapp` selections

St0rmBr3w:
Change the selection options available via feature flag and exposed by default within create-lz-oapp


ryandgoulding (approved):
lgtm

PR #1072, opened 2024-11-22
docs(examples/oft-solana): elaborate

nazreen:
- link to official Solana setup (Rust, Solana, Anchor)
- Solana private key instructions
- TESTNET naming clarification
- deploy command correction: `mainnet-beta` to `devnet`

### DIFF
--- a/examples/oft-solana/README.md
+++ b/examples/oft-solana/README.md
@@ -87,7 +87,7 @@ pnpm hardhat lz:oft:solana:create --eid 40168 --program-id <PROGRAM_ID>
 ```
 
 :important: You may specify the `--additional-minters` flag to add a CSV of additional minter keys to the mint
-multisig. If you do not want to, you must sepcify `--only-oft-store`. If you choose the latter approach, you can never
+multisig. If you do not want to, you must specify `--only-oft-store true`. If you choose the latter approach, you can never
 substitute in a different mint authority.
 
 For OFTAdapter:
@@ -103,7 +103,7 @@ pnpm hardhat lz:oft:solana:create --eid 40168 --program-id <PROGRAM_ID> --mint <
 ```
 
 :important: You may specify the `--additional-minters` flag to add a CSV of additional minter keys to the mint
-multisig. If you do not want to, you must sepcify `--only-oft-store`. If you choose the latter approach, you can never
+multisig. If you do not want to, you must specify `--only-oft-store store`. If you choose the latter approach, you can never
 substitute in a different mint authority.
 
 Make sure to update [layerzero.config.ts](./layerzero.config.ts) and set `solanaContract.address` with the `oftStore` address.
--- a/examples/oft-solana/README.md
+++ b/examples/oft-solana/README.md
@@ -12,13 +12,11 @@
 
 ## Setup
 
+Ensure that you have Rust, Solana and Anchor installed. If not, follow the instructions in the [official Solana guide](https://solana.com/docs/intro/installation#install-dependencies). On the page as well are instructions on creating a wallet and getting devnet SOL.
+
 We recommend using `pnpm` as a package manager (but you can of course use a package manager of your choice).
 Docker is required to build using anchor. We highly recommend that you use the most up-to-date Docker version to avoid any issues with anchor
 builds.
-
-Ensure that you have Rust, Solana and Anchor installed. If not, follow the instructions in the [official Solana guide](https://solana.com/docs/intro/installation#install-dependencies). On the page as well are instructions on creating a wallet and getting devnet SOL.
-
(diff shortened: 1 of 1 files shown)


examples/oft-solana/README.md
```diff
@@ -12,6 +12,8 @@
 
 ## Setup
 
+Ensure that you have Rust, Solana and Anchor installed. If not, follow the instructions in the [official Solana guide](https://solana.com/docs/intro/installation#install-dependencies). On the page as well are instructions on creating a wallet and getting devnet SOL.
```
St0rmBr3w, line 15: Add a trailing comma after Solana,

examples/oft-solana/README.md
```diff
@@ -12,6 +12,8 @@
...
+
 We recommend using `pnpm` as a package manager (but you can of course use a package manager of your choice).
 Docker is required to build using anchor. We highly recommend that you use the most up-to-date Docker version to avoid any issues with anchor
```
St0rmBr3w, line 18: Perhaps add a link to Docker as well?

examples/oft-solana/README.md
```diff
@@ -34,6 +36,41 @@ pnpm install
 pnpm test
 ```
 
+### Get Devnet SOL
+
+```bash
+solana airdrop 5 -u devnet
+```
+
+We recommend that you request 5 devnet SOL, which should be sufficient for this walkthrough. For the example here, we will be using Solana Devnet. If you are hitting rate limits after, you can also use the [official Solana faucet](https://faucet.solana.com/).
```
St0rmBr3w, line 45: Replace with "If you hit RPC rate limits after"

examples/oft-solana/README.md
```diff
@@ -34,6 +36,41 @@ pnpm install
...
+
+const keypairFilePath = `<KEYPAIR_FILE_PATH_HERE>`; // you can view this by running `solana config get`
+
+const data = fs.readFileSync(keypairFilePath, "utf8");
+const keypairJson = JSON.parse(data);
+const keypair = Keypair.fromSecretKey(Uint8Array.from(keypairJson));
+const base58EncodedPrivateKey = bs58.encode(keypair.secretKey);
+
+console.log(base58EncodedPrivateKey);
+```
```
St0rmBr3w, line 68: Maybe instead we can just add this as a hardhat task?

Not a problem to have this conversion referenced for now, but if we know the exact behaviour we can always add it directly.

Alternatively, @ryandgoulding, I think that this marks an important problem with our Solana DevEx.

We should instead of using the bs58.encode privateKey, use the JSON array format of bs58. The solana-cli uses this format as well, and I think is probably more approachable and consistent with the rest of the solana devex versus having them copy and paste the encoded private key.

examples/oft-solana/README.md
```diff
@@ -34,6 +36,41 @@ pnpm install
...
+
+Then, run `node getBase58Pk.js`
+
+Also set the `RPC_URL_SOLANA_TESTNET` value. Note that while the naming used here is `TESTNET`, it refers to the Solana Devnet. We use `TESTNET` to keep it consistent with the existing EVM testnets.
+
```
St0rmBr3w, line 73: You can link to the Solana Devnet link under "Chains" in the documentation here.

examples/oft-solana/README.md
```diff
@@ -85,7 +124,7 @@ pnpm hardhat lz:oft:solana:create --eid 40168 --program-id <PROGRAM_ID>
 ```
 
 :important: You may specify the `--additional-minters` flag to add a CSV of additional minter keys to the mint
```
St0rmBr3w, line 126: Can you fix these :important: tags? We've left them alone for a while, but they deserve to properly flag.

St0rmBr3w (changes requested):

examples/oft-solana/README.md
nazreen: Can you clarify what needs to be replaced ? The whole set of lines from 42 - 45 ? Or do you mean just the first 2 sentences ?

examples/oft-solana/README.md
nazreen: I vote for the second option. both Solana CLI and Anchor (defaults to) using the keypair in the default location of `~/.config/solana/id.json`. Would be good for us to do the same. Then devs won't need to even bother with this conversion from uint array to base58. We should of course provide the option to reference a different keypair file, just like Anchor.

examples/oft-solana/README.md
nazreen: will do

St0rmBr3w (dismissed):

examples/oft-solana/README.md
nazreen: or maybe just "If you are hitting rate limits after" yea ? I will assume this and push this commit now.

St0rmBr3w (approved):

PR #1073, opened 2024-11-22
docs: amend comment for quoteOFT

nazreen:
The current wording for notice comment of `quoteOFT` uses the term `quote` which in LZ's context, could have readers expect its return value(s) needing to be used directly as a msg.value.


packages/oft-evm/contracts/interfaces/IOFT.sol
```diff
@@ -104,7 +104,7 @@ interface IOFT {
     function sharedDecimals() external view returns (uint8);
 
     /**
-     * @notice Provides a quote for OFT-related operations.
+     * @notice Provides the fee breakdown and settings data for an OFT.
```
St0rmBr3w, line 107: Can you add to this notice that it is unused in the default implementation?

PR #1075, opened 2024-11-25
docs(oft-solana): various improvements

nazreen:
- use collapsible for script that gets solana private key in base58
- oftStore address instruction as own heading
- mention options
- fix typo in log message when deploying
- explainer on enforcedOptions and extraOptions

PR #1079, opened 2024-11-25
fix default layerzero.config.ts

zhangchitc:
- following oft-solana README, after setup both Sepolia and Solana contract, when trying to send OFT from Sepolia to Solana devnet using `npx hardhat send`, the sending tx will succeed, but eventually receiver side fail: https://testnet.layerzeroscan.com/tx/0x92d52ad346cae181948d2c64d0e156764ee58f5f0ae4a63c421fc1a4fe97425c
- the reason is that we neither specify gas config in `layerzero.config.ts` nor in task scripts
Xnip2024-11-25_23-14-54
Xnip2024-11-25_23-14-22
- so I think probably the clean way to fix this is just restore the default config

@nazreen @St0rmBr3w @DanL0

### DIFF
--- a/examples/oft-solana/layerzero.config.ts
+++ b/examples/oft-solana/layerzero.config.ts
@@ -1,4 +1,5 @@
 import { EndpointId } from '@layerzerolabs/lz-definitions'
+import { ExecutorOptionType } from '@layerzerolabs/lz-v2-utilities'
 
 import type { OAppOmniGraphHardhat, OmniPointHardhat } from '@layerzerolabs/toolbox-hardhat'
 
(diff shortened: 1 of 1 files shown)


nazreen:
thanks so much @zhangchitc! We realized that we were missing guidance on this. This is covered in #1075. Instead of giving the enforcedOptions directly, we want the builder to make the conscious choice of either using enforcedOptions or passing in the options when calling send(). This will now be covered in the README.

zhangchitc:
@nazreen cool, would be nice to get #1075 merged soon, as currently the solana example feels broken to me and I spent entire day debugging this...

nazreen:
@zhangchitc will try to get it approved and merged ASAP. and yes, there's a lot of room for improvement on the Solana examples. We're improving it every day now. Several PR's are now awaiting review. Thank you for your patience.

PR #1080, opened 2024-11-25
🗞️ Add Aptos CLI example directories and stubs

ryandgoulding:
### In this PR:

- added stubs for all relevant aptos-examples, including adding them to `create-lz-oapp` under the flag `LZ_ENABLE_EXPERIMENTAL_APTOS_EXAMPLES`: 

```typescript
// ./packages/create-lz-oapp/src/config.ts
...(process.env.LZ_ENABLE_EXPERIMENTAL_APTOS_EXAMPLES
            ? [
                  {
                      id: 'oft-aptos',
                      label: 'OFT (Aptos Fungible Asset)',
                      repository,
                      directory: 'examples/oft-aptos',
                      ref,
                  },
                  {
                      id: 'oft-adapter-aptos',
                      label: 'OFT Adapter (Aptos Fungible Asset)',
                      repository,
                      directory: 'examples/oft-adapter-aptos',
                      ref,
                  },
                  {
                      id: 'oft-adapter-aptos-coin',
                      label: 'OFT Adapter (Aptos Coin)',
                      repository,
                      directory: 'examples/oft-adapter-aptos-coin',
                      ref,
                  },
                  {
                      id: 'oft-aptos-coin',
                      label: 'OFT (Aptos Coin)',
                      repository,
                      directory: 'examples/oft-aptos-coin',
                      ref,
                  },
              ]
            : []),
```

This PR should be used for shipping the final version of the Aptos CLI SDK.

#### To-Do:

- update the ./sources in each example to be latest

- add @shankars99 and @AlexanderLiteplo SDK to each example


St0rmBr3w (changes requested):
@AlexanderLiteplo did you mean to push these core dependencies up?

packages/devtools-move/cli/operations/evm-wire.ts
```diff
@@ -0,0 +1,16 @@
+import { INewOperation } from '@layerzerolabs/devtools-extensible-cli'
+import { wireEvm } from '../../tasks/evm/wire-evm'
+
+class EVMWireOperation implements INewOperation {
+    vm = 'evm'
+    operation = 'wire'
+    description = 'Wire EVM contracts'
+    reqArgs = ['lz_config']
```
ryandgoulding, line 8: Why `--lz-config` instead of `--oapp-config` like everywhere else?

packages/devtools-extensible-cli/cli/AptosEVMCli.ts
```diff
@@ -0,0 +1,85 @@
+import { INewOperation, Operation } from './types/NewOperation'
+import { ArgumentParser } from 'argparse'
+import { initOperation } from './types/help'
+
+class AptosEVMCLI {
+    parser: ArgumentParser
+    args: any
```
ryandgoulding, line 7: Any time I see `any` in typescript I get a little nervous.  Is there any way to restrict the typing on this?

packages/devtools-move/sdk/endpoint.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    async getDefaultSendLibrary(eid: EndpointId): Promise<string> {
+        const result = await this.aptos.view({
+            payload: {
+                function: `${this.endpoint_address}::endpoint::get_default_send_lib`,
+                functionArguments: [eid],
+            },
+        })
+        return result[0] as string
```
ryandgoulding, line 25: What if the send library is not defined (i.e., result.length = 0)?

packages/devtools-move/sdk/endpoint.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    async getSendLibrary(oftAddress: string, dstEid: number): Promise<[string, boolean]> {
+        try {
+            const result = await this.aptos.view({
+                payload: {
+                    function: `${this.endpoint_address}::endpoint::get_effective_send_library`,
+                    functionArguments: [oftAddress, dstEid],
+                },
+            })
+            return [result[0] as string, result[1] as boolean]
```
ryandgoulding, line 36: similar to above comment

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
```diff
@@ -0,0 +1,47 @@
+// todo: implement connection builder that reads form yaml and fills in config with movement stuff if needed
```
ryandgoulding, line 1: Why yaml?

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
```diff
@@ -0,0 +1,47 @@
...
+                faucet: faucet,
+                indexer: indexer,
+            })
+        )
+    } else {
+        return new Aptos(new AptosConfig({ network: network }))
+    }
+}
+
+export function getChain(fullnode: string): string {
```
ryandgoulding, line 33: What if a non-sensical string is used for `fullnode`?  Costs very little to be careful.

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
```diff
@@ -0,0 +1,47 @@
+// todo: implement connection builder that reads form yaml and fills in config with movement stuff if needed
+// todo replace all mentions of aptos with move-vm instead
+
+import { Aptos, AptosConfig, Network } from '@aptos-labs/ts-sdk'
+
+const CHAIN_MOVEMENT = 'movement'
+const CHAIN_APTOS = 'aptos'
+
+const MOVEMENT_INDEXER_URLS = {
+    [Network.TESTNET]: 'https://indexer.testnet.porto.movementnetwork.xyz/v1/graphql',
```
ryandgoulding, line 10: Why are we relying on an indexer here?  I am concerned that we may overburden the endpoint.

packages/devtools-move/sdk/msgLib.ts
```diff
@@ -0,0 +1,122 @@
...
+    use_default_for_optional_dvns: true,
+    use_default_for_required_dvns: true,
+}
+
+const DEFAULT_EXECUTOR_CONFIG: ExecutorConfig = {
+    executor_address: '',
+    max_message_size: -1,
+}
+
+export class MsgLib {
```
ryandgoulding, line 44: Why did you go with your own abstraction rather than re-use the one already in monorepo?

https://github.com/LayerZero-Labs/monorepo/blob/main/packages/layerzero-v2/aptos/movevm-sdk/src/modules/uln302.ts

packages/devtools-move/sdk/oapp.ts
```diff
@@ -0,0 +1,277 @@
...
+    PrivateKey,
+    PrivateKeyVariants,
+    SimpleTransaction,
+} from '@aptos-labs/ts-sdk'
+
+import { EndpointId } from '@layerzerolabs/lz-definitions'
+
+import { hexAddrToAptosBytesAddr } from './utils'
+
+export class OAPP {
```
ryandgoulding, line 15: NIT:  I would prefer `OApp`, but not a big deal.

packages/devtools-move/tasks/evm/utils/types.ts
```diff
@@ -0,0 +1,98 @@
...
+    expiry: bigint
+}
+
+export type SetConfigParam = {
+    eid: eid
+    configType: number
+    config: string
+}
+
+export const ZEROADDRESS_EVM = '0x0000000000000000000000000000000000000000'
```
ryandgoulding, line 76: You are already importing ethers, so why not use https://github.com/ethers-io/ethers.js/blob/v5.7/packages/constants/src.ts/addresses.ts#L1?

packages/devtools-move/tasks/evm/wire-evm.ts
```diff
@@ -0,0 +1,152 @@
...
+import { createSetSendConfigTransactions } from './wire/setSendConfig'
+import { createSetSendLibraryTransactions } from './wire/setSendLibrary'
+import { executeTransactions } from './wire/transactionExecutor'
+
+import type { ContractMetadataMapping, NonEvmOAppMetadata, TxEidMapping } from './utils/types'
+import path from 'path'
+import dotenv from 'dotenv'
+
+/**
+ * @description Handles wiring of EVM contracts with the Aptos OApp
```
ryandgoulding, line 26: So you recreated the EVM transaction generation for only the EVM <> Aptos paths?  My concern is what about other VMs, such as Solana?  At least a few of our day 1 launch partners already have Solana deployments.

packages/devtools-move/tasks/evm/wire-evm.ts
```diff
@@ -0,0 +1,152 @@
...
+ */
+async function wireEvm(args: any) {
+    const envPath = path.resolve(path.join(args.rootDir, '.env'))
+    const env = dotenv.config({ path: envPath })
+    if (!env.parsed || env.error?.message !== undefined) {
+        console.error('Failed to load .env file.')
+        process.exit(1)
+    }
+
+    const privateKey = env.parsed.EVM_PRIVATE_KEY
```
ryandgoulding, line 37: What about a mnemonic?  The rest of our configs will use either mnemonic or pk, so I just want to prevent confusion if at all possible.

packages/devtools-move/tasks/evm/wire-evm.ts
```diff
@@ -0,0 +1,152 @@
...
+     */
+    TxTypeEidMapping.setPeer = await createSetPeerTransactions(contractMetaData, nonEvmOapp)
+    TxTypeEidMapping.setDelegate = await createSetDelegateTransactions(contractMetaData, nonEvmOapp)
+    TxTypeEidMapping.setEnforcedOptions = await createSetEnforcedOptionsTransactions(contractMetaData, nonEvmOapp)
+    TxTypeEidMapping.setSendLibrary = await createSetSendLibraryTransactions(contractMetaData, nonEvmOapp)
+    TxTypeEidMapping.setReceiveLibrary = await createSetReceiveLibraryTransactions(contractMetaData, nonEvmOapp)
+    TxTypeEidMapping.sendConfig = await createSetSendConfigTransactions(contractMetaData, nonEvmOapp)
+    TxTypeEidMapping.receiveConfig = await createSetReceiveConfigTransactions(contractMetaData, nonEvmOapp)
+
+    // TxTypeEidMapping.setReceiveLibraryTimeout = await createSetReceiveLibraryTimeoutTransactions(
```
ryandgoulding, line 126: Any reason this is commented out?

packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
```diff
@@ -0,0 +1,125 @@
...
+         *  - forkUrl = rpcUrl => mainnet mode
+         *
+         * The mapping helps keep track of the different chains and their respective providers
+         * We also create a signer object using the private key and provider
+         */
+        const newProvider = new providers.JsonRpcProvider(rpcUrlsMap[eid])
+        const signer = new ethers.Wallet(privateKey, newProvider)
+
+        accountEidMap[eid] = {
+            nonce: await signer.getTransactionCount(),
```
ryandgoulding, line 70: This is prone to nonce conflict if there is more than one transaction for a given eid.  If this is not a realistic use case, then no worries.  Otherwise, consider reading it into a map and incrementing as you go.

packages/devtools-move/tasks/move/build.ts
```diff
@@ -0,0 +1,85 @@
...
+async function buildMovementContracts(named_addresses: string, configPath: string) {
+    const aptosYamlConfig = await parseYaml(configPath)
+    const network = aptosYamlConfig.network
+    const lzNetworkStage = getLzNetworkStage(network)
+
+    // Get additional named addresses and combine with provided ones
+    const additionalAddresses = getNamedAddresses(lzNetworkStage)
+    const namedAddresses = named_addresses ? `${named_addresses},${additionalAddresses}` : additionalAddresses
+
+    const cmd = 'aptos'
```
ryandgoulding, line 24: Is there a specific assumed aptos version?  Should we check for that version?

packages/oft-evm/artifacts/OFT.sol/OFT.json
```diff
@@ -1064,7 +1064,7 @@
     "transferFrom(address,address,uint256)": "23b872dd",
```
ryandgoulding, line 1064: Why did this file change?  Probably should not have?

examples/oft-move-adapter/README.md
```diff
@@ -0,0 +1,127 @@
...
+Note: to overwrite previous deploy and build, you can use `--force-build true` for the build script and `--force-deploy true` for the deploy script.
+
+### Builds the contracts
+
+```bash
+pnpm run lz:sdk:move:build --lz-config move.layerzero.config.ts --named-addresses oft=$ACCOUNT_ADDRESS,oft_admin=$ACCOUNT_ADDRESS
+```
+
+### Checks for build, builds if not, then deploys the contracts, sets the delegate and initializes
+First modify deploy-move/OFTAdpaterInitParams.ts and replace the oftMetadata with your desired values:
```
DanL0, line 56: typo - OFTAdpater

examples/oft-move-adapter/README.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+## EVM Deployment
+
+```bash
+npx hardhat lz:deploy
+```
+
+Select only the evm networks (DO NOT SELECT APTOS)
+For deploy script tags use:
```
DanL0, line 127: missing content, doesn't say what to use

examples/oft-move-adapter/contracts/mocks/MyOFTMock.sol
DanL0, line 1: MyOFTMock is unnecessary in the oft-move-adapter because OFT liquidity should come from Aptos side (because this is Adapter example, token already exists on A[tps). Therefore minting any EVM OFTs supply in any setup with adapter isn't recommended and might encourage users to follow incorrect flow.

examples/oft-move-adapter/deploy/MyEVMOFT.ts
```diff
@@ -0,0 +1,53 @@
+import assert from 'assert'
+
+import { type DeployFunction } from 'hardhat-deploy/types'
+
+const contractName = 'MyOFT'
```
DanL0, line 5: another reason why MyOFTMock should be removed - it's not even used - here MyOFT is used not MyOFTMock

and that's correct in deploy script it should be MyOFT that's deployed not the mock, mock should be removed

packages/devtools-extensible-cli/cli/AptosEVMCli.ts
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        if (NewOperation.addArgs) {
+            for (const arg of NewOperation.addArgs) {
+                this.parser.add_argument(arg.name, arg.arg)
+            }
+        }
+    }
+
+    async execute(_callFromInheritance: boolean = false) {
```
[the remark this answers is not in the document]
AlexanderLiteplo: @shankars99

packages/devtools-extensible-cli/cli/AptosEVMCli.ts
AlexanderLiteplo: @shankars99

examples/oft-move-adapter/README.md
AlexanderLiteplo: fixed in a new update! good catch though :)

examples/oft-move-adapter/contracts/mocks/MyOFTMock.sol
AlexanderLiteplo: good point!

AlexanderLiteplo: @St0rmBr3w can you confirm?

St0rmBr3w: This is somewhat true, although in practice no one should deploy the MyOFTMock in general on mainnet. 

But yes, you should flag in README that the Aptos Adapter is meant for tokens that DO NOT already exist on EVM.

Liquidity should only come from Aptos outbound to OFTs. No existing liquidity should already exist on other EVMs.

packages/devtools-extensible-cli/cli/AptosEVMCli.ts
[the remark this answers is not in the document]
shankars99: that was a used variable but isnt now. removed it

packages/devtools-move/cli/operations/evm-wire.ts
shankars99: just a variable - changing it

packages/devtools-extensible-cli/cli/AptosEVMCli.ts
shankars99: `args: any` => `(method) ArgumentParser.parse_args(args?: string[], ns?: Namespace | object): any`  
is generated by the parser which loads an object with all the flags that are passed via the command line into it - these flags are defined in the sdk's operations and every operation like wire have required flags that are checked - and since they are asserted then we can assume they exist in the any  

```
this.validateArgs(exec_op.requiredArgs)
    for (const arg of args) {
            if (!this.args[arg]) {
            // parser loads flags with `_` instead of `-` so: `--oapp-config` would be loaded as `oapp_config`
                missingArgs.push(`--${arg.replace('_', '-')}=<value>`) 
                isValid = false
            }
        }
        if (!isValid) {
            throw new Error(`The following args are required: ${missingArgs.join(', ')}\n\n`)
        }
```
    
and you pass in the list of required args via 

```    
    class EVMWireOperation implements INewOperation {
        vm = 'evm'
        operation = 'wire'
        description = 'Wire EVM contracts'
        reqArgs = ['oapp_config']
    
        async impl(args: any): Promise<void> {
            await wireEvm(args)
        }
}
```

and you also specify what the arg is 

```
{
addArgs = [
      {
            name: '--oapp-config',
            arg: {
                help: 'path to the layerzeroconfig file',
                required: false,
            },
      },
     ...
]
```

packages/devtools-move/tasks/evm/utils/types.ts
shankars99: bruh.mp4

lol yeah changing that did not know this existed

packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
shankars99: yep this is handled in line 101 of the same file - will add a comment later at line 70 about this

`tx.nonce = accountEidMap[eid].nonce++`

packages/devtools-move/tasks/evm/wire-evm.ts
shankars99: i uh don't really know - prolly was to get through the wiring faster and i just forgot - will be uncommenting it

packages/devtools-move/tasks/evm/wire-evm.ts
shankars99: oh - will add that in

packages/devtools-move/tasks/evm/wire-evm.ts
shankars99: lemme add in a base-x parser rq

shankars99: just did https://github.com/LayerZero-Labs/devtools/commit/c477ada26aebc8b6172316151688bc51c597feb0

@AlexanderLiteplo you can use this in aptos wire . i will be adding it to evm wire later today

packages/devtools-move/sdk/endpoint.ts
shankars99: @AlexanderLiteplo

packages/devtools-move/sdk/endpoint.ts
shankars99: @AlexanderLiteplo

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
shankars99: to use aptos (even for building contracts) you need to run aptos init which creates a yaml file with the rpc url, account address, public key, (private key lol) - which aptos calls under the hood for all transactions, building, and deploying. so we figured we might as well just use it

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
shankars99: i think the value there is yanked from aptos config which is pre validated

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
shankars99: @AlexanderLiteplo if he knows what that is

packages/devtools-move/sdk/msgLib.ts
shankars99: @AlexanderLiteplo

packages/devtools-move/tasks/evm/wire-evm.ts
shankars99: done!

packages/devtools-move/tasks/move/build.ts
shankars99: hmm im not sure about this. i programmed the build and deploy so if there is one (@AlexanderLiteplo  do you know?) i can add in flags for it

tests-user/tests/create-lz-oapp.bats
```diff
@@ -357,6 +411,42 @@ teardown() {
...
+}
+
+@test "should work with npm & oft aptos coin example in CI mode" {
+    local DESTINATION="$PROJECTS_DIRECTORY/npm-oft-aptos-coin"
+
+    LZ_ENABLE_EXPERIMENTAL_MOVE_VM_EXAMPLES=1 npx --yes create-lz-oapp --ci --example oft-aptos-coin --destination $DESTINATION --package-manager npm
+    cd "$DESTINATION"
+    npm run compile
+    npm run test
+}
```
St0rmBr3w, line 439: if we're not gonna have the coin example either, you'll probably need to update the unit tests here

packages/devtools-move/tasks/move/build.ts
AlexanderLiteplo: We are currently using 6.0.1 so I'm just going to check if it's that or greater.

packages/devtools-move/tasks/evm/wire-evm.ts
AlexanderLiteplo: @shankars99 did you do this?

packages/devtools-move/sdk/endpoint.ts
AlexanderLiteplo: never returns undefined

packages/devtools-move/sdk/endpoint.ts
AlexanderLiteplo: rpc requrest will error if undefined

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
AlexanderLiteplo: this week i'm updating this to use env and get rid of yaml completely

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
AlexanderLiteplo: to connect to movement we need to specify and indexer, however this code is not currently in use and was built in prep for movement and will eventually be overwritten

packages/devtools-move/sdk/msgLib.ts
AlexanderLiteplo: didn't know that existed. at this point its been thoroughly tested though so not going to fix

packages/devtools-move/sdk/oapp.ts
AlexanderLiteplo: old abstraction - deleted

packages/devtools-move/sdk/moveVMConnectionBuilder.ts
AlexanderLiteplo: pulled from aptos config so will always be the same. for movement and initia update we will trashcan this method and rely on .env and move.layerzero.config for chain

PR #1078, opened 2024-11-25
[bug fix] AccountNotInitialized when InitOft in createOFT task

zhangchitc:
- when trying to create Solana OFT using `pnpm hardhat lz:oft:solana:create`, it always report
```
(base)  zhangchi@C4WQ36D2W1---Chi's-MacBook-Pro  ~/code/oft-solana  pnpm hardhat lz:oft:solana:create --eid 40168 --program-id HVyr1MU52zXpf4LDf9Uvd2nfzdSL5nQgqv2uTvULjwSe --only-oft-store true
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
storePda:  DKy8dpZ5p4fwtaibGWWgn2nLRZvtuHLX238gCZs7LJ27
No additional minters specified.  This will result in only the OFTStore being able to mint new tokens.
created SPL multisig @ GTEKmMS9651kv2Go8srie7MJTewJJuT365k6Qu8riaRa
createTokenTx: https://explorer.solana.com/tx/2dz1DE3RBZywyLHNzESn5WAn7ihbetG3tLj7JytPbp46t64VXpD6dGfvnnLayEgf8NWAWLeUfQngoyJ6oCZwRQNm?cluster=devnet
An unexpected error occurred:

SendTransactionError: Simulation failed. 
Message: Transaction simulation failed: Error processing Instruction 0: custom program error: 0xbc4. 
Logs: 
[
  "Program HVyr1MU52zXpf4LDf9Uvd2nfzdSL5nQgqv2uTvULjwSe invoke [1]",
  "Program log: Instruction: InitOft",
  "Program log: AnchorError caused by account: token_mint. Error Code: AccountNotInitialized. Error Number: 3012. Error Message: The program expected this account to be already initialized.",
  "Program HVyr1MU52zXpf4LDf9Uvd2nfzdSL5nQgqv2uTvULjwSe consumed 6198 of 200000 compute units",
  "Program HVyr1MU52zXpf4LDf9Uvd2nfzdSL5nQgqv2uTvULjwSe failed: custom program error: 0xbc4"
]. 
Catch the `SendTransactionError` and call `getLogs()` on it for full details.
    at Connection.sendEncodedTransaction (/Users/zhangchi/code/oft-
```
- However token_mint account is definitely initialized after createToken Tx
- It turns out this is RPC caching issue, we have to wait a while for RPC cache to refresh

I am using official devnet RPC: https://api.devnet.solana.com

@nazreen @St0rmBr3w @DanL0

PR #1089, opened 2024-11-27
🪲 race condition in token_mint / multisig creation

ryandgoulding:
Ensure that the multisig is ready to be used prior to continuing onwards.

### DIFF
--- a/examples/oft-solana/tasks/solana/utils.ts
+++ b/examples/oft-solana/tasks/solana/utils.ts
@@ -1,0 +1,19 @@
+import { Connection, PublicKey } from '@solana/web3.js'
+import { backOff } from 'exponential-backoff'
+
+export const getAccountInfo = async (connection: Connection, publicKey: PublicKey) => {
+    return backOff(
+        async () => {
+            const accountInfo = await connection.getAccountInfo(publicKey)
+            if (!accountInfo) {
+                throw new Error('Multisig account not found')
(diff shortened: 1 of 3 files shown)


examples/oft-solana/tasks/solana/utils.ts
```diff
@@ -0,0 +1,19 @@
+import { Connection, PublicKey } from '@solana/web3.js'
+import { backOff } from 'exponential-backoff'
+
+export const getAccountInfo = async (connection: Connection, publicKey: PublicKey) => {
```
nazreen, line 4: I think "getAccountInfo" should be called something else, either "getAccountInfoWithBackoff" or "confirmAccountInitialized"

nazreen (changes requested):

nazreen:
this will also help solve #1078

examples/oft-solana/tasks/solana/utils.ts
DanL0: I agree. Additional alternatives include: "assertAccountReadiness", "awaitAccountInfo"..

DanL0 (approved):
Tested on Solana Devnet. Seems working.

DanL0 (approved):

nazreen (approved):

nazreen:
tested. works.

PR #1090, opened 2024-11-28
docs(oft-solana): remove usage of solana-install

nazreen:
- in some instances, users might not have `solana-install` so defaulting to fetch from source is more reliable

Thanks to @DanL0 for flagging this


DanL0 (approved):
will be more reliable

PR #1095, opened 2024-12-02
remove AVM instruction

DanL0:
AVM requires higher rust version and it is incompatible with our rust-toolchain.yaml


nazreen:
For future reference,

The issue occurs because avm immediately tries to install the latest anchor version, and Anchor v0.30 does not work with Rust v1.75.0 (defined in `rust-toolchain.toml`)

This would look like:

`cargo install --git https://github.com/coral-xyz/anchor --tag v0.29.0 avm --force`

However I do prefer simplifying the install steps, so this PR is good.

nazreen (approved):

PR #1100, opened 2024-12-03
fix CLI commands instruction in solana 

DanL0:
1. get priority fees command was missing "lz:" prefix so it was broken
2. show meaningful error if wire command is running without solana program id
before: `error:   [OApp] Failed to check OApp configuration: TypeError: Cannot read properties of undefined (reading 'toBase58')`
after: `error:   Missing --solana-program-id CLI argument`
3. fix --only-oft-store param instruction

### DIFF
--- a/examples/oft-solana/tasks/common/wire.ts
+++ b/examples/oft-solana/tasks/common/wire.ts
@@ -87,6 +87,11 @@ task(TASK_LZ_OAPP_WIRE)
         // Then we grab the programId from the args
         const programId = args.solanaProgramId
 
+        if (!programId) {
+            logger.error('Missing --solana-program-id CLI argument')
+            return
+        }
+
         const configurator = args.internalConfigurator
 
         //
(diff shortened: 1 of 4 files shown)


examples/oft-solana/tasks/common/wire.ts
```diff
@@ -87,6 +87,11 @@ task(TASK_LZ_OAPP_WIRE)
         // Then we grab the programId from the args
         const programId = args.solanaProgramId
 
+        if (!programId) {
```
nazreen, line 90: currently, the param defines the `args.solanaProgramId` as optional

```
 // Only pass this if you deployed a new OFT program, if you are using the default
 // LayerZero OFT program you can omit this
.addParam('solanaProgramId', 'The OFT program ID to use', undefined, publicKey, true)
```
(last param sets `isOptional` to `true`)

If we want to make it compulsory for the time being, maybe we should just define it immediately at the addParam by setting `isOptional` to false, and amending the comment

DanL0: it's impossible to set it to mandatory because the parent task that it calls has it set to false therefore it won't compile if you change that flag to false

also there is no such thing as default OFT program, everyone deploys their own. I will remove that comment.

nazreen (dismissed):

nazreen (dismissed):

nazreen (approved):

PR #1104, opened 2024-12-04
fix: bump solana/web3.js version to 1.95.8

St0rmBr3w:
In response to a security incident that Anza has released details about, we are upgrading @solana/web3.js to version 1.95.8. Earlier today, a compromised publish-access account led to the release of malicious versions (1.95.6 and 1.95.7), which allowed attackers to steal private key material from applications that directly handle private keys.

While the affected versions have been unpublished, it is critical for all developers to upgrade to 1.95.8 to ensure security.


DanL0:
Is this PR still relevant after https://github.com/LayerZero-Labs/devtools/pull/1102 was merged?

It doesn't seem to update any solana/web3.js package.

PR #1108, opened 2024-12-04
🧹 Fix logic to compare compute setting to static counterparts, rather than evaluating static ComputeSetting directly

QuitCrypto:
The existing logic evaluates static compute setting values as conditions for ternary expressions, which will always evaluate truthy. They should instead be comparing the existing compute setting to their static counterparts.

### DIFF
--- a/.changeset/eight-llamas-deny.md
+++ b/.changeset/eight-llamas-deny.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/ua-devtools-evm": patch
+---
+
+Fixes an breaking issue when running lz:read:resolve-command with map or reduce compute settings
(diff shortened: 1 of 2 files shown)


yargo13:
Nice catch, thanks! Can you generate a changeset with `pnpm changeset` ?

yargo13:
Also saw that linter failed, fix would be
```
            const mapper =
                computeSetting === ComputeSetting.OnlyReduce
                    ? (this.logger.info('OnlyReduce setting is used. Skipping map step.'),
                      (r: RequestResponsePair) => r.response)
                    : (r: RequestResponsePair) => this.lzMap(compute, r, timeMarker)
            const reducer =
                computeSetting === ComputeSetting.OnlyMap
                    ? (this.logger.info('OnlyMap setting is used. Skipping reduce step.'),
                      (mappedResponses: string[]) => mappedResponses.join(''))
                    : (mappedResponses: string[]) => this.lzReduce(compute, cmd, mappedResponses, timeMarker)
```

yargo13 (approved):

yargo13 (changes requested):
Let's put it as p

QuitCrypto:
Re-marked it as a patch and rebased to sync it up with main again

yargo13 (approved):

PR #1117, opened 2024-12-07
Update layerzero.config.ts

zh95:
The commented-out section uses ExecutorOptionType. 
When you want use "enforcedOptions" in the file and call "npx hardhat lz:oapp:wire" to enable it, it will show the error: "ExecutorOptionType is not defined."
Adding import { ExecutorOptionType } from '@layerzerolabs/lz-v2-utilities'; resolves the issue.


ryandgoulding (changes requested):
This will fail `lint` as the import is not used (i.e., it is commented out).

PR #1119, opened 2024-12-09
🧹 improve quote oft default impl

ryandgoulding:


EWCunha (approved):

packages/oft-evm/test/OFT.t.sol
```diff
@@ -28,6 +28,7 @@ import { NativeOFTAdapterMockCodec } from "./lib/NativeOFTAdapterMockCodec.sol";
 
 import "forge-std/console.sol";
 import { TestHelperOz5 } from "@layerzerolabs/test-devtools-evm-foundry/contracts/TestHelperOz5.sol";
+import "../contracts/interfaces/IOFT.sol";
```
TRileySchwarz, line 31: You can import this explicitly via the OFTCore import above?

TRileySchwarz (approved):

PR #1126, opened 2024-12-11
upgraded lz-evm-sdk, lz-definitions

KaiStevenson:


packages/oapp-alt-evm/artifacts/IOAppCore.sol/IOAppCore.json
St0rmBr3w, line 1: Please don't add artifacts if they're not necessary for the change being submitted.

St0rmBr3w (changes requested):
Some artifacts got included in the PR, can you remove them?

PR #1133, opened 2024-12-16
feat(oft-solana/wire task): error when no solana secret key passed in

nazreen:
- In wire task, current behavior is to warn (easily overlooked) when no `--solana-secret-key` is passed in and generate a random keypair. This does not make sense and leads to confusion as the resulting error of `Attempt to debit an account but found no record of a prior credit.` is obscure.
- updated behavior: throw error when no`--solana-secret-key` is provided

### DIFF
--- a/examples/oft-solana/tasks/common/wire.ts
+++ b/examples/oft-solana/tasks/common/wire.ts
@@ -70,14 +70,13 @@ task(TASK_LZ_OAPP_WIRE)
         //
         //
 
-        if (args.solanaSecretKey == null) {
-            logger.warn(
-                `Missing --solana-secret-key CLI argument. A random keypair will be generated and interaction with solana programs will not be possible`
-            )
+        if (!args.solanaSecretKey) {
+            logger.error('Missing --solana-secret-key CLI argument.')
+            return
         }
 
         // The first step is to create the user Keypair from the secret passed in
-        const wallet = args.solanaSecretKey ?? Keypair.generate()
+        const wallet = args.solanaSecretKey
         const userAccount = wallet.publicKey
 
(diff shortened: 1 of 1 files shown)


St0rmBr3w:
I beleive this is due to the fact that a lot of solana web3.js REQUIRES a keypair for logging and getting values.
The problem is that the getter function's shouldn't necessarily require a valid keypair.
In 90% of cases it doesn't matter since the team is using the project too, but I believe we needed to include this since the CLI will throw some errors otherwise.

Need to investigate where in the code this breaks down.

nazreen:
@St0rmBr3w the PR makes `--solana-secret-key` compulsory for the wire task, which is not a read-only function. 

Maybe you are referring to tasks that inherit the wire task, like`lz:oapp:init:solana`? well that is also a write task.

The thing is, there's also inconsistency in terms of how the solana private key is retrieved

In `lz:oft:solana:create`, it just reads the `SOLANA_PRIVATE_KEY` env var, through `deriveConnection` 

This is also just modifying the wire task in `oft-solana`, which of course would involve writing for Solana side.

If the concern is that this file might be used for other repos and therefor wrongly require a Solana private key when the project does not need Solana, then we can do a conditional early return, only require `solanaSecretKey` when `programId` is present (but note that currently in this task we early return when `programId` is not provided)

nazreen:
@St0rmBr3w 

> I beleive this is due to the fact that a lot of solana web3.js REQUIRES a keypair for logging and getting values.
> Need to investigate where in the code this breaks down.

Okay, so I should text the get config scripts, right ?

if they break -> scrap this PR.
if they don't break -> we can accept this PR.

PR #1142, opened 2025-01-01
🪲 config get regression

ryandgoulding:
#988 introduced a regression where `to.address` is used when it should be `from.address`.  This caused inappropriate values to be returned to the CLI for the `lz:oapp:config:get` task only.  It did not affect the wiring task which did not suffer the same bug.

This was tested against a testnet deployed OFT to ensure proper functionality of the task was restored.


bholcomb8 (approved):
LGTM!

Maliksb11:
Ok

PR #1143, opened 2025-01-01
🪲 Solana EndpointV2SDK should use `delegate` as the `admin`

ryandgoulding:
Currently, the Solana EndpointV2SDK uses the local hot wallet as the `admin` all of the time.  Although this works well when the OFT is owned by an EOA, it is 100% incorrect after an alternative delegate has been set.

This change swaps in the `delegate` in appropriate places, fixing the bug.


Maliksb11:
Ok

nazreen (approved):

PR #1147, opened 2025-01-05
🧹bump version

ryandgoulding:


ryandgoulding:
bumps version minimum to one that has a fixed `lz:oapp:config:get`.

nazreen (approved):

PR #1150, opened 2025-01-07
🪲squads multisig set send receive libs

ryandgoulding:
This change does the following:

* removes actual unused methods in oft-devtools-solana.  They were actual dead code.
* changes `protocol-devtools-solana` to rely on `@layerzerolabs/oft-v2-solana-sdk`.  This may seem non-intuitive, but the existing solita `setSendLibrary(...)` is broken.  Essentially, it would *always* fail in a squads multisig use case as the provided signer would fail to be Authorized (it wasn't the oapp, nor the delegate).  The updated implementation carefully takes care of this bug.
* bumps `protocol-devtools-solana` to target `es2020`, required to use BigInt nomenclature `0n`.
* Changes `ua-devtools-solana` to use the appropriate `delegate` or `admin`, depending on the access control requirements.  This is something that has been broken since day one.

All of this was tested thoroughly on mainnet:
* Optimism gnosis safe: https://app.safe.global/transactions/queue?safe=oeth:0xb9fe286eFC3dd86e46dc33fE0966E96f7Dfa3d43
* Solana squads safe: https://app.squads.so/squads/ET1xBw5JBwpyJVWZFoGYXwjPh1bxFn4XNTBEfKCMQdqR/transactions

First, optimism was added.  Then, arbitrum:

https://layerzeroscan.com/address/3CS8Q4k4TcCHn1E6cYvBD6HTAgDFg29zqdyaA7ih2jpW


packages/ua-devtools-solana/src/oft/sdk.ts
```diff
@@ -176,23 +176,24 @@ export class OFT extends OmniSDK implements IOApp {
...
-                    await this._setPeerFeeBpsIx(eid),
-                    oft.initOAppNonce({ admin, oftStore }, eid, normalizedPeer),
-                    await this._createSetPeerAddressIx(normalizedPeer, eid),
+                    await this._createSetPeerAddressIx(normalizedPeer, eid), // admin
+                    oft.initSendLibrary({ admin: delegate, oftStore }, eid), // delegate
+                    oft.initReceiveLibrary({ admin: delegate, oftStore }, eid), // delegate
+                    await this._setPeerEnforcedOptionsIx(new Uint8Array([0, 3]), new Uint8Array([0, 3]), eid), // admin
+                    await this._setPeerFeeBpsIx(eid), // admin
+                    oft.initOAppNonce({ admin: delegate, oftStore }, eid, normalizedPeer), // delegate
+                    await this._createSetPeerAddressIx(normalizedPeer, eid), // admin but is this needed?  set twice...
```
ravinagill15, line 193: When is it initially set? During deploy?

ravinagill15 (approved):
LGTM 🚀

packages/ua-devtools-solana/src/oft/sdk.ts
ryandgoulding: A few lines above.  I will try to remove it in a separate PR.  This one already took a ton of testing and it adds minimal cost to the tx.  Thanks!

PR #1177, opened 2025-01-20
🪲temporarily disable rpc validation tests

ryandgoulding:
These tests use live rpcs which are blocking our CI.


ryandgoulding:
This test uses a live RPC, which occasionally does not respond.  I am still fighting through replacing it with a mock server, but at very low priority.  We can just disable for now and run manually if we make changes to these files.

PR #1179, opened 2025-01-20
Update DVNsToAddresses to filter out deprecated dvns

ItsAdel:
`metadata-tools` can be used in `layerzero.config.ts` to `generateConnectionsConfig`. However if deprecated DVNs are listed, they are not properly filtered out and running `pnpm hardhat lz:oapp:wire` sometimes wires up their oapp to the deprecated dvn which will cause transactions to be inflight indefinitely.
___
In https://metadata.layerzero-api.com/v1/metadata/deployments
we can see config like
```
"dvns": {
   "0x3ed2211f49ce343d70cb8ded927ca6c4a6198101": {
    "version": 2,
    "canonicalName": "Bitgo",
    "id": "bitgo"
   },
   "0xdd1da938b19614d6db8c3973c89908df234ad1ce": {
    "version": 2,
    "canonicalName": "IntellectEU"
   },
   "0xe67ef84173d024603a844c4aea6a3a15ccccc32c": {
    "version": 2,
    "canonicalName": "Blockdaemon",
    "id": "blockdaemon"
   },
   "0xd44e25bea2bedcceceb7e104d5843a55d208e8a9": {
    "version": 2,
    "canonicalName": "Japan Blockchain Foundation",
    "id": "joc"
   },
   "0x02feab4e6ca6eebd60d85347762de70ca9ce162a": {
    "id": "bitgo",
    "version": 2,
    "canonicalName": "Bitgo",
    "deprecated": true
   },
```
…
Here we see two different Bitgo entires, one is deprecated, one is not
However, when wiring up the paths using
`pnpm hardhat lz:oapp:wire` it doesn’t filter out the deprecated dvn
I faced this issue in https://scan-testnet.layerzero-api.com/v1/messages/tx/0x2cacc12e4a5c47fc8f51671aa32577773a22cee8aba83e240ec7af0fec98f290
where it tries to use the deprecated dvn `0x02feab4e6ca6eebd60d85347762de70ca9ce162a`
and fails verification causing transaction to be inflight indefinitely

### DIFF
--- a/packages/metadata-tools/src/config-metadata.ts
+++ b/packages/metadata-tools/src/config-metadata.ts
@@ -46,7 +46,8 @@ export function DVNsToAddresses(dvns: string[], chainKey: string, metadata: IMet
         for (const [dvnAddress, dvnDetails] of metadataDVNs) {
             if (dvnDetails.canonicalName === dvn && !dvnDetails.lzReadCompatible && dvnDetails.version === 2) {
                 if (dvnDetails.deprecated) {
-                    console.log(`Warning: DVN "${dvn}" is deprecated.`)
+                    console.log(`Warning: DVN "${dvn}" is deprecated. Ignoring ...`)
+                    continue
                 }
 
                 dvnAddresses.push(dvnAddress)
--- a/packages/metadata-tools/src/config-metadata.ts
+++ b/packages/metadata-tools/src/config-metadata.ts
@@ -44,12 +44,12 @@ export function DVNsToAddresses(dvns: string[], chainKey: string, metadata: IMet
 
         let i = 0
         for (const [dvnAddress, dvnDetails] of metadataDVNs) {
-            if (dvnDetails.canonicalName === dvn && !dvnDetails.lzReadCompatible && dvnDetails.version === 2) {
-                if (dvnDetails.deprecated) {
-                    console.log(`Warning: DVN "${dvn}" is deprecated. Ignoring ...`)
-                    continue
-                }
-
+            if (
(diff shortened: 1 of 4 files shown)


packages/metadata-tools/src/config-metadata.ts
```diff
@@ -46,7 +46,8 @@ export function DVNsToAddresses(dvns: string[], chainKey: string, metadata: IMet
         for (const [dvnAddress, dvnDetails] of metadataDVNs) {
             if (dvnDetails.canonicalName === dvn && !dvnDetails.lzReadCompatible && dvnDetails.version === 2) {
                 if (dvnDetails.deprecated) {
-                    console.log(`Warning: DVN "${dvn}" is deprecated.`)
+                    console.log(`Warning: DVN "${dvn}" is deprecated. Ignoring ...`)
```
DanL0, line 49: I think this might be a bit misleading when we ignore deprecated dvn but then we actually put another dvn with the same name as in case of BitGo. Can you remove this console.log completely and maybe move the deprecated check one if clause above?

DanL0 (changes requested):

DanL0:
Please also create `pnpm changeset` for metadata-tools package only with minor version bump so NPM version package can be automatically bumped by the bot after PR is merged. 🙏

packages/metadata-tools/src/config-metadata.ts
ItsAdel: Yea that makes more sense, updating

DanL0 (approved):

PR #1188, opened 2025-01-23
Aptos-Move OApp example

AlexanderLiteplo:


examples/oapp-aptos-move/.env.example
```diff
@@ -0,0 +1,15 @@
...
+#
+#   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-
+#  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \
+# `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'
+
+# By default, the examples support both mnemonic-based and private key-based authentication
+# 
+# You don't need to set both of these values, just pick the one that you prefer and set that one
+MNEMONIC=
+PRIVATE_KEY=
```
shankars99, line 15: EVM_PRIVATE_KEY pls

examples/oapp-aptos-move/CHANGELOG.md
shankars99, line 1: changelog to be deleted. theyre only used on existing subrepos. since this is a new one we can yeet

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -0,0 +1,60 @@
...
+    const wallet = new ethers.Wallet(privateKey, provider)
+
+    // Contract address
+    const contractAddress = '0xB9e27d6D01fb6f63362C968BD30142840E0A2A0c'
+
+    // Create contract instance
+    const myOApp = new ethers.Contract(contractAddress, ABI, wallet)
+
+    // Destination endpoint ID for Aptos
+    const aptosEid = 40108 // Replace with actual Aptos EID
```
shankars99, line 30: maybe use EndpointID.<chain> like we do in hardhat config instead of hardcoding it?

shankars99 (commented):
just a few changes

examples/oapp-aptos-move/README.md
```diff
@@ -0,0 +1,218 @@
...
+        },
+        ...
+    ]
+```
+
+Then run the wire commands:
+
+Commands:
+
+```bash
```
St0rmBr3w, line 113: Can you explain more how the config setters work?

examples/oapp-aptos-move/move.layerzero.config.ts
```diff
@@ -0,0 +1,156 @@
...
+}
+
+const bscContract: OmniPointHardhat = {
+    eid: EndpointId.BSC_V2_TESTNET,
+    contractName: 'MyOApp',
+}
+
+const aptosContract: OmniPointHardhat = {
+    eid: EndpointId.APTOS_V2_TESTNET,
+    contractName: 'oapp',
```
St0rmBr3w, line 18: Why are the contract names different here? Shouldn't they be the same syntax wise?

examples/oapp-aptos-move/move.layerzero.config.ts
```diff
@@ -0,0 +1,156 @@
...
+        },
+        {
+            from: bscContract,
+            to: aptosContract,
+            config: {
+                enforcedOptions: [
+                    {
+                        msgType: MsgType.SEND,
+                        optionType: ExecutorOptionType.LZ_RECEIVE,
+                        gas: 200_000, // gas limit in wei for EndpointV2.lzReceive
```
St0rmBr3w, line 114: Can you be more specific here about Aptos gas options? What is being set, and why 200_000? Alvin said the gas limit should be much lower on Aptos, closer to 1_500 gas units.

examples/oapp-aptos-move/scripts/aptos-move-get-count.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import { Aptos, AptosConfig, Network } from '@aptos-labs/ts-sdk'
+import * as dotenv from 'dotenv'
+
+// Load environment variables from .env file
+dotenv.config()
+
+// OApp configuration
+const OAPP_ADDRESS = '' // Set your OApp's address on Aptos
```
St0rmBr3w, line 15: Why doesn't this just read from the layerzero.config.ts? It has all the OApp declarations there.

examples/oapp-aptos-move/scripts/aptos-move-send.ts
```diff
@@ -0,0 +1,122 @@
...
+    throw new Error('Please set APTOS_PRIVATE_KEY and ACCOUNT_ADDRESS in your .env file')
+}
+
+// Configuration constants
+const APTOS_PRIVATE_KEY = process.env.APTOS_PRIVATE_KEY
+const ACCOUNT_ADDRESS = process.env.ACCOUNT_ADDRESS
+
+// OApp configuration
+const OAPP_ADDRESS = '' // Set your OApp's address on Aptos
+const REMOTE_EID = EndpointId.BSC_V2_TESTNET // Destination chain endpoint ID
```
St0rmBr3w, line 42: Same here, why does this hardcode the eid and oapp address and not just take them either from layerzero config or as arguments in the task / script?

examples/oapp-aptos-move/scripts/aptos-move-send.ts
```diff
@@ -0,0 +1,122 @@
...
+const aptos = new Aptos(new AptosConfig({ network: NETWORK }))
+
+/**
+ * Sends a cross-chain message using LayerZero's OApp protocol.
+ * This function demonstrates the complete flow of sending a message from Aptos to another chain.
+ */
+async function send() {
+    // Configure message options with gas for execution
+    const options = Options.newOptions().addExecutorLzReceiveOption(BigInt(30000))
+    const extraOptions = options.toBytes()
```
St0rmBr3w, line 60: You shouldn't hardcode extra options here, it should be an optional argument passed via flag.

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -0,0 +1,61 @@
...
+    const contractAddress = '<your-contract-address>'
+
+    // Create contract instance
+    const myOApp = new ethers.Contract(contractAddress, ABI, wallet)
+
+    // Destination endpoint ID for Aptos
+    const aptosEid = EndpointId.APTOS_V2_TESTNET
+
+    // Message to send
+    const message = 'Hello Aptos!'
```
St0rmBr3w, line 34: similar to prev question, why hardcode values versus passing them in via flag in the CLI?

examples/oapp-aptos-move/scripts/get-count.ts
```diff
@@ -0,0 +1,26 @@
...
+ * A utility script to verify cross-chain message delivery by checking the counter value.
+ * The counter increments each time a message is successfully received by the OApp,
+ * providing a simple way to confirm that cross-chain communication is working as expected.
+ */
+async function main() {
+    const abi = ['function counter() view returns (uint256)']
+
+    const contractAddress = 'your-contract-address'
+
+    const provider = new ethers.providers.JsonRpcProvider('your-rpc-url')
```
St0rmBr3w, line 13: Why pass an argument here when you can use the CLI to get the RPC?

examples/oapp-aptos-move/move.layerzero.config.ts
St0rmBr3w, line 1: I'm still of the opinion that you should just override the vanilla create-lz-oapp getter function to ignore Aptos OmniPoints, that way you can have your connections still in a single layerzero.config file.

Makes less sense imo to keep them as separate files.

examples/oapp-aptos-move/scripts/aptos-move-get-count.ts
```diff
@@ -0,0 +1,47 @@
...
+const aptos = new Aptos(new AptosConfig({ network: NETWORK }))
+
+/**
+ * Gets the current counter value from the OApp contract on Aptos.
+ * This function demonstrates how to call a view function on an Aptos contract.
+ */
+async function getCount() {
+    try {
+        // Call the view function to get the counter value
+        const counter = await aptos.view({
```
St0rmBr3w, line 28: Is this meant to still be added? Wondering why getCount() is needed at all.

examples/oapp-aptos-move/README.md
AlexanderLiteplo: not sure what you mean by this
do you want us to explain what the config does?

examples/oapp-aptos-move/scripts/aptos-move-get-count.ts
AlexanderLiteplo: given this is a script for testing, if users explicitly set their own addresses then that is one less thing that can break and confuse them

examples/oapp-aptos-move/scripts/aptos-move-get-count.ts
AlexanderLiteplo: this just confirms that the counter was bumped

its for partners to test their newly deployed oapp

examples/oapp-aptos-move/scripts/aptos-move-get-count.ts
AlexanderLiteplo: I can delete it if u don't want though

PR #1192, opened 2025-01-27
🧹bump lz-definitions across the project

ryandgoulding:
Just bump lz-definitions, which includes new Endpoint and contract definitions.  Other upgrades should be considered on a more conservative case-by-case basis.


DanL0 (changes requested):
seems tests are failing

ryandgoulding:
> seems tests are failing

snapshots updated

ravinagill15 (approved):
LGTM!

PR #1197, opened 2025-01-28
🧹 only run user tests using pnpm package manager

ryandgoulding:
Remove 16 tests that rarely find anything worth addressing by just using a single package manager.


ItsAdel:
I might be lacking context here...

but couldn't an enduser simply choose to use a different package manager such as yarn. And then that makes these removed tests relevant to keep?

ryandgoulding:
@ItsAdel these extra package managers add ~20 mins to each run and have never found a legit bug.

PR #1185, opened 2025-01-22
feat(oft-solana): add fallback for when `getSimulationComputeUnits` fails

nazreen:
## Motivation
We've found that in times of network congestion, the RPC method that supports the `getSimulationComputeUnits` function would fail, resulting in scripts being stuck as every transaction currently relies on `getSimulationComputeUnits`.  This blocks the deployment of Solana OFTs.

## Solution
Given that as part of a Solana OFT deployment, we are executing a standard set of instructions, we can introduce a fallback by way of hardcoded CU estimates for each transaction type. The following is introduced into `examples/oft-solana/tasks/solana/index.ts`:

```
export enum TransactionType {
    CreateToken = 'CreateToken',
    CreateMultisig = 'CreateMultisig',
    InitOft = 'InitOft',
    SetAuthority = 'SetAuthority',
    InitConfig = 'InitConfig',
    SendOFT = 'SendOFT',
}

const TransactionCuEstimates: Record<TransactionType, number> = {
    // for the sample values, they are: devnet, mainnet
    [TransactionType.CreateToken]: 125_000, // actual sample: (59073, 123539), 55785 (more volatile as it has CPI to Metaplex)
    [TransactionType.CreateMultisig]: 5_000, // actual sample: 3,230
    [TransactionType.InitOft]: 70_000, // actual sample: 59207, 65198 (note: this is the only transaction that createOFTAdapter does)
    [TransactionType.SetAuthority]: 8_000, // actual sample: 6424, 6472
    [TransactionType.InitConfig]: 42_000, // actual sample: 33157, 40657
    [TransactionType.SendOFT]: 230_000, // actual sample: 217,784
}
```

We use the fallback in `getComputeUnitPriceAndLimit`:

```
...
export const getComputeUnitPriceAndLimit = async (
    connection: Connection,
    ixs: Instruction[],
    wallet: KeypairSigner,
    lookupTableAccount: AddressLookupTableAccount,
    transactionType: TransactionType
) => {
    const { averageFeeExcludingZeros } = await getFee(connection)
    const priorityFee = Math.round(averageFeeExcludingZeros)
    const computeUnitPrice = BigInt(priorityFee)

    let computeUnits

    try {
        computeUnits = await getSimulationComputeUnits(
            connection,
            ixs.map((ix) => toWeb3JsInstruction(ix)),
            toWeb3JsPublicKey(wallet.publicKey),
            [lookupTableAccount]
        )
    } catch (e) {
[...]

PR #1216, opened 2025-01-31
feat: calldata as an execution mode + default gasLimit 

shankars99:
1. using `--only-calldata` as an execution mode skips transaction submission and only logs in transactions under `/transactions/${--oapp-config}/calldata/run-id.json`
2. Testnets have flaky `gasLimit` and may not always pass. Defaulting on a million (`1_000_000`) should it fail (don't worry the excess gas is refunded, also this is mostly for testnets)

### DIFF
--- a/packages/devtools-move/tasks/evm/wire-evm.ts
+++ b/packages/devtools-move/tasks/evm/wire-evm.ts
@@ -164,13 +164,17 @@ async function wireEvm(args: any) {
     try {
         anvilForkNode = new AnvilForkNode(rpcUrlSelfMap, 8545)
 
-        if (args.simulate === 'true') {
-            const forkRpcMap = await anvilForkNode.startNodes()
-            await executeTransactions(omniContracts, TxTypeEidMapping, forkRpcMap, 'dry-run', privateKey, args)
+        if (args.only_calldata === 'true') {
+            await executeTransactions(omniContracts, TxTypeEidMapping, rpcUrlSelfMap, 'calldata', privateKey, args)
         } else {
-            console.warn('--simulate set to false\n Skipping simulation and going directly to broadcast')
+            if (args.simulate === 'true') {
+                const forkRpcMap = await anvilForkNode.startNodes()
+                await executeTransactions(omniContracts, TxTypeEidMapping, forkRpcMap, 'dry-run', privateKey, args)
+            } else {
+                console.warn('--simulate set to false\n Skipping simulation and going directly to broadcast')
+            }
+            await executeTransactions(omniContracts, TxTypeEidMapping, rpcUrlSelfMap, 'broadcast', privateKey, args)
         }
-        await executeTransactions(omniContracts, TxTypeEidMapping, rpcUrlSelfMap, 'broadcast', privateKey, args)
     } catch (error) {
         throw new Error(`Failed to wire EVM contracts: ${error}`)
     } finally {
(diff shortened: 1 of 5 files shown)


packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
```diff
@@ -103,12 +103,25 @@ export async function executeTransactions(
                 const provider: providers.JsonRpcProvider = new providers.JsonRpcProvider(rpcUrlsMap[eid])
                 const signer = accountEidMap[eid].signer
 
-                populatedTx.gasLimit = await provider.estimateGas(populatedTx)
+                // Try to estimate gas limit, if it fails, use a default value
+                const gasLimit = await provider
+                    .estimateGas(populatedTx)
+                    .then((gasLimit) => {
```
ItsAdel, line 109: nit: you can emit the then block, no custom logic is happening here

packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
```diff
@@ -103,12 +103,25 @@ export async function executeTransactions(
                 const provider: providers.JsonRpcProvider = new providers.JsonRpcProvider(rpcUrlsMap[eid])
                 const signer = accountEidMap[eid].signer
 
-                populatedTx.gasLimit = await provider.estimateGas(populatedTx)
+                // Try to estimate gas limit, if it fails, use a default value
+                const gasLimit = await provider
```
ItsAdel, line 107: Asked for something similar in a previous PR but would be cool to do exponential backoff and retry a couple times before defaulting to default values.

https://github.com/LayerZero-Labs/devtools/pull/1185#discussion_r1931323450

packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
```diff
@@ -103,12 +103,25 @@ export async function executeTransactions(
...
+                        return BigNumber.from(1_000_000)
+                    })
+
+                populatedTx.gasLimit = gasLimit
                 populatedTx.gasPrice = accountEidMap[eid].gasPrice
                 populatedTx.nonce = accountEidMap[eid].nonce++
 
                 let sendTx: Promise<providers.TransactionResponse> | undefined = undefined
-                if (simulation === 'broadcast') {
+                if (simulation === 'broadcast' || simulation === 'dry-run') {
```
ItsAdel, line 124: nit: probably better to do `if (simulation !== 'calldata')` if we are are to have other modes later where we would have to keep adding to this list of conditionals.

packages/devtools-move/tasks/evm/wire-evm.ts
ItsAdel, line 1: Can we make it an enum or a string literal union in TypeScript for better type safety?

e.g. `type SimulationMode = 'calldata' | 'broadcast' | 'dry-run';`

ItsAdel (commented):
some possible improvements

packages/devtools-move/tasks/evm/wire-evm.ts
shankars99: done

packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
shankars99: done

packages/devtools-move/tasks/evm/wire/transactionExecutor.ts
shankars99: nah im good

PR #1222, opened 2025-01-31
Diff for troubleshooting build of `examples/oft-aptos-move`

andygolay:
This branch shows the changes I made to build the `examples/oft-aptos-move` package.

I don't recommend merging this branch; there are many extra modified files and I'm not sure why. I'm also not too clear on why these changes made the package build (other than to eliminate specific build errors that came up), or what the side effects of the changes might be. So, this branch is more for troubleshooting reference if helpful.

The changes I made were in `packages`:

image

- In a few `ts-config.json` files, I added `"declaration": true,` or changed `"declaration": false,` to `"declaration": true,`

- in `anvil-test.ts`, add `pruneHistory: undefined,` on line 25

- In `protocol-devtools/src/uln302/schema.ts`, I modified `Uln302UlnConfigSchema` pretty heavily. This is probably a hack and would likely be dangerous to merge, but it got me past a related type error I got.


andygolay:
Closing in favor of a PR already created by LayerZero team.

PR #1233, opened 2025-02-04
Remove Solana from CI

DanL0:
Removes Solana from CI because currently build is broken and works only due to caching.

Tests were disabled before anyway:
```
# FIXME The Solana tests need to be ported to either use a stable deployment
# or a local node. Until then, we disable the tests
# LZ_DEVTOOLS_ENABLE_SOLANA_TESTS: 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@
 # 
 # This issue does not affect users, it's only related to the test runner
 # so the code will still work on node 18.16.0
-ARG NODE_VERSION=20.10.0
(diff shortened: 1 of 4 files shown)


Dockerfile
shankars99, line 1: im afraid that changing the Dockerfile would affect devcontainer (`.devcontainer/devcontainer.local.json`) since it depends on Dockerfile. 
This may result in us not having solana in future devtools named volume

AlexanderLiteplo (approved):
Looks good! Lets get this unblocked.

PR #1244, opened 2025-02-06
Aptos cli local imports

AlexanderLiteplo:
re-doing this because for some reason the changes are gon


AlexanderLiteplo:
Merged with seperate PR

PR #1252, opened 2025-02-07
chore(ua-devtools-solana): update solana deps to version that supports oft202

nazreen:
2 weeks ago, the exports of `@layerzerolabs/oft-v2-solana-sdk` were changed
- `accounts` were removed
- `oft202` was added
- the existing `oft` became an alias for `oft302`

`ua-devtools-solana` should be updated to use a version that has the above changes.


shankars99 (approved):
lgtm

PR #1257, opened 2025-02-11
Shankar/create images for both architectures

shankars99:
building ontop of https://github.com/LayerZero-Labs/devtools/pull/1249

this uses `:shankar-create_images_for_both_architectures` for images generated by https://github.com/LayerZero-Labs/devtools/actions/runs/13253301966

and has a fix for `examples/onft721-zksync` 
=> where it includes `eraVersion: '1.0.0'` under `hardhat.config.ts` `solidity`


.devcontainer/README.md
```diff
@@ -12,12 +12,15 @@
 
 ## Usage
 
+USE VS-CODE. CURSOR DOES NOT WORK.
+
 ### Prebuilt images (default)
 
 By default, the devcontainer will use the prebuilt base images from [GHCR](https://github.com/LayerZero-Labs/devtools/pkgs/container/devtools-dev-base).
+(temporary change: using "image": "ghcr.io/layerzero-labs/devtools-dev-base:shankar-create_images_for_both_architectures")
```
ryandgoulding, line 20: Fine to unblock the pipeline.  Maybe we can rename this later?

.github/workflows/actions/install-dependencies/action.yaml
```diff
@@ -3,6 +3,14 @@ description: Install everything we need to build this repo
 runs:
   using: "composite"
   steps:
+    # Clean Hardhat cache directory manually
```
ryandgoulding, line 6: Can you explain why we need to do this manually?

.github/workflows/actions/setup-build-cache/action.yaml
```diff
@@ -11,18 +11,18 @@ runs:
       uses: actions/cache@v4
       with:
         path: node_modules/.cache/turbo
-        key: ${{ runner.os }}-turbo-${{ github.ref_name }}-${{ github.sha }}
+        key: ${{ runner.os }}-${{ runner.arch }}-turbo-${{ github.ref_name }}-${{ github.sha }}
```
ryandgoulding, line 14: Is this related to https://github.com/LayerZero-Labs/devtools/pull/1051?

Dockerfile
```diff
@@ -78,14 +81,20 @@ RUN apt-get install --yes \
     # Parallel is a utilit we use to parallelize the BATS (user) tests
     parallel \
     # Utilities required to build solana
-    pkg-config libudev-dev llvm libclang-dev protobuf-compiler \
+    pkg-config libudev-dev llvm libclang-dev protobuf-compiler cmake\
```
ryandgoulding, line 84: NIT: space after `cmake`.  Can be done in a follow up I know these jobs take forever.

Dockerfile
```diff
@@ -147,22 +153,24 @@ FROM machine AS avm
 
 WORKDIR /app/avm
 
+ENV RUST_TOOLCHAIN_VERSION_ANCHOR=1.83.0
+RUN rustup default ${RUST_TOOLCHAIN_VERSION_ANCHOR}
```
ryandgoulding, line 157: Isn't the default already 1.83.0 now?

.devcontainer/README.md
shankars99: we can but i dont really know how to tag it to latest/main - so lets leave it at that? also makes it so that going back to the original latest or main (from 4 months ago) is easier to switch to if we find something here breaks idk

.github/workflows/actions/install-dependencies/action.yaml
shankars99: running dual caches for hardhat - on `amd64` and `arm64` because the binaries they generate (for `zksolc` and hardhat's compiler) is platform dependant) and hardhat checks if the zksolc library exists
```
ls /root/.cache/hardhat-nodejs/compilers-v2/
linux-amd64  wasm  zksolc  zkvm-solc
```
 -> it finds it -> it tries it and you get the error based on pipe `Error: write EPIPE`

.github/workflows/actions/setup-build-cache/action.yaml
shankars99: nope this is related to the above comment. we need separate caches for the `amd64` and `arm64` builds

Dockerfile
shankars99: mental note to add this onto the one when we rename the image to main

Dockerfile
shankars99: it is yes, but I have that in for if we ever change the global version in the future but want anchor to build on 1.83

PR #1266, opened 2025-02-12
feat(oft-solana): fix typo and mention how to transfer ownership

nazreen:
Fix typo for flag from `--multisigKey` to `--multisig-key`

Ideally, the new section is covered in a docs page. Once that is created, just link to there.

### DIFF
--- a/examples/oft-solana/README.md
+++ b/examples/oft-solana/README.md
@@ -269,7 +269,7 @@ npx hardhat lz:oapp:init:solana --oapp-config layerzero.config.ts --solana-secre
 npx hardhat lz:oapp:wire --oapp-config layerzero.config.ts --solana-secret-key <PRIVATE_KEY> --solana-program-id <PROGRAM_ID>
 ```
 
-With a squads multisig, you can simply append the `--multisigKey` flag to the end of the above command.
+With a squads multisig, you can simply append the `--multisig-key` flag to the end of the above command.
 
 ### Mint OFT on Solana
 
--- a/examples/oft-solana/README.md
+++ b/examples/oft-solana/README.md
@@ -331,14 +331,44 @@ pnpm hardhat lz:oft:solana:setauthority --eid <SOLANA_EID> --mint <TOKEN_MINT> -
 The `OFTStore` is automatically added as a mint authority to the newly created mint authority, and does not need to be
 included in the `--additional-minters` list.
 
-## Common Errors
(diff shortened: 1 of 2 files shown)


examples/oft-solana/README.md
```diff
@@ -331,14 +331,42 @@ pnpm hardhat lz:oft:solana:setauthority --eid <SOLANA_EID> --mint <TOKEN_MINT> -
...
+and execute the transactions.
+
+**Then, set the `owner`.**
+
+How to set owner: https://docs.layerzero.network/v2/developers/evm/create-lz-oapp/configuring-pathways#adding-owner
+
+Now, run for another time
+
+```
+npx hardhat lz:oapp:wire --oapp-config layerzero.config.ts --solana-secret-key <PRIVATE_KEY> --solana-program-id <PROGRAM_ID>
```
ryandgoulding, line 359: isnt this supposed to be this instead?

>npx hardhat lz:ownable:transfer-ownership --oapp-config layerzero.config.ts

ryandgoulding (changes requested):

examples/oft-solana/README.md
nazreen: both actually work. I did think which was better to tell people to use. But as it stands, `lz:ownable:transfer-ownership` doesn't cover `delegate`, so I thought, why not just keep it simple and ask devs to run the same command twice (in this case, it's `wire`). 

longer term, we should have a single task that takes care of change of `delegate` and `owner`. internally we are toying with the idea of renaming `lz:ownable:transfer-ownership` to something like `lz:access-control:transfer` and have it handle both `delegate` and `owner` changes

do you think think here it's better to just use `lz:ownable:transfer-ownership` ?

ryandgoulding: I didn't think `wire` set the admin though?  Here is the configurator and it only sets the delegate to my knowledge?

https://github.com/LayerZero-Labs/devtools/blob/main/packages/ua-devtools/src/oapp/config.ts#L619

nazreen: @ryandgoulding you're right. I must've gotten things mixed up when I was user-testing. My bad. Will amend this now.

nazreen: @ryandgoulding pushed

ryandgoulding (approved):

PR #1267, opened 2025-02-12
🪲 temporarily remove hardhat cache

ryandgoulding:
This is breaking since it was made architecture-dependent. Rather than try to fix, just disable the cache right now.


ryandgoulding:
Will merge only if we can't figure out the reason the cache isn't working correctly.  This just prepares the option.

shankars99 (approved):

PR #1271, opened 2025-02-12
Cache/bring back hardhat caching 

shankars99:
bringing back architecture specific hardhat cache


ItsAdel (approved):
noticed the last PR to remove it. guess cache is all good and it seems like the same code removed

PR #1273, opened 2025-02-13
fix: anchor build has a race condition where it tries building the rust toolchain concurrently across different anchor builds

shankars99:
`anchor build` runs a rust toolchain build process when run, this creates temp files that it later tries copying in. 
The issue is that we run `pnpm turbo compile` in parallel - this makes it so that we have more than 1 `anchor build` running at the same time. 

These `anchor builds` end up in a state where they each create temporary files (`/root/.rustup/downloads/`) that they want to copy into the "main" rust directory (`/root/.rustup/toolchains/`). This breaks because the first `anchor build` that completes building locks the rustup file as it is now building the solana contracts. During this time the other `anchor build` completes and tries copying its new files into `/root/.rustup/toolchains/` which fails because of the original `anchor build` process. 

This PR pre-builds the rust toolchain that results in `anchor builds` skipping the "build rust toolchain step"

- the original PR that had this issue and fails at the compile step `@layerzerolabs/lzapp-migration-example#compile`<https://github.com/LayerZero-Labs/devtools/actions/runs/13282282114/job/37117263197?pr=1202> 
- my test branch that incorporates these changes locally that now fails with `@layerzerolabs/lzapp-migration-example#test` due to a logic issue `Stack offset of 4608 exceeded max offset of 4096 by 512 bytes` <https://github.com/LayerZero-Labs/devtools/actions/runs/13296998189/job/37131714610?pr=1272>

note: this might be fixed by simply setting the default toolchain in `Dockerfile` to 1.75.0. let us use this patch for the time being

### DIFF
--- a/.github/workflows/actions/install-dependencies/action.yaml
+++ b/.github/workflows/actions/install-dependencies/action.yaml
@@ -34,3 +34,11 @@ runs:
     - name: Rebuild NPM packages
       shell: bash
       run: pnpm rebuild --recursive
+
+    # Install Rustup
+    - name: Install Rustup
+      shell: bash
+      run: |
+        export CARGO_HOME=/root/.cargo
+        export RUSTUP_HOME=/root/.rustup
+        rustup default 1.75.0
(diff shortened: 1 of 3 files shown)


St0rmBr3w (approved):
LGTM!

PR #1278, opened 2025-02-13
Movement and Initia CLI

AlexanderLiteplo:
Ready for review and testing.


.github/workflows/reusable-publish.yaml
shankars99, line 1: can you rebase with main? these changes are no longer required now that we have the image built from main

.github/workflows/reusable-test.yaml
shankars99, line 1: same as 

can you rebase with main? these changes are no longer required now that we have the image built from main

Dockerfile
shankars99, line 1: same as

can you rebase with main? these changes are no longer required now that we have the image built from main

PR #1281, opened 2025-02-14
improve initialization logic in upgradeable oft

DanL0:
- add disableInitializers to oft and oft adapter upgradeable
- fix foundry compilation warnings in OFT.t.sol by adding view keyword

### DIFF
--- a/examples/oft-upgradeable/contracts/MyOFTAdapterUpgradeable.sol
+++ b/examples/oft-upgradeable/contracts/MyOFTAdapterUpgradeable.sol
@@ -4,7 +4,9 @@ pragma solidity ^0.8.22;
 import { OFTAdapterUpgradeable } from "@layerzerolabs/oft-evm-upgradeable/contracts/oft/OFTAdapterUpgradeable.sol";
 
 contract MyOFTAdapterUpgradeable is OFTAdapterUpgradeable {
-    constructor(address _token, address _lzEndpoint) OFTAdapterUpgradeable(_token, _lzEndpoint) {}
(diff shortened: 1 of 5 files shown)


examples/oft-upgradeable/contracts/MyOFTAdapterUpgradeable.sol
```diff
@@ -4,7 +4,9 @@ pragma solidity ^0.8.22;
 import { OFTAdapterUpgradeable } from "@layerzerolabs/oft-evm-upgradeable/contracts/oft/OFTAdapterUpgradeable.sol";
 
 contract MyOFTAdapterUpgradeable is OFTAdapterUpgradeable {
-    constructor(address _token, address _lzEndpoint) OFTAdapterUpgradeable(_token, _lzEndpoint) {}
+    constructor(address _token, address _lzEndpoint) OFTAdapterUpgradeable(_token, _lzEndpoint) {
+        _disableInitializers();
```
shankars99, line 8: maybe add why we're doing this as a comment?

nazreen: I don't think a comment is strictly necessary since it's a general thing and not LZ-specific
but if we do want to leave a comment then maybe just link directly to the function's docs ? https://docs.openzeppelin.com/contracts/4.x/api/proxy#Initializable-_disableInitializers--

DanL0: yeah, this is pretty standard and well documented already in upgradeable space; we don't need to add a comment on it IMO but if we want we can open another PR

PR #1282, opened 2025-02-14
feat: enable `anchor keys sync` auto-replacement by removing `program_id_from_env!`

nazreen:
## Context

- Currently, we require developers to manually replace the program ID in the OFT's `lib.rs`. This creates an opportunity for human error. We have encountered cases where the developer has forgotten to do this. This will not result in an error during program deployment, but will come up when running the init-config/wire commands.
  - Impact: re-building and re-deployment would be required
- We already ask developers to run `anchor keys sync`, which normally would remove the need for the above manual programId replacement. However, it won't auto replace the value since we use `Pubkey::new_from_array(program_id_from_env!(XXX))`

## Motivation

If we remove the usage of `program_id_from_env!`, then we can remove the need for manual replacement of `programId`, thereby removing once opportunity for human error during development. Replacement is handled by the existing `anchor keys sync` command. Improved DevEx unlocked.

## Changes

- Removed `program_id_from_env!` from `declare_id!`

### DIFF
--- a/examples/lzapp-migration/programs/oft202/src/lib.rs
+++ b/examples/lzapp-migration/programs/oft202/src/lib.rs
@@ -19,10 +19,7 @@ use oapp::{
 use solana_helper::program_id_from_env;
 use state::*;
 
-declare_id!(Pubkey::new_from_array(program_id_from_env!(
-    "OFT_ID",
-    "3ThC8DDzimnnrt4mvJSKFpWQA3UvnbzKM3mT6SHoNQKu"
-)));
+declare_id!("3ThC8DDzimnnrt4mvJSKFpWQA3UvnbzKM3mT6SHoNQKu");
 
 pub const OFT_SEED: &[u8] = b"OFT";
 pub const PEER_SEED: &[u8] = b"Peer";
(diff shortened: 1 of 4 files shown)


examples/oft-solana/programs/oft/src/lib.rs
```diff
@@ -17,10 +17,7 @@ use oapp::{
 use solana_helper::program_id_from_env;
 use state::*;
 
-declare_id!(Pubkey::new_from_array(program_id_from_env!(
```
ryandgoulding, line 20: @ziming-zung can you confirm this should be fine?

examples/lzapp-migration/programs/oft202/src/lib.rs
```diff
@@ -19,10 +19,7 @@ use oapp::{
 use solana_helper::program_id_from_env;
 use state::*;
 
-declare_id!(Pubkey::new_from_array(program_id_from_env!(
-    "OFT_ID",
-    "3ThC8DDzimnnrt4mvJSKFpWQA3UvnbzKM3mT6SHoNQKu"
-)));
+declare_id!("3ThC8DDzimnnrt4mvJSKFpWQA3UvnbzKM3mT6SHoNQKu");
```
ryandgoulding, line 22: @ziming-zung can you confirm this should be fine?

nazreen:
closing in favor of #1326

PR #1296, opened 2025-02-18
feat(oft-solana): remove need to pass in solana secret key flag

nazreen:
lzapp-migration-example will be updated in a separate PR (it also needs its init task to be renamed/refactored)

## Testing

```
LAYERZERO_EXAMPLES_REPOSITORY_REF=#nazreen/solana-rm-secret-key-flag LZ_ENABLE_SOLANA_OFT_EXAMPLE=1 npx create-lz-oapp@latest
```

### DIFF
--- a/examples/oft-solana/tasks/common/wire.ts
+++ b/examples/oft-solana/tasks/common/wire.ts
@@ -1,4 +1,4 @@
-import { Keypair, PublicKey } from '@solana/web3.js'
+import { PublicKey } from '@solana/web3.js'
 import { subtask, task } from 'hardhat/config'
 
 import { firstFactory } from '@layerzerolabs/devtools'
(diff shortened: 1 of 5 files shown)


examples/oft-solana/tasks/common/wire.ts
```diff
@@ -24,7 +26,6 @@ import type { SignAndSendTaskArgs } from '@layerzerolabs/devtools-evm-hardhat/ta
 interface Args {
     logLevel: LogLevel
     solanaProgramId: PublicKey
-    solanaSecretKey?: Keypair
```
St0rmBr3w, line 27: Is there a reason we remove the arg entirely and not just allow them to specify the filepath?

i was thinking we were going to try to read directly the solana-cli default signer. If this is too complex for what it's worth, can understand just loading from .env.

EWCunha (approved):

EWCunha (approved):
Tested and worked well

examples/oft-solana/tasks/common/wire.ts
nazreen: yes, I do think we need to have the keypair filepath be the default, but at the same time, I am not 100% sure about removing the option of specifying the private key directly (through the env var). there could be devs who would like to use the account from their wallet, in which case, they'd be exporting the private key in base58 form.

much like how we allow the evm private key to be specified either via mnemonic or hex, for Solana, we can allow both private key string (base58) or keypair file path. 

i intend on adding the keypair file path feature in a separate PR. this PR was focused on removing solana secret key as a flag, as regardless, neither keypair file path or secret key string should need to be passed in as a param. should just be set via env var like evm mnemonic/private key.

nazreen: will wait for your reply on this first @St0rmBr3w

PR #1299, opened 2025-02-18
🧹audit reviews 

shankars99:
1. `OFTAltCore` rejects transactions where `msg.value > 0`
+ this is because in an `alt variant` we use ERC20 tokens instead of native tokens and you can't pass ERC20 tokens via `msg.value`
+ updates `OFTAltCore` and tests - `MyOFTAlt.t.sol` and `OFTAlt.t.sol` 

2. Caching endpoint's native token:
+ to save gas by reducing a call on every transaction since the token is immutable at the endpoint <https://github.com/LayerZero-Labs/LayerZero-v2/blob/main/packages/layerzero-v2/evm/protocol/contracts/EndpointV2Alt.sol#L16-L20>


shankars99:
`@layerzerolabs/devtools-evm-hardhat-test:test: error:   ⤫ ========== RPC URL validation failed for network(s): arbitrum
`

failed with - im merging

PR #1303, opened 2025-02-19
using rustc 1.75.0

shankars99:
New base image sets the `default rustup` toolchain version to `1.75.0`

This cleans up some ugly code introduced in the anchor-solana race condition resolution PR https://github.com/LayerZero-Labs/devtools/commit/1edf6d7ec45ce45c996ab2e8ada162f45eda7287 where we manually install `rustc v1.75.0` in a `github action` and `docker-compose.yaml`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ARG NODE_VERSION=20.10.0
 # and the base image is built locally
 # 
 # The CI environment will use base images from https://github.com/LayerZero-Labs/devtools/pkgs/container/devtools-dev-base
-# e.g. ghcr.io/layerzero-labs/devtools-dev-base:main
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -114,6 +114,8 @@ FROM machine AS aptos
 WORKDIR /app/aptos
 
 ARG APTOS_VERSION=6.0.1
+ENV RUST_TOOLCHAIN_VERSION_APTOS=1.83.0
+RUN rustup default ${RUST_TOOLCHAIN_VERSION_APTOS}
```
ItsAdel, line 118: very hacky but i see what ur doing here

ItsAdel (approved):
left a comment but lgtm

PR #1327, opened 2025-02-25
🗞️ solana cli debugging tasks

ryandgoulding:


nazreen:
@ryandgoulding , super excited for this PR as I was wishing for something similar and wanted to work on it.

I think there's a need for a single command that we could ask partners to run that they can then copy the output of to share with us, which would save a lot of back and forth.

Could we maybe have only one hardhat task, either with the name `lz:oft:solana:debug` or `lz:oft:debug` with flags that would make the checks more narrow ?

i.e. Running `lz:oft:debug` by default would print all the following:

- all contracts' state (for Solana this is OFT Store) but pre-filtered only to most important values (this of course does not necessarily need to be achieved by this PR)

and running ``lz:oft:debug --solana` would only print the solana related output

alternatively, if you think the above is overkill (I was thinking we should try to give feature parity on the EVM side)

then just `lz:oft:solana:debug` that outputs (for now) the same info as the 3 separate tasks currently in this PR

this can later be expanded to also check whether PeerConfig accounts are already created for the Solana OFT's peers among other things

nazreen (approved):

PR #1334, opened 2025-02-28
🗞️ foundry scripts for LayerZero

shankars99:
# Scripting in foundry

Across LayerZero we have a bunch of scripts -
Krak's Gas Profiling - <https://github.com/LayerZero-Labs/devtools/tree/main/examples/oft/scripts>
Dan's lzReceive Simulate - <https://github.com/DanL0/integrations-foundry-tooling>

Some of these scripts are baked into examples or are stand alone repositories. This PR aims to fix that by:

1. Creating a new package `@layerzerolabs/script-devtools-evm-foundry` located at `/packages/script-devtools-evm-foundry` (following what we do with `TestHelper`)
2. Create a new level 1 script - `LZScripts` that exports scripts internal to it
3. Internal scripts (level 2 scripts) - (`GasProfile` and `SimulateReceive`) these actually implement the foundry script.

Doing this not only gives developers the ability to import these helpful scripts into their packages 
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.0;

import { console } from "forge-std/console.sol";

import { LZScripts } from "@layerzerolabs/script-devtools-evm-foundry/script/LZScripts.s.sol";

contract LZScript is LZScripts {
    function run() public pure override {
        console.log("Hello, World!");
    }
}
``` 

But also run them from the terminal due to the compilation of `script/LZScript.s.sol` generating the `artifacts` for all the Scripts with a `run()` function. 

```shell
$ forge script SimulateReceive
[⠊] Compiling...
No files changed, compilation skipped
Traces:
  [1406766] → new SimulateReceive@0x9f7cF1d1F558E57ef88a59ac3D47214eF25B6A06
    └─ ← [Return] 6916 bytes of code

  [6585] SimulateReceive::run()
    ├─ [0] VM::ffi(["curl", "-X", "GET", "https://scan.layerzero-api.com/v1/messages/tx/0x88a79d59222628aba22c5c616580212c80beae353aaa8ffbd3976dd5aa5d12f8", "-H 'accept: application/json'"])
    │   └─ ← [Revert] vm.ffi: FFI is disabled; add the `--ffi` flag to allow tests to call external commands
    └─ ← [Revert] vm.ffi: FFI is disabled; add the `--ffi` flag to allow tests to call external commands


Gas used: 27649
Error: script failed: vm.ffi: FFI is disabled; add the `--ffi` flag to allow tests to call external commands
```
(This fails because I didn't use the right flags)

Note: This works for scripts that have `run()` implemented. So if it is a custom implementation like Gas Profiler you should be creating a new script under `script/`


St0rmBr3w:
Question I have is should we change the lower level scripts structure to make them more easily callable at the top level via this import?

St0rmBr3w:
I think this is the best way to go, we should prioritize implementing this package.

packages/script-devtools-evm-foundry/script/LZScripts.s.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.22;
+
+import { SimulateScript } from "./SimulateReceive/utils.sol";
+import { GasProfilingScript } from "./GasProfiling/utils.sol";
+
+abstract contract LZScripts is SimulateScript, GasProfilingScript {
+    function IS_WORKING() public pure returns (bool) {
```
St0rmBr3w, line 8: What is the point of this?

shankars99: `IS_WORKING` ? lol its something that foundry does in their `Scripts` so i slotted it in

ryandgoulding (changes requested):
requires rebase

packages/toolbox-foundry/src/forge-std
ravinagill15, line 1: We had to update to the [latest commit hash](https://github.com/foundry-rs/forge-std/commit/60acb7aaadcce2d68e52986a0a66fe79f07d138f) from the forge-std repo because we were running into errors with `forge build` not recognizing some syntax like `vm.snapshotState()` and `vm.randomUint()`

tests-user/lib/bats-assert
ravinagill15, line 1: there were newer commits in the following repos as well hence the changes here:

https://github.com/bats-core/bats-support

https://github.com/bats-core/bats-assert

https://github.com/bats-core/bats-core

packages/script-devtools-evm-foundry/scripts/SimulateReceive/base58decoder.sol
```diff
@@ -0,0 +1,63 @@
...
+            }
+        }
+
+        // Handle leading zeros
+        for (uint256 i = 0; i < input.length && input[i] == BASE58_CHARS[0]; i++) {
+            output[outputIndex++] = 0;
+        }
+
+        // Reverse the output to get the correct byte order
+        bytes memory reversedOutput = new bytes(outputIndex);
```
ItsAdel, line 46: nit: I think theres a way to construct this right to left, so you wouldn't have to reverse the output at the end

ravinagill15: The scripts have been copied over from the following sources so I am hesitant to tinker with them without more tests for now -- I think it is fine for now and if we think it is overly complex, will revisit!

Krak's Gas Profiling - https://github.com/LayerZero-Labs/devtools/tree/main/examples/oft/scripts
Dan's lzReceive Simulate - https://github.com/DanL0/integrations-foundry-tooling

PR #1344, opened 2025-03-05
🪲 OFTAltCore reverts if non zero msg value

shankars99:
The current `OFTAltCore::send()` was `payable` to match the interface's definition. However it did not have a revert for when `msg.value != 0`.

This means that users can send transactions to the `OFT` with `{value: msg.value}` which would simply get locked up in the OFT. This PR introduces a custom revert message for when `msg.value != 0`


ItsAdel (approved):
lgtm

St0rmBr3w (approved):

PR #1357, opened 2025-03-11
feat: make validate rpc taks revert

clauBv23:
also on this PR:
- modified the rpc health check to allow attempts.
- supported the stage (mainnet, testnet, sandbox) param
- increased the timeout
- more descriptive logs


ravinagill15 (commented):
Looks great!

Let's add a changeset to ensure these changes are pushed to npm and can be picked up by other repos. I think we should also update the tests (they were commented out due to unstable RPCs, but it would be good to verify the changes and then re-comment the tests if they are still causing issues

ravinagill15 (approved):

PR #1355, opened 2025-03-10
🧹typescript exports for move suite (devtools-extensible-cli, devtools-move, oft-move)

shankars99:
Painful chore where I export all the `typescript` scripts from `/cli`, `/sdk`, and `/tasks` which not only fixes imports (you no longer need to tag the exact file) but also creates `/dist` for importing across `esm` and `cjs`

and a few fixes like re-used class names

also refer <https://github.com/LayerZero-Labs/devtools/pull/1356>

PR #1361, opened 2025-03-12
🧹oapp-aptos-example import path update

shankars99:
updating `oapp-aptos` 's `scripts/cli` which was missed in the earlier PR  <https://github.com/LayerZero-Labs/devtools/pull/1355>


ravinagill15 (approved):
🚀

PR #1366, opened 2025-03-13
🪲 exports from devtools-extensible-cli are fixed

ryandgoulding:
Otherwise, we would need to import as @layerzerolabs/devtools-extensible-cli/cli in all clients.

### DIFF
--- a/.changeset/shy-monkeys-deny.md
+++ b/.changeset/shy-monkeys-deny.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/devtools-extensible-cli": patch
+---
+
+Fix exports from devtools-extensible-cli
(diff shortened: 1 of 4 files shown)


shankars99:
Wait really? Doesn't pointing tsup to that particular index.ts fix it? It's what I'm doing in all the other packages. And the other devtools packages point to `src/index.ts` and we don't import them as @layerzerolabs/something/src

shankars99 (approved):
unsure why CI fails with `devtools-extensible-cli` <https://github.com/LayerZero-Labs/devtools/actions/runs/13836174152/job/38711629185?pr=1364#step:8:1688> 

CI seems to point to the actual file for this package and since the `index.ts` is under `/cli` the package build fails - this can explain why i needed `eslint.rc` to ignore packages `devtools-move` and `oft-move`.

Since this PR just changes the location of `index.ts` I'm fine with approving it

PR #1383, opened 2025-03-16
🪲 aptos/movement support for Solana peer

ryandgoulding:


St0rmBr3w (commented):
I think this is too opinionated since you can have multiple OFT deployments per program, so defaulting to a specific json might be too restrictive.

PR #1397, opened 2025-03-19
🚀  feat(metadata-tools): allow for custom fetchMetadata implementation

nazreen:
This PR

- adds a unit test for `generateConnectionsConfig` (added before function was changed)
- adds a new param to `generateConnectionsConfig` (currently only accepts `fetchMetadata` that defaults to `defaultFetchMetadata`)
- creates a `defaultFetchMetadata` function
- extract out the interface `IMetadataDvns` from `IMetadata` for better reusability
- unit test for custom `fetchMetadata` implementation

### DIFF
--- a/packages/metadata-tools/src/config-metadata.ts
+++ b/packages/metadata-tools/src/config-metadata.ts
@@ -216,12 +216,12 @@ export async function translatePathwayToConfig(
     return configs
 }
 
-/// allow for a custom metadataUrl
+// allow for a custom metadataUrl
 async function defaultFetchMetadata(metadataUrl = METADATA_URL): Promise<IMetadata> {
     return (await fetch(metadataUrl).then((res) => res.json())) as IMetadata
 }
 
-/// allow for a custom metadataFetcher
+// allow for a custom fetchMetadata
 export async function generateConnectionsConfig(
     pathways: TwoWayConfig[],
     fetchMetadata: () => Promise<IMetadata> = defaultFetchMetadata
--- a/packages/metadata-tools/src/config-metadata.ts
+++ b/packages/metadata-tools/src/config-metadata.ts
@@ -224,7 +224,7 @@ export async function defaultFetchMetadata(metadataUrl = METADATA_URL): Promise<
 // allow for a custom fetchMetadata
 export async function generateConnectionsConfig(
     pathways: TwoWayConfig[],
-    fetchMetadata: () => Promise<IMetadata> = defaultFetchMetadata
+    { fetchMetadata = defaultFetchMetadata }: { fetchMetadata?: () => Promise<IMetadata> }
 ) {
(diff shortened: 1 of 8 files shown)


nazreen:
Current passing check is caused by timeout for a TON test

```
 transactions/signer › OmniSignerTON › signAndSend › should send a token transfer
@layerzerolabs/devtools-ton:test: 
@layerzerolabs/devtools-ton:test:     thrown: "Exceeded timeout of 60000 ms for a test.
@layerzerolabs/devtools-ton:test:     Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."
```

packages/metadata-tools/src/config-metadata.ts
```diff
@@ -217,8 +216,17 @@ export async function translatePathwayToConfig(
...
-    const metadata = (await fetch(METADATA_URL).then((res) => res.json())) as IMetadata
+// allow for a custom metadataUrl
+export async function defaultFetchMetadata(metadataUrl = METADATA_URL): Promise<IMetadata> {
+    return (await fetch(metadataUrl).then((res) => res.json())) as IMetadata
+}
+
+// allow for a custom fetchMetadata
+export async function generateConnectionsConfig(
+    pathways: TwoWayConfig[],
+    fetchMetadata: () => Promise<IMetadata> = defaultFetchMetadata
```
DanL0, line 227: Second parameter should be an object not a function

```
{
   fetchMetadata: defaultFetchMetadata
}
```

This is to allow future configuration extension without breaking changes (assuming we have defaults for everything)

packages/metadata-tools/test/config-metadata.test.ts
```diff
@@ -14,6 +14,92 @@ describe('config-metadata', () => {
...
+                        [], // optional DVNs + threshold
+                    ],
+                    [1, 1],
+                    [undefined, undefined],
+                ],
+            ]
+
+            // Generate config using our custom fetchMetadata
+            const config = await generateConnectionsConfig(pathways, customFetchMetadata)
+            console.log(config)
```
DanL0, line 98: unnecessary console log

DanL0 (changes requested):

packages/metadata-tools/src/config-metadata.ts
nazreen: My bad. I forgot that despite also having suggested it. Will fix now.

packages/metadata-tools/test/config-metadata.test.ts
nazreen: woops. removing now.

DanL0 (approved):

DanL0 (approved):

PR #1407, opened 2025-03-21
🚀 feat(oft-solana): support loading keypair via path

nazreen:
# Motivation

The default Solana devex when it comes to keypairs is to use the keypair at the default path of `~/.config/solana/id.json`. We should support this.

# Change to DevEx

The edit to the README.md sums it up nicely:

> By default, the scripts will use the keypair at the default location `~/.config/solana/id.json`. If you want to use this keypair, there is no need to set any environment variable. There will, however, be a prompt when running certain commands to confirm that you want to use the default keypair.
> 
> If you wish to use a different keypair, then you can set either of the following in the `.env`:
> 
> 1. `SOLANA_PRIVATE_KEY` - this can be either in base58 string format (i.e. when imported from a wallet) or the Uint8 Array in string format (all in one line, e.g. `[1,1,...1]`).
> 
> 2. `SOLANA_KEYPAIR_PATH` - the location to the keypair file that you want to use.

# Changes

- update @solana-developers/helpers to latest version (current version has a bug with the getKeypairFromEnvironment function)
- due to above, added override for @solana/web3.js, otherwise the Keypair times will mismatch (between 1.95 and 1.98)

# Testing

There are these scenarios to test:

1. readOnly mode
2. when both `SOLANA_PRIVATE_KEY` and `SOLANA_KEYPAIR_PATH` are not set (uses default)
3. when only one of the following is set: `SOLANA_PRIVATE_KEY` or `SOLANA_KEYPAIR_PATH`
4. when both `SOLANA_PRIVATE_KEY` and `SOLANA_KEYPAIR_PATH` are set and their publicKey values match
5. when both `SOLANA_PRIVATE_KEY` and `SOLANA_KEYPAIR_PATH` are set and their publicKey values do not match

To test, run:
```
LAYERZERO_EXAMPLES_REPOSITORY_REF=#<BRANCH_NAME> LZ_ENABLE_SOLANA_OFT_EXAMPLE=1 npx create-lz-oapp@latest
```
Select OFT (Solana)

Deploy an OFT program.
Then you can proceed to the test scenarios.

1. readOnly mode
For readOnly mode, you can run the following immediately:

```bash
npx hardhat lz:oft:solana:debug --eid 40168 --oft-store 9Ccpw8khitczfmj7ThadPVsrwuvNzSPMeCCrApsXYqBU      
```

2. when both `SOLANA_PRIVATE_KEY` and `SOLANA_KEYPAIR_PATH` are not set (uses default)

Run the create script:

```
pnpm hardhat lz:oft:solana:create --eid 40168 --program-id $PROGRAM_ID --only-oft-store true
```

Observe the terminal output.

You can exit when it asks you to confirm regarding `onlyOftStore`

3. when only one of the following is set: `SOLANA_PRIVATE_KEY` or `SOLANA_KEYPAIR_PATH`

For `SOLANA_PRIVATE_KEY`:
- You can run `solana-keygen new -o temp-keypair.json ` in the project directory to generate a keypair at a custom path
Then, you can view the contents of temp-keypair.json and copy the whole file's value (ensuring it's all in one line) and paste it as the value for `SOLANA_PRIVATE_KEY.

For `SOLANA_KEYPAIR_PATH`:
- Set the value to `temp-keypair.json` (ensuring to unset `SOLANA_PRIVATE_KEY` when this is set)

set either of the env vars and run the same create script mentioned previously.
Observe the terminal output.
You can exit when it asks you to confirm regarding `onlyOftStore`

4. when both `SOLANA_PRIVATE_KEY` and `SOLANA_KEYPAIR_PATH` are set and their publicKey values match

set the two env vars as mentioned above, leaving both set, run the same create script mentioned previously.
Observe the terminal output.
You can exit when it asks you to confirm regarding `onlyOftStore`

5. when both `SOLANA_PRIVATE_KEY` and `SOLANA_KEYPAIR_PATH` are set and their publicKey values do not match

leaving the `SOLANA_PRIVATE_KEY` value unchanged,
you can generate another keypair in the project directory under another filename, then update the `SOLANA_KEYPAIR_PATH` value.
run the same create script mentioned previously.
Observe the terminal output.
You can exit when it asks you to confirm regarding `onlyOftStore`


St0rmBr3w (approved):
LGTM!

examples/oft-solana/tasks/solana/index.ts
```diff
@@ -42,29 +41,99 @@ const LOOKUP_TABLE_ADDRESS: Partial<Record<EndpointId, PublicKey>> = {
...
         throw new Error(`${key} is not defined in the environment variables.`)
     }
     return value
 }
 
-/**
- * Extracts the SOLANA_PRIVATE_KEY from the environment.  This is purposely not exported for encapsulation purposes.
- */
-const getSolanaPrivateKeyFromEnv = () => getFromEnv('SOLANA_PRIVATE_KEY')
+// TODO in another PR: considerg moving keypair related functions to tasks/solana/utils.ts
```
ryandgoulding, line 61: NIT:  `considerg` -> `consider`

examples/oft-solana/tasks/solana/index.ts
```diff
@@ -42,29 +41,99 @@ const LOOKUP_TABLE_ADDRESS: Partial<Record<EndpointId, PublicKey>> = {
...
+    }
+
+    // Attempt to load from each source
+    const keypairEnvPrivate = process.env.SOLANA_PRIVATE_KEY
+        ? getKeypairFromEnvironment('SOLANA_PRIVATE_KEY')
+        : undefined // #1 SOLANA_PRIVATE_KEY
+    const keypairEnvPath = process.env.SOLANA_KEYPAIR_PATH
+        ? await getKeypairFromFile(process.env.SOLANA_KEYPAIR_PATH)
+        : undefined // #2 SOLANA_KEYPAIR_PATH
+    const keypairDefaultPath = await getKeypairFromFile() // #3 ~/.config/solana/id.json
```
ryandgoulding, line 78: This should be in a try/catch.  If the file does not exist or is not readable, it will throw:

https://github.com/solana-developers/helpers/blob/main/src/lib/keypair.ts#L33

ryandgoulding (changes requested):

St0rmBr3w (changes requested):
Add a changeset here as well

examples/oft-solana/tasks/solana/index.ts
nazreen: I think we should just let the exception bubble up, as the functions already handle various invalid situations:

e.g. https://github.com/solana-developers/helpers/blob/main/src/lib/keypair.ts#L33-L45
````
  } catch (error) {
    throw new Error(`Could not read keypair from file at '${filepath}'`);
  }

  // Parse contents of file
  let parsedFileContents: Uint8Array;
  try {
    parsedFileContents = Uint8Array.from(JSON.parse(fileContents));
  } catch (thrownObject) {
    const error = thrownObject as Error;
    if (!error.message.includes("Unexpected token")) {
      throw error;
    }
    throw new Error(`Invalid secret key file at '${filepath}'!`);
  }
  return Keypair.fromSecretKey(parsedFileContents);
};
```

if we were to catch it, we either will lose the context already given for free, or we'd need to do error message matching again, duplicating the code.

nazreen: or, if you think it's better, then it'll just be catching to add additional context to it, e.g.

```typescript
let keypairEnvPath

try {
  keypairEnvPath = process.env.SOLANA_KEYPAIR_PATH
        ? await getKeypairFromFile(process.env.SOLANA_KEYPAIR_PATH)
        : undefined // #2 SOLANA_KEYPAIR_PATH
} catch (e) {
throw new Error(`Error when trying keypairEnvPath provided via process.env.SOLANA_KEYPAIR_PATH: ${e.message}`)
}

```

ryandgoulding: My point is that not everyone has a key defined at ~/.config/solana/id.json.  If we assume it is there, it could throw an exception if someone is using the environment variables instead.

nazreen: You're right. I overlooked that case. let me update the PR to handle this.

nazreen: @ryandgoulding I've pushed the fix. thanks again for spotting the issue.

nazreen:
@ryandgoulding change requests addressed

ryandgoulding (approved):

PR #1410, opened 2025-03-24
🚀 examples/oapp: update to use simple config

nazreen:
Changes
- add metadata-tools dep
- update LZ config to use simple config
- change from Ethereum Sepolia to OP Sepolia


examples/oapp/layerzero.config.ts
```diff
@@ -1,96 +1,81 @@
...
+        value: 0,
+    },
+    {
+        msgType: 2,
+        optionType: ExecutorOptionType.COMPOSE,
+        index: 0,
+        gas: 80000,
+        value: 0,
+    },
+]
```
St0rmBr3w, line 41: We should not be defining msgType 2 in enforcedOptions unless asked specifically imo, but will follow up internally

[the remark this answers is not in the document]
nazreen: what's the conclusion? remove?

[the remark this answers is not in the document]
nazreen: just read the thread. will remove.

examples/oapp/layerzero.config.ts
```diff
@@ -1,96 +1,68 @@
...
 }
 
-const fujiContract: OmniPointHardhat = {
+const avalancheContract: OmniPointHardhat = {
     eid: EndpointId.AVALANCHE_V2_TESTNET,
     contractName: 'MyOApp',
 }
 
-const amoyContract: OmniPointHardhat = {
+const polygonContract: OmniPointHardhat = {
```
St0rmBr3w, line 16: Do we want to remove Polygon from this list too and do something more stable like ARBSEP?

Polygon just has block confirmation issues and I'd hate for that to block teams.

examples/oapp/layerzero.config.ts
```diff
@@ -1,96 +1,68 @@
...
 
-const config: OAppOmniGraphHardhat = {
-    contracts: [
-        {
-            contract: fujiContract,
-            /**
-             * This config object is optional.
-             * The callerBpsCap refers to the maximum fee (in basis points) that the contract can charge.
-             */
+const EVM_ENFORCED_OPTIONS: OAppEnforcedOption[] = [
```
St0rmBr3w, line 21: Can you add a comment for teams that they need to profile this amount based on the gas limit required to call OApp._lzReceive(...)?

St0rmBr3w (changes requested):
Minor changes, otherwise looks good!

St0rmBr3w (dismissed):

shankars99 (commented):
missing changeset - `pnpm changeset` 

else is good to merge imo @nazreen

ryandgoulding (changes requested):
Shankar requested changeset, so marking this as "Request changes" so that it shows that way in PR review.

nazreen:
@ryandgoulding changeset pushed

shankars99 (approved):

PR #1414, opened 2025-03-25
Aptos Move Example: update README 

AlexanderLiteplo:
Update README.md with token address details and verification process for fungible assets


ItsAdel (approved):
lgtm, but wish there was a better verification process

PR #1421, opened 2025-03-27
🚀 feat(oft-solana): auto-populate oftStore address

nazreen:
# Motivation

Remove need for manual input of OFT Store in the LZ config. Reduce human error.

# Notes

will implement custom deployment filename in another PR. currently it's fixed to `OFT.json`

### DIFF
--- a/examples/oft-solana/layerzero.config.ts
+++ b/examples/oft-solana/layerzero.config.ts
@@ -3,6 +3,8 @@ import { ExecutorOptionType } from '@layerzerolabs/lz-v2-utilities'
 import { generateConnectionsConfig } from '@layerzerolabs/metadata-tools'
 import { OAppEnforcedOption, OmniPointHardhat } from '@layerzerolabs/toolbox-hardhat'
 
+import { getOftStoreAddress } from './tasks/solana'
+
 // Note:  Do not use address for EVM OmniPointHardhat contracts.  Contracts are loaded using hardhat-deploy.
 // If you do use an address, ensure artifacts exists.
 const sepoliaContract: OmniPointHardhat = {
@@ -12,7 +14,7 @@ const sepoliaContract: OmniPointHardhat = {
 
 const solanaContract: OmniPointHardhat = {
     eid: EndpointId.SOLANA_V2_TESTNET,
-    address: '', // NOTE: update this with the OFTStore address.
+    address: getOftStoreAddress(EndpointId.SOLANA_V2_TESTNET), // NOTE: this will return an empty string if the create task has not been run.
 }
 
 const EVM_ENFORCED_OPTIONS: OAppEnforcedOption[] = [
--- a/examples/oft-solana/layerzero.config.ts
+++ b/examples/oft-solana/layerzero.config.ts
@@ -14,7 +14,7 @@ const sepoliaContract: OmniPointHardhat = {
 
 const solanaContract: OmniPointHardhat = {
     eid: EndpointId.SOLANA_V2_TESTNET,
-    address: getOftStoreAddress(EndpointId.SOLANA_V2_TESTNET), // NOTE: this will return an empty string if the create task has not been run.
+    address: getOftStoreAddress(EndpointId.SOLANA_V2_TESTNET),
 }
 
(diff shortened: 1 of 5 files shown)


examples/oft-solana/layerzero.config.ts
```diff
@@ -12,7 +14,7 @@ const sepoliaContract: OmniPointHardhat = {
 
 const solanaContract: OmniPointHardhat = {
     eid: EndpointId.SOLANA_V2_TESTNET,
-    address: '', // NOTE: update this with the OFTStore address.
+    address: getOftStoreAddress(EndpointId.SOLANA_V2_TESTNET), // NOTE: this will return an empty string if the create task has not been run.
```
St0rmBr3w, line 17: @nazreen I'm wondering if instead of this, we can normalize the deployment process based on contractName?

If I have say MyOFT on EVM and the OFT Program on Solana, perhaps we can normalize it somehow such that the contractName argument will know to look for the OFT deployment artifact we generate? It might help with simplifying the config wiring more?

St0rmBr3w: This shouldn't necessarily block this PR, but I wonder if it's a more long-term solution.

St0rmBr3w (dismissed):

examples/oft-solana/tasks/solana/index.ts
```diff
@@ -167,6 +167,11 @@ export const getSolanaDeployment = (
     return JSON.parse(fileContents)
 }
 
+export const getOftStoreAddress = (eid: EndpointId) => {
+    const { oftStore } = getSolanaDeployment(eid)
```
ryandgoulding, line 171: I believe we should throw instead of returning `undefined` here:
```
if (!oftStore) {
    throw new Error('oftStore not added to Solana contract in layerzero configuration file')
 }
 ```
 or similar.

**Review by ryandgoulding — CHANGES_REQUESTED** [review]
Feel like we should revert if oftStore is undefined.  Thoughts?

File: examples/oft-solana/layerzero.config.ts (hunk 1 above)
Comment (nazreen) [reply examples/oft-solana/layerzero.config.ts:17]: My guess is that, if we specify `contractName` for a contract object, it might treat it as an EVM contract object.
I'm basing this assumption on how specifying `address` makes the contract object be treated as a Solana contract object. Maybe the reverse is true too. Not sure, needs to be tested.

But what you're suggesting relates to how @ryandgoulding is pointing out how there are some teams that use the same program to create multiple OFTs. Also consider that even OFT Adapters will use the same OFT program.

For Solana, the program is just the program, and is not the token name. So there's less of a need to change it compared to EVM where the contract name reflects the token name.

File: examples/oft-solana/tasks/solana/index.ts (hunk 2 above)
Comment (nazreen) [reply examples/oft-solana/tasks/solana/index.ts:171]: Oh yes! Will use the DebugLogger for this one. Thanks for the spot.

File: examples/oft-solana/layerzero.config.ts (hunk 1 above)
Comment (nazreen) [reply examples/oft-solana/layerzero.config.ts:17]: @St0rmBr3w I think in terms of developer ergonomics, the more glaring thing right now is..

why does the developer need to pass in the solana eid, when they don't need to do the same for the EVM chains?

File: examples/oft-solana/tasks/solana/index.ts (hunk 2 above)
Comment (nazreen) [reply examples/oft-solana/tasks/solana/index.ts:174]: ok, actually, it was never going to return `undefined` if the deployment file was never created, because in `getSolanaDeployment` there's already:

```
    if (!existsSync(filePath)) {
        throw new Error(`Could not find Solana deployment file for eid ${eid} at: ${filePath}`)
    }
```

but sure, I can add another throw in `getOftStoreAddress` 

will apply debugLogger in `getSolanaDeployment` instead

nazreen: Screenshot 2025-03-28 at 11 38 43 AM

ryandgoulding (dismissed):

nazreen:
@ryandgoulding sorry, I pushed a lint fix commit. Would need your review again 🙏

ryandgoulding (approved):

PR #1428, opened 2025-03-28
Sync updates to escrow protection for aptos

abaltes-lz:
Syncing updates from internal monorepo - escrow is made to be not directly withdrawal by oft account or oft object owner (if object deployed).

Also unnecessary dependency on `oapp_compose` for aptos adapter OFT is removed


mok-lz (approved):

shankars99 (changes requested):
~~missing changeset

pls run `pnpm changeset`~~

oops this is an examples change

shankars99 (approved):

PR #1433, opened 2025-04-01
Add superchain ERC20 example

TRileySchwarz:
Created an example ERC20 token that can be used in the mint and burn adapter OFT's that adheres to both the adapter IMintableBurnable.sol interface, as well as OP superchain: 'IERC7802' standard. This has not been audited, but should not be used in production yet.

### DIFF
--- a/examples/mint-burn-oft-adapter/contracts/mocks/SuperChainMintBurnERC20.sol
+++ b/examples/mint-burn-oft-adapter/contracts/mocks/SuperChainMintBurnERC20.sol
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.22;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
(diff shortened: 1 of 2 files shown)


ryandgoulding:
What is the status of this PR?

examples/mint-burn-oft-adapter/contracts/mocks/SuperChainMintBurnERC20.sol
```diff
@@ -0,0 +1,62 @@
...
+
+interface IERC7802 {
+    event CrosschainBurn(address indexed from, uint256 amount, address indexed sender);
+    event CrosschainMint(address indexed to, uint256 amount, address indexed sender);
+
+    function crosschainBurn(address _from, uint256 _amount) external;
+    function crosschainMint(address _to, uint256 _amount) external;
+}
+
+contract SuperChainMintBurnERC20 is ERC20, IERC7802, IMintableBurnable, Ownable {
```
ryandgoulding, line 16: Why is this in `examples` instead of `oft-evm`?  Shouldn't it be in `oft-evm` so folks don't need to copy + paste to use it?

PR #1434, opened 2025-04-03
Add OFTPermitUpgradable

RyanRHall:
This PR adds an `OFTPermitUpgradeable` contract to the codebase, for those building tokens with ERC-2612 support. 

This PR is only an example, I did not write new tests or create the companion `OFTPermit` (non-upgradable) contract. There are probably better ways to accomplish this without creating a copy for every possible ERC20 extension.


shankars99:
please re-open if being worked on

PR #1436, opened 2025-04-03
Upgradable OFT contract with custom token & access control

RyanRHall:
This PR demonstrates some potential changes to the OFT token implementation that allow the token builder to "BYO" token and access control. This should support any burnable/mintable ERC20 implementation and any kind of access control, rather than forcing the token builder to use OZ's ERC20 and OZ's Ownable.

This PR does not have tests, nor is it complete (I only tinkered with the upgradable contracts).


shankars99:
please re-open if being worked on

PR #1438, opened 2025-04-07
🪲 Fix to allow move OFTs to wire a Solana peer

ryandgoulding:
Prior to this fix, it was impossible to wire a Solana peer.  This change makes determining the contractAddress more flexible, by reading it from hardhat-deploy deployments for EVM peers and taking it directly from the configuration for Solana peers.  Supporting other blockchains is future work and considered out of scope for this PR.

### DIFF
--- a/packages/devtools-move/tasks/move/utils/moveVMOftConfigOps.ts
+++ b/packages/devtools-move/tasks/move/utils/moveVMOftConfigOps.ts
@@ -4,26 +4,25 @@ import * as path from 'path'
 import {
     ChainType,
     EndpointId,
-    Stage,
     endpointIdToChainType,
     endpointIdToStage,
     getNetworkForChainId,
+    Stage,
 } from '@layerzerolabs/lz-definitions'
 import { ExecutorOptionType, Options } from '@layerzerolabs/lz-v2-utilities'
 
-import { MessageLibFactory } from '../../../sdk/messageLibFactory'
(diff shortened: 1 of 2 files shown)


packages/devtools-move/tasks/move/utils/moveVMOftConfigOps.ts
```diff
@@ -142,19 +141,42 @@ export function evmAddressToAptos(address: string, eid: string): string {
...
+): Promise<string | null> => {
+    switch (endpointIdToChainType(connection.to.eid)) {
+        case ChainType.EVM:
+            return await getEVMDeploymentAddress(connection)
+        case ChainType.SOLANA:
+            return connection.to.address ?? null
+        default:
+            throw new Error(`Unsupported peer chain type: ${endpointIdToChainType(connection.to.eid)}`)
+    }
+}
```
ravinagill15, line 170: I think we need a case for `ChainType.APTOS` and `ChainType.INITIA` in case we are wiring from movement to aptos or movement to initia for example

ryandgoulding: Agreed, although there are far more places that require changes to support APTOS, INITIA, etc.  As such, I would rather keep that as separate work.  This change is small and makes Solana work;  it does not add a regression for MOVEMENT or APTOS as those chains will not work for other reasons.

ravinagill15 (approved):
🚀

PR #1444, opened 2025-04-09
feat: add solana composer example

St0rmBr3w:
### In this PR:

- creates an OFT composer examples for Raydium swaps on Solana


ziming-zung:
## lz_compose Flow
   1. The offchain executor first retrieves the data from the `lz_compose_types_accounts` PDA to identify which accounts are needed for the `lz_compose_types` instruction
   2. The executor then calls the `lz_compose_types` instruction with these accounts, which returns the complete list of accounts needed for `lz_compose` call
   3. The executor assembles all these accounts and calls `lz_compose`
   4. Inside `lz_compose`, this composer program should:
      - Validate the message is from the expected OFT (stored in composer pda)
      - Perform the Raydium swap via CPI call
      - Clear the compose message state via the `clear_compose` CPI call
      - Emit events

After a thorough review, I have some comments.
1. For the `init_composer` instruction:
    I recommend implementing two clearly defined PDA in this  instruction:
   
     a) The `composer` PDA:
        - Can be derived using a seed pattern such as `[COMPOSER_SEED, oft_pda.key()]`
        - Should store critical data like:
          • The OFT PDA address for validation during compose operations
          • Any configuration
        - This account acts as the main program state and authority
     
     b) The `lz_compose_types_accounts` PDA:
        - Must be derived specifically with `[LZ_COMPOSE_TYPES_SEED, composer.key()]`
        - Try to keep only necessary account addresses, which cannot be deterministically derived at runtime through PDA
     This mode ensures that the offchain executor can correctly discover all required accounts

2. For the `lz_compose_types` instruction:

   - This instruction functions as a view function and should not modify any state
      * All account annotations should reflect this by removing `#[account(mut)]` macro
   - The implementation should ensure that accounts appear in the correct order for both the swap operation and subsequent clearing of the compose state:
     * First retrieve the base accounts needed for swap operations
     * Then append the accounts required by `get_accounts_for_clear_compose`
     * Return a complete, ordered list of accounts that the executor will provide to `lz_compose`

3. For the `lz_compose` instruction:
   - Add some validation checks, like this:
    ```rust
    require(params.from == oft_pda);
    require(params.to == composer_pda);
    ```

St0rmBr3w:
This PR is blocked until the implementation of the `lz_compose_types_with_alt` and `lz_compose_with_alt`

cooldev900:
@St0rmBr3w @ziming-zung
When do you expect this PR could be merged?
I am looking forward to an example like this.

PR #1447, opened 2025-04-14
🧹 make test_quoteOFT virtual

ryandgoulding:
This is so I can override it in an implementation of the test that uses fees.


St0rmBr3w (approved):
LGTM!

PR #1454, opened 2025-04-23
fix bug with double initialization of send and receive library in Solana SDK

DanL0:
We can't blindly init send and receive library without checking if it's initialized already. In Solana this leads to revert every time.


nazreen (dismissed):

packages/ua-devtools-solana/src/oft/sdk.ts
```diff
@@ -180,19 +180,34 @@ export class OFT extends OmniSDK implements IOApp {
...
+        }
+
+        const isReceiveLibraryInitialized = await this.isReceiveLibraryInitialized(eid)
+        if (!isReceiveLibraryInitialized) {
+            instructions.push(
+                oft.initReceiveLibrary({ admin: delegate, oftStore }, eid) // delegate
+            )
+        }
+
+        instructions.push(
```
nazreen, line 201: since these ixns are not wrapped around any conditionals, why not just have them be included in the `instructions` declaration on line 183?

DanL0: I want to preserve original order of instructions

nazreen: ok makes sense, but can you add a comment above the .push to clarify that? as others might wonder the same

nazreen (approved):

PR #1462, opened 2025-04-25
add initia compile and test

shankars99:
## What is added
Certain examples were missing unit tests and compile for initia

## What is not added
Since aptos and initia use a wrapper to build and test. We need testing for that. Support for this does not exist in devtools-move notably the inclusion of --dev flag


bholcomb8 (approved):
LGTM!

PR #1486, opened 2025-05-08
Use native gas estimation in hardhat send command

RyanRHall:
500k gas is too low for x-chain transfers with larger DVNs


St0rmBr3w:
Thank you for catching! We actually took your feedback and are refactoring the entire send task to be more streamlined in #1498

PR #1487, opened 2025-05-08
🧹 Cleanup aptos-move examples for accelerator demo

AlexanderLiteplo:
Adjusting the examples to support aptos out of the box rather than movement


ItsAdel (approved):

St0rmBr3w (approved):
LGTM!

PR #1492, opened 2025-05-12
CI: tell prettier to ignore package.json

nazreen:
Several PRs are right now blocked to a prettier check that's impossible to fix. 

The prettier (reported by eslint) error involves needing to move `overrides` in the example `package.json` from a lower line to a higher line. After doing so and re-running CI, it would then error saying it needs to be moved from the higher line to the lower line.

Research and attempts at resolving through other means were unsuccessful. This is happening for all the examples.

Note: I tried to tell eslint to ignore prettier's rules, but then the run of prettier would cause the CI to abort. It's triggered via the `lint:js` script:

```
"lint:js": "eslint '**/*.{js,ts,json}' && prettier --check .",
```

If no better solution is found, we should merge this in to unblock the other PRs.

Why telling prettier to ignore package.json isn't too bad:
- prettier is just a formatter, we don't lose confidence on correctness by skipping `package.json`
- if the package.json is malformed, it would show up anyways when running any of the other scripts

### DIFF
--- a/examples/lzapp-migration/.prettierignore
+++ b/examples/lzapp-migration/.prettierignore
@@ -8,3 +8,4 @@ out/
 *.yaml
 *.lock
 package-lock.json
+package.json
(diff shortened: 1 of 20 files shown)


shankars99 (approved):

shankars99:
Potentially temporary merge unless a better solution is found

PR #1497, opened 2025-05-13
🪲 validate incorrect wss rpc

shankars99:
The test that validate's incorrect wss rpc should return an object with status: 0, it looks like the provider or something has changed the api and it is now returning 

```
{
    status: 1,
    signal: null,
    output: [ null, null, null ],
    pid: 2390,
    stdout: null,
    stderr: null
}
```

signal, output, stdout, stderr being `null` is now used to test this fail case as the rpc should return `null`

also re-enabling a safe test that was disabled earlier due to the safe hack


bholcomb8 (approved):
LGTM!

PR #1498, opened 2025-05-14
chore: normalize oft send task

St0rmBr3w:
### In this PR:

This PR refactors and unifies our cross-chain “send OFT” task so that both EVM and Solana versions share the same entrypoint, configuration, and output formatting. Rather than maintaining two separate Hardhat tasks, we now have a single `lz:oft:send` task that:

1. Parses a common set of parameters (`srcEid`, `dstEid`, `amount`, `to`, plus optional overrides).
2. Dispatches to either `sendEvm()` or `sendSolana()` based on the source `endpointId`.
3. Uses a shared `DebugLogger` for warnings and clean, professional console output (including tx hashes & LayerZero Scan links).

Screenshot 2025-05-14 at 4 50 58 PM

---

### ✨ What Changed

* **Single Master Task**

  * Introduced `task("lz:oft:send", ...)` in `tasks/sendOFT.ts`.
  * Consolidates both EVM and Solana send logic under one Hardhat command.
  * Enables calling any arbitrary OFT regardless of whether there's a local deployment (optional).

* **Unified Arguments**

  * `srcEid`, `dstEid`, `amount`, `to` are now universal.
  * Optional overrides (`oftAddress`, `oftProgramId`, `tokenProgram`, `computeUnitPriceScaleFactor`) work for both flows.

* **Chain-Type Dispatch**

  * Look up `ChainType` via `endpointIdToChainType(srcEid)`.
  * Call `sendEvm(args, hre)` for EVM, or `sendSolana(args)` for Solana.
  * Throw a clear error if an unsupported chain type is passed.

* **Shared DebugLogger**

  * Prints any override warning (`USING_OVERRIDE_OFT`) when a user overrides the OFT address or program.
  * After execution, always prints:

    * `SENT_VIA_OFT` → “Sent X tokens”
    * `TX_HASH` → the transaction hash
    * `EXPLORER_LINK` → the LayerZero Scan URL

* **Code Organization**

  * `evm/sendEvm.ts` and `solana/sendSolana.ts` now only contain chain-specific plumbing.
  * The new task file handles parameter parsing, dispatch, and final logging.

* **Typesafe Args**

  * `MasterArgs` defines the superset of all parameters.
  * `EvmArgs` and `SolanaArgs` extend it for their own needs.

### ✅ Impact

* **Developer DX**: One command to rule them all—no more “send-evm” vs “send-solana”.
* **Consistency**: Identical logging and UX across chains.
[...]

PR #1503, opened 2025-05-20
Draft: Compose Aptos Move Example

AlexanderLiteplo:
Looking for initial review from @St0rmBr3w to ensure design parity with EVM and Solana examples.

Questions:
1. What params would you like to be passed from src to dest? 
2. Is it ok to hardcode the pool address, LZUSDC address dst chain (as they are right now -- see compose.move) OR would you like pool and token we are swapping to address passed as params to the composer?
3. Would you like the OFT we sending tokens with (currently USDe) to be hardcoded into the send side script or instead be left empty for the user to input?
4. Are there any improvements you would like to suggest?

### DIFF
--- a/examples/compose-aptos-move/scripts/sendOFTWithCompose.ts
+++ b/examples/compose-aptos-move/scripts/sendOFTWithCompose.ts
@@ -100,7 +100,7 @@ async function main() {
     const srcRpcUrl = 'https://base-rpc.publicnode.com' // Base Mainnet RPC
     const srcOftContractAddress = '0x5d3a1Ff2b6BAb83b63cd9AD0787074081a52ef34' // USDe OFT address on Base
     const destEndpointId = 30108 // Aptos Mainnet endpoint ID
-    const aptosComposerAddress = '0x3ee478eec9694d427fcfbcf8209b66ed09879681533cc949a9865479c0e38eae'
+    const aptosComposerAddress = 'YOUR_COMPOSER_ADDRESS'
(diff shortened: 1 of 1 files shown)


examples/compose-aptos-move/scripts/sendOFTWithCompose.ts
```diff
@@ -0,0 +1,177 @@
+import * as fs from 'fs'
```
St0rmBr3w, line 1: Is there anyway you can try to implement consistent runner logic for send like in: https://github.com/LayerZero-Labs/devtools/pull/1498

St0rmBr3w (changes requested):

examples/compose-aptos-move/scripts/sendOFTWithCompose.ts
```diff
@@ -0,0 +1,177 @@
...
+    const provider = new ethers.providers.JsonRpcProvider(srcRpcUrl)
+    const wallet = new ethers.Wallet(privateKey, provider)
+
+    const oftContract = new ethers.Contract(srcOftContractAddress, abiJson, wallet)
+
+    // Ensure the 'to' address is correctly formatted as bytes32
+    const toAddressBytes32 = ethers.utils.hexZeroPad(aptosComposerAddress, 32)
+
+    // Create the options
+    const options = Options.newOptions().addExecutorComposeOption(0, 100000, 0).toHex()
```
St0rmBr3w, line 147: Don't hardcode options

shankars99:

PR #1508, opened 2025-05-21
Use common task names for Aptos Move example

St0rmBr3w:
## Summary
- rename Aptos Move tasks to `lz:oft:aptos:*`
- update package scripts and README to use the new commands
- document compiling Move contracts via `pnpm compile`

## Testing
- `pnpm test` *(fails: command exited (2))*


bholcomb8 (approved):
LGTM, but @AlexanderLiteplo should give this at least a quick glance

PR #1522, opened 2025-05-27
refactor(solana): internalize account checker and priority fee getter

nazreen:
## Changes
- moves duplicated helper functions in `lzapp-migration` and `oft-solana` into `packages/devtools-solana`
  - helper functions moved: `assertAccountInitialized` and `getPrioritizationFees` 
  
I have manually tested affected code paths.  

## Testing

To test the priority fees getter, run

```
npx hardhat lz:solana:get-priority-fees --eid 40168 --address HkqvKCFt5mnuKhLB82UcJQdBtCv6DDgYQyGnPvyefNht  
```

To test the account inited checker, you can run the createOFT command (requires OFT program deployment first), which makes use of the checker:

```
pnpm hardhat lz:oft:solana:create --eid 40168 --program-id <PROGRAM_ID> --only-oft-store true
```


nazreen:
addressed build errors and tested again. ready for review once CI checks pass.

examples/lzapp-migration/tasks/solana/getPrioFees.ts
```diff
@@ -27,7 +26,7 @@ task('lz:solana:get-priority-fees', 'Fetches prioritization fees from the Solana
         devtoolsTypes.string
     )
     .setAction(async ({ eid, address }: GetPrioFeesTaskArgs) => {
-        const { connection } = await deriveConnection(eid)
+        const { connection } = await deriveConnection(eid, true)
```
St0rmBr3w, line 29: Did we normalize deriveConnection from another package? Just trying to understand where the `true` field is coming from

nazreen: That optional field was added to the deriveConnection method in oftSolana's implementation to allow for readOnly mode which will use a random/ephemeral keypair. But in this case, it also switches off the prompt to confirm using the keypair at the default path, which should not be needed since this task is read only.

DanL0 (approved):

PR #1539, opened 2025-05-30
feat(oft-solana): warn when options missing

nazreen:
## Motivation
Prevent send failures when users forget to set `extraOptions` and their OFT has no enforced options configured.

## Changes
- added enforced option checks in `sendEvm` and `sendSolana`
- prompt user if both extra options and enforced options are absent
- added changeset for `oft-solana` example

## Verification
- `pnpm lint:fix`
- `pnpm build` *(fails: forge not found)*
- `pnpm test:local` *(fails: docker not found)*

### DIFF
--- a/examples/oft-solana/tasks/evm/sendEvm.ts
+++ b/examples/oft-solana/tasks/evm/sendEvm.ts
@@ -5,7 +5,7 @@ import { HardhatRuntimeEnvironment } from 'hardhat/types'
 
 import { makeBytes32 } from '@layerzerolabs/devtools'
 import { createGetHreByEid } from '@layerzerolabs/devtools-evm-hardhat'
-import { createLogger } from '@layerzerolabs/io-devtools'
+import { createLogger, promptToContinue } from '@layerzerolabs/io-devtools'
 import { ChainType, endpointIdToChainType, endpointIdToNetwork } from '@layerzerolabs/lz-definitions'
 
 import layerzeroConfig from '../../layerzero.config'
@@ -95,6 +95,22 @@ export async function sendEvm(
         extraOptions: extraOptions ? extraOptions.toString() : '0x',
         composeMsg: composeMsg ? composeMsg.toString() : '0x',
         oftCmd: '0x',
+    }
+
+    if (!extraOptions) {
+        try {
+            const enforced = await oft.enforcedOptions(dstEid, 1)
+            if (!enforced || /^0x0*$/.test(enforced)) {
+                const proceed = await promptToContinue(
+                    'No extra options were included and OFT has no set enforced options. Your quote / send will most likely fail. Continue ?'
+                )
+                if (!proceed) {
+                    throw new Error('Aborted due to missing options')
+                }
+            }
+        } catch (error) {
+            logger.debug(`Failed to check enforced options: ${error}`)
+        }
     }
 
     // 6️⃣ Quote (MessagingFee = { nativeFee, lzTokenFee })
--- a/.changeset/missing-options-check.md
+++ b/.changeset/missing-options-check.md
@@ -1,0 +1,4 @@
+---
+"@layerzerolabs/oft-solana-example": patch
+---
+Add prompt when sending tokens without extra options and no enforced options.
--- a/examples/oft-solana/tasks/solana/sendSolana.ts
+++ b/examples/oft-solana/tasks/solana/sendSolana.ts
@@ -6,7 +6,7 @@ import { TOKEN_PROGRAM_ID } from '@solana/spl-token'
 import { PublicKey } from '@solana/web3.js'
 import bs58 from 'bs58'
 
-import { createLogger } from '@layerzerolabs/io-devtools'
+import { createLogger, promptToContinue } from '@layerzerolabs/io-devtools'
 import { EndpointId, endpointIdToNetwork } from '@layerzerolabs/lz-definitions'
 import { addressToBytes32 } from '@layerzerolabs/lz-v2-utilities'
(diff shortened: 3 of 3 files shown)


examples/oft-solana/tasks/evm/sendEvm.ts
```diff
@@ -97,6 +97,22 @@ export async function sendEvm(
         oftCmd: '0x',
     }
 
+    if (!extraOptions) {
+        try {
+            const enforced = await oft.enforcedOptions(dstEid, 1)
```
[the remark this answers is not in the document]
nazreen: the variable, if created, is more accurately "MSG_TYPE"

nazreen:
blocked by aptos example error

nazreen:
checks now all passing

examples/oft-solana/tasks/common/utils.ts
```diff
@@ -95,6 +98,18 @@ function formatBigIntForDisplay(n: bigint) {
     return n.toLocaleString().replace(/,/g, '_')
 }
 
+function bufferEquals(a: Uint8Array, b: Uint8Array): boolean {
```
DanL0: is it possible to replace these functions with using existing LZ OptionsBuilder TypeScript class?

nazreen: I searched the `@layerzerolabs/lz-v2-utilities` package. There is no utility to check whether a hex string is an empty options type 3.

nazreen: oh turns out this `bufferEquals` helper is totally unnecessary as it's possible to do

```
Buffer.from(optionsBytes).equals()
```

there's a built in `equals` method on buffers.

But I think I also know how to make the Solana options checking slightly simpler.

Thanks for flagging this @DanL0 !

nazreen: simplified

```
export function isEmptyOptionsSolana(optionsBytes: Uint8Array): boolean {
    const optionsHex = Buffer.from(optionsBytes).toString('hex')
    return optionsHex === '0003' // empty options type 3 without 0x prefix
}
```

nazreen: @DanL0

examples/oft-solana/tasks/common/utils.ts
```diff
@@ -95,6 +101,17 @@ function formatBigIntForDisplay(n: bigint) {
...
+    return !optionsHex || optionsHex === '0x' || optionsHex === '0x0003' // 0x0003 is an empty options type 3
+}
+
+export function isEmptyOptionsSolana(optionsBytes?: Uint8Array): boolean {
+    if (!optionsBytes) {
+        return true // Treat undefined or null as empty options
+    }
+    return Buffer.from(optionsBytes).toString('hex') === '0003' // empty options type 3 without 0x prefix
+}
+
```
St0rmBr3w, line 114: Why is this needed? Can't we just pass `Options.newOptions()` from the lz-v2-utilities package?

examples/oft-solana/tasks/evm/sendEvm.ts
```diff
@@ -97,6 +85,27 @@ export async function sendEvm(
         oftCmd: '0x',
     }
 
+    // Check whether there are extra options or enforced options. If not, warn the user.
+    // Read on Message Options: https://docs.layerzero.network/v2/concepts/message-options
+    if (!extraOptions) {
+        try {
+            const enforcedOptions = composeMsg
+                ? await oft.enforcedOptions(dstEid, MSG_TYPE.SEND_AND_CALL)
+                : await oft.enforcedOptions(dstEid, MSG_TYPE.SEND)
```
St0rmBr3w, line 94: Does this actually work? Enforced options aren't actually defined in the IOFT interface. https://github.com/LayerZero-Labs/devtools/blob/main/packages/oft-evm/contracts/interfaces/IOFT.sol

St0rmBr3w (approved):
LGTM! Just minor comment.

examples/oft-solana/tasks/common/utils.ts
nazreen: semantically, it can be clearer, but it's not as helpful imo. 

for Solana, `Options.newOptions().toBytes()` returns a Uint8Array which cannot be directly compared using `==`

e.g.

```javascript
const a = new Uint8Array([1, 2, 3]);
const b = new Uint8Array([1, 2, 3]);
console.log(a === b); // false
```

If we wanted to compare the `optionsBytes` directly, we'd need something like:

```
function uint8ArraysEqual(a: Uint8Array, b: Uint8Array): boolean {
  if (a.length !== b.length) return false;
  for (let i = 0; i < a.length; i++) {
    if (a[i] !== b[i]) return false;
  }
  return true;
}
```

but since the Uint8Array class has `toString('hex') ` in Node.js, I used that to simplify the comparison and not require an additional helper.

for EVM, yes, I can add in `Options.newOptions()` if you think that is better semantically

examples/oft-solana/tasks/evm/sendEvm.ts
nazreen: That's right, that's why I had to change it to use `OFT` (view line 58 on this same file). I initially wanted to leave in `IOFT` in but just merge in `IOAppOptionsType3`s ABI but the merging of ABIs wasn't working. Not sure if you have any general advice on that.

examples/oft-solana/tasks/common/utils.ts
nazreen: @St0rmBr3w bumping up my reply

St0rmBr3w (approved):
LGTM!

PR #1547, opened 2025-06-02
🪲 hyperliquid - skip fork tests when failed vm.createFork()

shankars99:
Since hyperliquid's testnet rpc goes down at times, this causes our CI to fail. A temporary solution before 
we can migrate forking to mainnet rpc - which is considerably more stable (needs hyperliquid to enable write precompiles and we will use a core spot deployed by someone else) 

I would rather not use `deployCodeTo` since `etch` is more meant for precompiles.


AlexanderLiteplo (approved):
lgtm

PR #1552, opened 2025-06-03
docs(solana examples): update keygen related instructions

nazreen:
The current instruction to run solana-keygen new is redundant as keypair generation will be done by anchor keys sync if keys do not yet exist.


bholcomb8 (approved):
LGTM!

PR #1561, opened 2025-06-10
DEVREL-518 Feat: add send task to oft

St0rmBr3w:
### In this PR:
- adds a normalized send task like that in examples/oft-solana
- proof of work:  
  - oft: https://testnet.layerzeroscan.com/tx/0xe06e49eadf715a6fcdab8dade2875ad7f2c80fbd666d71da33a1c203e6ab2ed7
  - maba: https://testnet.layerzeroscan.com/tx/0x1bcb9825c20a952b18234b8a7be9b0d0809ff598d57218d507661ebda520fc4e


examples/oft/package.json
```diff
@@ -48,6 +50,7 @@
     "@types/chai": "^4.3.11",
     "@types/mocha": "^10.0.6",
     "@types/node": "~18.18.14",
+    "bs58": "^5.0.0",
```
shankars99: do we want to keep the solana stuff out of this oft and only in the solana oft example? i think it is safe to assume that everyone that has a peer on solana is already on the solana oft example and trickling in the aptos-move send scripts into every example feels unncessary because they have a evm<>x-vm example

examples/oft/tasks/sendOFT.ts
```diff
@@ -0,0 +1,88 @@
+import { task, types } from 'hardhat/config'
+import { HardhatRuntimeEnvironment } from 'hardhat/types'
+
+import { ChainType, endpointIdToChainType, endpointIdToNetwork } from '@layerzerolabs/lz-definitions'
+
+import { EvmArgs, sendEvm } from './sendEvm'
+import { SendResult } from './types'
+import { DebugLogger, KnownOutputs, KnownWarnings, getBlockExplorerLink } from './utils'
+
+interface MasterArgs {
```
shankars99, line 11: why not define the interface in `types.ts`?

examples/oft/package.json
St0rmBr3w: Yeah I think that's valid, also can just use addressToBytes32() anyways rather than makeBytes() to save some time on conversion.

examples/oft/tasks/sendOFT.ts
St0rmBr3w: I chose to define it in the task itself so that first time viewers had a bit of a clearer idea of where the params came from. Could move to types.rs.

nazreen: I think the interface for a task's Args should be kept local in the task itself

putting in a types.ts makes sense if it's reused across various files in differing situations

DanL0 (approved):
Looks good.

Two improvement ideas for future:
1. Support OFT adapter
2. Support dynamic oapp-config parameter, because we accept such parameter in other commands

I have implemented 2 and tested:
```diff
diff --git a/tasks/sendEvm.ts b/tasks/sendEvm.ts
index 4c9118d..d260e37 100644
--- a/tasks/sendEvm.ts
+++ b/tasks/sendEvm.ts
@@ -2,15 +2,14 @@ import { BigNumber, ContractTransaction } from 'ethers'
 import { parseUnits } from 'ethers/lib/utils'
 import { HardhatRuntimeEnvironment } from 'hardhat/types'
 
-import { createGetHreByEid } from '@layerzerolabs/devtools-evm-hardhat'
+import { createGetHreByEid, OmniPointHardhat } from '@layerzerolabs/devtools-evm-hardhat'
 import { createLogger } from '@layerzerolabs/io-devtools'
 import { ChainType, endpointIdToChainType, endpointIdToNetwork } from '@layerzerolabs/lz-definitions'
 import { addressToBytes32 } from '@layerzerolabs/lz-v2-utilities'
 
-import layerzeroConfig from '../layerzero.config'
-
 import { SendResult } from './types'
 import { DebugLogger, KnownErrors, getLayerZeroScanLink } from './utils'
+import path from 'path'
 
 const logger = createLogger()
 
@@ -19,6 +18,7 @@ export interface EvmArgs {
     dstEid: number
     amount: string
     to: string
+    oappConfig: string
     minAmount?: string
     extraOptions?: string
     composeMsg?: string
@@ -26,7 +26,7 @@ export interface EvmArgs {
 }
 
 export async function sendEvm(
-    { srcEid, dstEid, amount, to, minAmount, extraOptions, composeMsg, oftAddress }: EvmArgs,
+    { srcEid, dstEid, amount, to, minAmount, extraOptions, composeMsg, oftAddress, oappConfig }: EvmArgs,
     hre: HardhatRuntimeEnvironment
 ): Promise<SendResult> {
     if (endpointIdToChainType(srcEid) !== ChainType.EVM) {
@@ -51,8 +51,9 @@ export async function sendEvm(
     if (oftAddress) {
         wrapperAddress = oftAddress
     } else {
-        const { contracts } = typeof layerzeroConfig === 'function' ? await layerzeroConfig() : layerzeroConfig
-        const wrapper = contracts.find((c) => c.contract.eid === srcEid)
+        const layerZeroConfig = (await import(path.resolve('./', oappConfig))).default
+        const { contracts } = typeof layerZeroConfig === 'function' ? await layerZeroConfig() : layerZeroConfig
+        const wrapper = contracts.find((c: { contract: OmniPointHardhat }) => c.contract.eid === srcEid)
         if (!wrapper) throw new Error(`No config for EID ${srcEid}`)
         wrapperAddress = wrapper.contract.contractName
             ? (await srcEidHre.deployments.get(wrapper.contract.contractName)).address
diff --git a/tasks/sendOFT.ts b/tasks/sendOFT.ts
index 6875e27..8e2f9d5 100644
--- a/tasks/sendOFT.ts
+++ b/tasks/sendOFT.ts
@@ -27,6 +27,12 @@ task('lz:oft:send', 'Sends OFT tokens cross‐chain from EVM chains')
     .addParam('dstEid', 'Destination endpoint ID', undefined, types.int)
     .addParam('amount', 'Amount to send (human readable units, e.g. "1.5")', undefined, types.string)
     .addParam('to', 'Recipient address (20-byte hex for EVM)', undefined, types.string)
+    .addOptionalParam(
+        'oappConfig',
+        'Path to the LayerZero config file',
+        'layerzero.config.ts',
+        types.string
+    )
     .addOptionalParam(
         'minAmount',
         'Minimum amount to receive in case of custom slippage or fees (human readable units, e.g. "1.5")',
```

examples/oft/tasks/sendOFT.ts
```diff
@@ -0,0 +1,88 @@
...
+    .addParam('amount', 'Amount to send (human readable units, e.g. "1.5")', undefined, types.string)
+    .addParam('to', 'Recipient address (20-byte hex for EVM)', undefined, types.string)
+    .addOptionalParam(
+        'minAmount',
+        'Minimum amount to receive in case of custom slippage or fees (human readable units, e.g. "1.5")',
+        undefined,
+        types.string
+    )
+    .addOptionalParam(
+        'extraOptions',
```
AlexanderLiteplo: Are we expecting people to paste options bytes into their terminal? I think it would be more useful to leave a commented out section in the code where people can construct extraOptions if they want. Also it would be very helpful to have a dedicated param for executorReceiveOptions where people can put enter their gas e.g.: `--executorReceiveOptionsGas 80000`. Executor gas is the main use case of options (especially for hackers) so It would be nice to not have to manually encode that somehow and then paste the encoding into terminal. This would also make testing easy and allow builders to try a bunch of different receive gas so that they can have an optimal config setting.

AlexanderLiteplo: It would be useful to have optional gas params for all options actually.

AlexanderLiteplo (dismissed):

examples/oft/tasks/sendOFT.ts
St0rmBr3w: Resolved this.

nazreen (approved):

PR #1569, opened 2025-06-13
feat: enabling custom branches on create-lz-oapp 

shankars99:
## What is added
This PR adds in `--branch` tag support for `create-lz-oapp` that takes in both refs and fullpath urls (`/tree/` and `/blob/`) and `--base-repository` that takes in a base url (ex: <https://github.com/layerzero-labs/devtools.git>)

The current `create-lz-oapp` lets you use a different ref for existing examples. However this may not be ideal as you might want to test out examples before we merge them into main. 

## Testing locally: 
On CI the tests run grab the latest `create-lz-oapp` package from a locally published `npm registry`. You can simulate this locally by running `pnpm link create-lz-oapp -w && pnpm i` from devtools root or `cd packages/create-lz-oapp && pnpm link -w`

## Changes that had to be made
Extracting out the examples into a file `examples.json` - this helps with imports (importing `ts` files (`config.ts` from other branches and compiling them is delicate) and as a side effect also improves maintenance in the future as our examples grow in size. 

`experimental` in the new config file is left as an array as it can be loaded with more than 1 experimental flags, which allows for varied permutations. 

I created an example repo `test/omnicall-new-createlzoapp` based on the `omni-call` branch : <https://github.com/LayerZero-Labs/devtools/blob/test/omnicall-new-createlzoapp/packages/create-lz-oapp/src/examples.json> with the updated example config file

(note: root `turbo.json` can also have some of the `LZ_ENABLE_EXPERIMENTAL` flags cleared)

Usage example:
```
$ LZ_ENABLE_EXPERIMENTAL_OMNI_CALL_EXAMPLE=1 pnpm create-lz-oapp --branch test/omnicall-new-createlzoapp

info:    [create-lz-oapp] Cloning repository: https://github.com/LayerZero-Labs/devtools.git and ref: feat/ovault-evm
Cloning into '/tmp/lz-oapp-1750686595032'...
info:    [create-lz-oapp] Cleaning up temporary directory: /tmp/lz-oapp-1750686595032
✔ Where do you want to start your project? … ./my-lz-oapp
? Which example would you like to use as a starting point? › - Use arrow-keys. Return to submit.
❯   OApp
    OFT
    OFTAdapter
    ONFT721
    OVault (EVM)
```

```
$ LZ_ENABLE_EXPERIMENTAL_OMNI_CALL_EXAMPLE=1 pnpm create-lz-oapp --branch https://github.com/LayerZero-Labs/devtools/tree/test/omnicall-new-createlzoapp

info:    [create-lz-oapp] Cloning repository: https://github.com/LayerZero-Labs/devtools.git and ref: feat/ovault-evm
Cloning into '/tmp/lz-oapp-1750686543798'...
info:    [create-lz-oapp] Cleaning up temporary directory: /tmp/lz-oapp-1750686543798
✔ Where do you want to start your project? … ./my-lz-oapp
? Which example would you like to use as a starting point? › - Use arrow-keys. Return to submit.
❯   OApp
    OFT
    OFTAdapter
    ONFT721
    OVault (EVM)
```

Todo: There maybe some issues with pathing that I will look into


package.json
```diff
@@ -41,6 +41,9 @@
     "hardhat-deploy": "^0.12.1",
     "rustbn.js": "^0.3.0"
   },
+  "dependencies": {
```
shankars99, line 44: This is required for the bats test to pass

EWCunha (commented):
Great job! Your work is very appreciated!

EWCunha (approved):

packages/create-lz-oapp/src/config.ts
```diff
@@ -1,220 +1,109 @@
...
+        const parts = url.split('/blob/')
+        if (parts.length >= 2) {
+            const ref = parts[1]
+            if (ref) {
+                return ref
+            }
+        }
+    }
+
+    return url
```
ItsAdel, line 33: i think throwing here makes more sense. if the function is expected to extract and return a branch and input is url, it never makes sense that input is the branch name.

packages/create-lz-oapp/src/config.ts
```diff
@@ -1,220 +1,109 @@
...
+}
+
+export const getExamples = async (branch?: string, logLevel = 'info'): Promise<Example[]> => {
+    const logger = createModuleLogger('create-lz-oapp', logLevel)
     /**
      * To enable example development in a custom repository
      * we open the repository URL field to be taken from the environment
      */
-    const repository = process.env.LAYERZERO_EXAMPLES_REPOSITORY_URL || 'https://github.com/LayerZero-Labs/devtools.git'
+    const repository = 'https://github.com/LayerZero-Labs/devtools.git'
```
ItsAdel: Maybe leave this in incase someone is working on their own forked version of devtools?

packages/create-lz-oapp/src/config.ts
```diff
@@ -1,220 +1,109 @@
...
+    const filteredExamples = oapp_examples.filter((example) => {
+        if (!example.experimental) {
+            return true
+        }
+        return example.experimental.some((key) => {
+            const value = process.env[key]
+            if (value === undefined) {
+                return false
+            }
+            return value === 'true' || value === '1' || value === 'yes'
```
ItsAdel, line 101: it could be any value and will show up in the cli. this is causing a bit of a disparity but okay i guess to enforce this.

packages/create-lz-oapp/src/examples.json
ItsAdel, line 1: Seems like this is something people will miss updating when making more examples. We should have some sort of checklist / references in other places of the codebase so we remember to do this.

packages/create-lz-oapp/src/config.ts
shankars99: that would be for direct urls

packages/create-lz-oapp/src/examples.json
shankars99: i agree. we could have formal docs on this

packages/create-lz-oapp/src/config.ts
shankars99: jantools was supposed to enforce this. the tests have it against `1` and `yes`

packages/create-lz-oapp/src/config.ts
shankars99: good idea

shankars99: and done

ItsAdel (approved):
lgtm

PR #1580, opened 2025-06-19
INT-294: Added upgradeable MABA

EWCunha:


packages/oft-evm-upgradeable/contracts/oft/MintBurnOFTAdapterUpgradeable.sol
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * @dev Constructor for initializing the contract with token and endpoint addresses.
+     * @param _token The address of the token.
+     * @param _lzEndpoint The address of the LayerZero endpoint.
+     * @param _minterBurner The address of the contract responsible for minting and burning tokens.
+     */
+    constructor(
+        address _token,
+        address _lzEndpoint,
+        IMintableBurnable _minterBurner
```
ItsAdel, line 37: shouldn't this also be part of the initialize?

imo only the core LZ part `_lzEndpoint` should be in the constructor.

both `_token` and `_minterBurner` should be in the initialize i think

The initialize function correctly sets up this state in the proxy. If `_token` and `_minterBurner` were only in the constructor, their values would be lost upon upgrade.

ItsAdel: discussed offline

ItsAdel (approved):
lgtm

shankars99 (approved):

PR #1584, opened 2025-06-19
feat: Synchronous Vault Composers

shankars99:
All PRs related to OVaults will be merged into this branch and treat it as the `main` branch. 
Post audit this branch will be merged into `main`


examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+ * @notice There are 3 supported modes that correspond to <amount>, <gas>, <value> in the forge script below:
+ * @notice forge script script/SendScript.s.sol --private-key $PRIVATE_KEY --rpc-url $RPC_URL_BASE_TESTNET --sig "exec(address,string,uint256,uint256,uint128,uint128)" $SHARE_OFT_BASE_SEP  "base-sep" 1ether 0 50000 [--broadcast]
+0  0.000025ether --broadcast
+ * @notice 1. Send to HyperEVM 
+                - true, false, false
+ * @notice 2. Send to HyperCore 
+                - true, true, false
+ * @notice 3. Send to HyperCore + Fund HyperCore with HYPE 
+                - true, true, true
+ */
```
St0rmBr3w: Are these supposed to say HyperEVM and HyperCore still?

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+            dstEid,
+            _lzComposeGas,
+            _lzComposeValue
+        );
+
+        vm.startBroadcast();
+
+        MessagingFee memory msgFee = quoteSend(sendParam);
+
+        _send(sendParam, msgFee);
```
St0rmBr3w: NIT: why do we underscore the function _send when it's not passed as an argument?

examples/ovault-evm/contracts/MyOVault.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.22;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
+import { OVault } from "@layerzerolabs/ovault-evm/contracts/OVault.sol";
+import { OFTAdapter } from "@layerzerolabs/oft-evm/contracts/OFTAdapter.sol";
+
+contract MyOVault is OVault {
+    constructor(string memory name, string memory symbol, address asset) OVault(name, symbol, asset) {}
```
St0rmBr3w: NIT: for consistency shouldn't this follow the underscoring of arguments?

packages/ovault-evm/contracts/OVault.sol
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * The drawback of this approach is that the virtual shares do capture (a very small) part of the value being accrued
+ * to the vault. Also, if the vault experiences losses, the users try to exit the vault, the virtual shares and assets
+ * will cause the first user to exit to experience reduced losses in detriment to the last users that will experience
+ * bigger losses. Developers willing to revert back to the pre-v4.9 behavior just need to override the
+ * `_convertToShares` and `_convertToAssets` functions.
+ *
+ * To learn more, check out our xref:ROOT:erc4626.adoc[ERC-4626 guide].
+ * ====
+ */
```
St0rmBr3w: Is the purpose of this comment to just surface the ERC4626 comment underneath? If so, shouldn't we redirect users there and for the purposes of our contract simplify and say that OVault is ERC4626 and inherits all the same risks and attack vectors 4626 vaults share? Just seems strange to mass copy a comment like this from a known contract.

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -0,0 +1,446 @@
...
+        } catch (bytes memory errMsg) {
+            failedMessages[_guid] = FailedMessage(oft, sendParam, _refundOFT, refundSendParam, msg.value);
+            emit OVaultError(_guid, oft, errMsg); /// @dev Since the ovault can revert with custom errors, the error message is valuable
+            emit lzComposeInFailedState(bytes4(IOVaultComposer.OVaultError.selector), errMsg);
+            return;
+        }
+
+        /// @dev Try sending the vault out tokens to the receiver on the target chain (can also be the HUB chain)
+        try this.send{ value: msg.value }(oft, sendParam) {
+            emit Sent(_guid, oft);
```
St0rmBr3w: NIT: would it make more sense to include this event in the send implementation rather than declaring it here?

examples/ovault-evm/script/SendScript.s.sol
shankars99: `_send` is an internal function

examples/ovault-evm/script/SendScript.s.sol
shankars99: definitely not lol

packages/ovault-evm/contracts/OVault.sol
shankars99: Yep, i can change the comment to something that conveys that

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: that `Sent` is to indicate that the `lzCompose` succeeded. The others - `refund`, `retry`, `retryWithSwap` have their own events

examples/ovault-evm/script/ExecutorCompose302.s.sol
```diff
@@ -0,0 +1,48 @@
...
+            _index,
+            _message,
+            _extraData,
+            _gasLimit
+        );
+        vm.stopBroadcast();
+    }
+}
+
+interface IExecutor {
```
yota-toyama-lz: Why don't we import this from the `@layerzerolabs/lz-evm-messagelib-v2` package?

examples/ovault-evm/script/ExecutorCompose302.s.sol
```diff
@@ -0,0 +1,48 @@
...
+
+    function run() public {
+        address _from = 0x130853D8212878F8b7Dda6D98e4cB52E09899485;
+        address _to = 0xD06487aCfebAb0F73300e726f0E5f341F4F2B510;
+        bytes32 _guid = 0xb8be4ae5b90dfd06ad6363fd5736d08a9e789137bab4b094d9395556f07d0d3a;
+        uint16 _index = 0;
+        bytes
+            memory _message = hex"000000000000000700009d280000000000000000000000000000000000000000000000000de0b6b3a7640000000000000000000000000000c632389d59e494750920161960c9e5867c59db1800000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000009d35000000000000000000000000c632389d59e494750920161960c9e5867c59db180000000000000000000000000000000000000000000000000de0b6b3a7640000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000120000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000";
```
yota-toyama-lz: Are these addresses and binary data simply random values? At least, the executor address seems to be the one deployed on a testnet.

St0rmBr3w:
@shankars99 can we remove the pending audit label?

examples/ovault-evm/contracts/MyOVault.sol
```diff
@@ -0,0 +1,19 @@
...
+}
+
+contract MyShareOFTAdapter is OFTAdapter {
+    constructor(
+        address _token,
+        address _lzEndpoint,
+        address _delegate
+    ) OFTAdapter(_token, _lzEndpoint, _delegate) Ownable(_delegate) {}
+}
```
St0rmBr3w: Why is there a MyShareOFT and also a MyShareOFTAdapter in the same project?

St0rmBr3w: IIRC the share is auto created via the Vault deployment so why do we need a MyShareOFT.sol?

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+    /**
+     * @notice Handles LayerZero compose operations for vault transactions with automatic refund functionality
+     * @dev This composer is designed to handle refunds to an EOA address and not a contract
+     * @dev Any revert in handleCompose() causes a refund back to the src EXCEPT for InsufficientMsgValue
+     * @param _composeCaller The OFT contract address used for refunds, must be either ASSET_OFT or SHARE_OFT
+     * @param _guid LayerZero's unique tx id (created on the source tx)
+     * @param _message Decomposable bytes object into [composeHeader][composeMessage]
+     */
+    function lzCompose(
+        address _composeCaller, // The OFT used on refund, also the vaultIn token.
```
St0rmBr3w: This might make more sense to be `_composeSender` to match sendCompose interface, or `_sendComposeCaller` to not confuse with the caller of `lzCompose`.

Alternatively, add more descriptive code comments that this is the caller of `endpoint.sendCompose` from the OFT.

examples/ovault-evm/contracts/MyOVault.sol
shankars99: `ShareOFT` is meant for share token on non hub networks. 
`ShareOFTAdapter` is meant for the share OFT on the hub network. 

the deploy scripts handle the checking if hub or not.

packages/ovault-evm/.gitignore
```diff
@@ -0,0 +1,13 @@
...
+# artifacts;  ignore all files except the local contract artifacts.
+artifacts/*
+# !artifacts/Fee.sol/
+# !artifacts/IFee.sol/
+# !artifacts/IOFT.sol/
+# !artifacts/OFTComposeMsgCodec.sol/
+# !artifacts/OFTMsgCodec.sol/
+# !artifacts/OFT.sol/
+# !artifacts/OFTAdapter.sol/
+# !artifacts/OFTCore.sol/
```
yota-toyama-lz: Why do we keep comments of those?

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+
+import { Script, console } from "forge-std/Script.sol";
+import { OFT } from "@layerzerolabs/oft-evm/contracts/OFT.sol";
+import { OFTComposeMsgCodec } from "@layerzerolabs/oft-evm/contracts/libs/OFTComposeMsgCodec.sol";
+
+/**
+ * @title Lets the user send an LZ OFT transfer to transfer an amount of OFT from a source EVM chain to OVault testnet or mainnet
+ * @notice There are 3 supported modes that correspond to <amount>, <gas>, <value> in the forge script below:
+ * @notice forge script script/SendScript.s.sol --private-key $PRIVATE_KEY --rpc-url $RPC_URL_BASE_TESTNET --sig "exec(address,string,uint256,uint256,uint128,uint128)" $SHARE_OFT_BASE_SEP  "base-sep" 1ether 0 50000 [--broadcast]
+0  0.000025ether --broadcast
```
yota-toyama-lz: Are these lines correctly formatted?

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+ * @notice 1. Send to HyperEVM 
+                - true, false, false
+ * @notice 2. Send to HyperCore 
+                - true, true, false
+ * @notice 3. Send to HyperCore + Fund HyperCore with HYPE 
+                - true, true, true
```
yota-toyama-lz: Is Hyperliquid stuff related to this script?

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    function setUp() public {
+        chainIdToSrcEid[421614] = 40231; // arb sep
+        chainIdToSrcEid[11155420] = 40232; // opt sep
+        chainIdToSrcEid[84532] = 40245; // base sep
+
+        chainNameToDstEid["arb-sep"] = 40231; // arb sep
+        chainNameToDstEid["opt-sep"] = 40232; // opt sep
+        chainNameToDstEid["base-sep"] = 40245; // base sep
+        chainNameToDstEid["bad-eid"] = 100000; // bad eid
```
yota-toyama-lz: What is this `bad-eid` for?

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: yeah changed the name

examples/ovault-evm/contracts/MyOVault.sol
St0rmBr3w: Might be worth adding code comments explaining that in the MyShareOFT and MyShareOFTAdapter source file that Adapter is meant for Hub chain and ShareOFT is meant for spokes.

shankars99:
> @shankars99 can we remove the pending audit label?

It's ongoing, we're waiting on the second audit. Once it's back I'll take it off and we can merge

examples/ovault-evm/contracts/MyOVault.sol
shankars99: Oh yeah I haven't focused on the example and readme yet. This PR is mostly to get the packages in and the examples got added cuz I had to create one to deploy contracts for front end.

But yeah we'll get to this once packages are in

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+
+import { AddressCast } from "@layerzerolabs/lz-evm-protocol-v2/contracts/libs/AddressCast.sol";
+
+import { OptionsBuilder } from "@layerzerolabs/oapp-evm/contracts/oapp/libs/OptionsBuilder.sol";
+
+import { IOFT, SendParam, OFTReceipt, MessagingReceipt, MessagingFee } from "@layerzerolabs/oft-evm/contracts/interfaces/IOFT.sol";
+
+import { Script, console } from "forge-std/Script.sol";
+import { OFT } from "@layerzerolabs/oft-evm/contracts/OFT.sol";
+import { OFTComposeMsgCodec } from "@layerzerolabs/oft-evm/contracts/libs/OFTComposeMsgCodec.sol";
```
yota-toyama-lz: Isn't this unused?

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+        return myOFT_SRC.quoteSend(_sendParam, false);
+    }
+
+    function _send(SendParam memory _sendParam, MessagingFee memory _fee) internal {
+        myOFT_SRC.send{ value: _fee.nativeFee }(_sendParam, _fee, msg.sender);
+    }
+
+    function addressToBytes32(address _addr) internal pure returns (bytes32) {
+        return bytes32(uint256(uint160(_addr)));
+    }
```
yota-toyama-lz:
```suggestion
```

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+        uint256 _tokenAmount,
+        uint256 _minAmount,
+        uint32 _dstEid,
+        uint128 _lzComposeGas,
+        uint128 _lzComposeValue
+    ) public view returns (SendParam memory sendParam) {
+        bytes memory options;
+        bytes memory composeMsg = "";
+        uint32 dstEid = _dstEid;
+        bytes32 to = addressToBytes32(_receiver);
```
yota-toyama-lz:
```suggestion
        bytes32 to = _receiver.toBytes32();
```

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+                dstEid: dstEid,
+                to: to,
+                amountLD: _tokenAmount,
+                minAmountLD: _minAmount,
+                extraOptions: hex"",
+                composeMsg: composeMsg,
+                oftCmd: hex""
+            });
+            composeMsg = abi.encode(hopSendParam);
+            to = addressToBytes32(_composer);
```
yota-toyama-lz:
```suggestion
            to = _composer.toBytes32();
```

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+ * @notice 3. Send to HyperCore + Fund HyperCore with HYPE 
+                - true, true, true
+ */
+contract SendScript is Script {
+    using AddressCast for address;
+    using OptionsBuilder for bytes;
+
+    address payable public address_composer;
+
+    OFT public myOFT_SRC;
```
yota-toyama-lz: Is there any reason why this variable named differently than the others? How about `mySrcOFT`?

examples/ovault-evm/script/SendScript.s.sol
```diff
@@ -0,0 +1,144 @@
...
+    address public address_src_OFT;
```
yota-toyama-lz: Is this state variable uninitialized?

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// ========================== GLOBAL VARIABLE FUNCTIONS =====================================
+    function VAULT() external view returns (IERC4626);
+
+    function ASSET_OFT() external view returns (address);
+    function ASSET_ERC20() external view returns (address);
+    function SHARE_OFT() external view returns (address);
+    function SHARE_ERC20() external view returns (address);
+
+    function ENDPOINT() external view returns (address);
+    function VAULT_EID() external view returns (uint32);
```
yota-toyama-lz, line 37: I don't see many of these conventions of global variable functions named in the capital case. But at the same time, I'm not familiar with the Solidity code either... Should we use the normal camel case, or keep these names to distinguish them from normal accessor methods?

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @dev This function can be overridden to implement custom quoting logic
+     */
+    function quoteSend(
+        address oft,
+        uint256 vaultInAmount,
+        SendParam memory sendParam
+    ) external view returns (MessagingFee memory);
+
+    /// ========================== Receive =====================================
+    receive() external payable;
```
yota-toyama-lz, line 74: When do we need to pay Ether to OVault composers? Does it need to be in this interface? I thought OVault composers are more like pass-through contracts...

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -0,0 +1,72 @@
...
+pragma solidity ^0.8.22;
+
+import { IERC4626 } from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+
+import { IOAppComposer } from "@layerzerolabs/oapp-evm/contracts/oapp/interfaces/IOAppComposer.sol";
+import { SendParam, MessagingFee } from "@layerzerolabs/oft-evm/contracts/interfaces/IOFT.sol";
+
+interface IVaultComposerSync is IOAppComposer {
+    /// ========================== EVENTS =====================================
+    event Sent(bytes32 indexed guid); // 0x27b5aea9
```
yota-toyama-lz, line 11: Are these numbers in comments selectors? Why do we put them?

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+     */
+    constructor(address _vault, address _assetOFT, address _shareOFT) {
+        VAULT = IERC4626(_vault);
+
+        ASSET_OFT = _assetOFT;
+        ASSET_ERC20 = IOFT(ASSET_OFT).token();
+        SHARE_OFT = _shareOFT;
+        SHARE_ERC20 = IOFT(SHARE_OFT).token();
+
+        ENDPOINT = address(IOAppCore(ASSET_OFT).endpoint());
```
yota-toyama-lz, line 64: Is it okay to cache the LayerZero endpoint of an OApp?

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+     * @dev This function can be overridden to implement custom quoting logic
+     */
+    function quoteSend(
+        address _oft,
+        uint256 _vaultInAmount,
+        SendParam memory _sendParam
+    ) external view virtual returns (MessagingFee memory) {
+        if (_oft == ASSET_OFT) {
+            _sendParam.amountLD = VAULT.previewDeposit(_vaultInAmount);
+        } else {
```
yota-toyama-lz: Should we also check the `_oft == SHARE_OFT` and revert otherwise for the usability as this function is marked as `external`? IIUC, we return technically wrong fees when we specify a random `_oft`.

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+
+        /// @dev try...catch to handle the compose operation. if it fails we refund the user
+        try this.handleCompose{ value: msg.value }(_composeCaller, composeFrom, composeMsg, amount) {
+            emit Sent(_guid);
+        } catch (bytes memory _err) {
+            /// @dev A revert where the msg.value passed is lower than the min expected msg.value is handled separately
+            /// This is because it is possible to re-trigger from the endpoint the compose operation with the right msg.value
+            if (bytes4(_err) == InsufficientMsgValue.selector) {
+                assembly {
+                    revert(add(32, _err), mload(_err))
```
yota-toyama-lz, line 120: So is this only the best way to re-throw errors in Solidity as we don't have any way to catch custom errors?

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// ========================== Error Messages =====================================
+    error ShareOFTNotAdapter(address shareOFT); // 0xfc1514ae
+    error ShareTokenNotVault(address shareERC20, address vault); // 0x0e178ab6
+    error AssetTokenNotVaultAsset(address assetERC20, address vaultAsset); // 0xba9d665f
+
+    error OnlyEndpoint(address caller); // 0x91ac5e4f
+    error OnlySelf(address caller); // 0xa19dbf00
+    error OnlyValidComposeCaller(address caller); // 0x84fb3f0d
+
+    error InsufficientMsgValue(uint256 expectedMsgValue, uint256 actualMsgValue); // 0x7cb769dc
```
yota-toyama-lz, line 23: In my understanding, this is not always "insufficient" `msg.value` error. Should we rename it like `Invalid*` or create another error type?

```suggestion
    error InvalidMsgValue(uint256 expectedMsgValue, uint256 actualMsgValue); // 0x7cb769dc
```

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+     * @param _oft The OFT contract address used for refunding
+     * @param _message The original message that was sent
+     * @param _amount The amount of tokens to refund
+     * @param _refundAddress Address to receive the refund
+     */
+    function _refund(address _oft, bytes calldata _message, uint256 _amount, address _refundAddress) internal virtual {
+        /// @dev Extracted from the _message header. Will always be part of the _message since it is created by lzReceive
+        SendParam memory refundSendParam;
+        refundSendParam.dstEid = OFTComposeMsgCodec.srcEid(_message);
+        refundSendParam.to = OFTComposeMsgCodec.composeFrom(_message);
```
yota-toyama-lz, line 341: Should we also reuse the `to` address passing it in the caller?

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+     * @param _assetOFT The address of the asset OFT (Omnichain Fungible Token) contract
+     * @param _shareOFT The address of the share OFT contract (must be an adapter)
+     *
+     * Requirements:
+     * - Share token must be the vault itself
+     * - Asset token must match the vault's underlying asset
+     * - Share OFT must be an adapter (approvalRequired() returns true)
+     */
+    constructor(address _vault, address _assetOFT, address _shareOFT) {
+        VAULT = IERC4626(_vault);
```
yota-toyama-lz, line 57: Do we need to intake the `_vault` argument separately from the `_shareOFT` argument? Maybe only for me, but I feel like it's a bit confusing...

packages/ovault-evm/contracts/VaultComposerSync.sol
yota-toyama-lz, line 64: Also, should we check the endpoint identity between an asset OFT and a share OFT?

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+
+    /**
+     * @dev Internal function that deposits assets and sends shares to another chain
+     * @param _depositor The depositor (bytes32 format to account for non-evm addresses)
+     * @param _assetAmount The number of assets to deposit
+     * @param _sendParam Parameter that defines how to send the shares
+     * @param _refundAddress Address to receive excess payment of the LZ fees
+     * @notice This function first deposits the assets to mint shares, validates the shares meet minimum slippage requirements,
+     *         then sends the minted shares cross-chain using the OFT (Omnichain Fungible Token) protocol
+     * @notice The _sendParam.amountLD is updated to the actual share amount minted, and minAmountLD is reset to 0 for the send operation
```
yota-toyama-lz, line 183: Should we copy or move this comment to the external function of `depositAndSend` as well?

I'm also not sure if we should expose this `SendParam` arguments directly in these two `external` functions or decompose them into separate arguments. as the integrity of the parameters is guaranteed by us but not by the callers of the functions. For example, the `sendParam.amountLD` can be an arbitrary number.

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+        if (ASSET_ERC20 != address(VAULT.asset())) {
+            revert AssetTokenNotVaultAsset(ASSET_ERC20, address(VAULT.asset()));
+        }
+
+        /// @dev ShareOFT must be an OFT adapter. We can infer this by checking 'approvalRequired()'.
+        /// @dev burn() on tokens when a user sends changes totalSupply() which the asset:share ratio depends on.
+        if (!IOFT(SHARE_OFT).approvalRequired()) revert ShareOFTNotAdapter(SHARE_OFT);
+
+        /// @dev Approve the vault to spend the share and asset tokens held by this contract
+        IERC20(SHARE_ERC20).approve(_vault, type(uint256).max);
```
yota-toyama-lz: Why do we need this approval explicitly? The contract of ERC-4626 itself implements ERC-20 but they are not different contracts. https://eips.ethereum.org/EIPS/eip-4626 So can't they do whatever they want to with the shares? Is it just because that's how they are implemented in OpenZeppelin?

packages/ovault-evm/contracts/VaultComposerSync.sol
yota-toyama-lz: @TRileySchwarz and I discussed this offline. It's technically not necessary to be passed in but good to have as it allows extra validation to ensure that we passed the correct OVault and share OFT pair.

packages/ovault-evm/contracts/VaultComposerSync.sol
yota-toyama-lz: @TRileySchwarz and I discussed this offline. In this use case, this is fine because we are not using any untrustful third-party OFTs and the LayerZero endpoint is considered to be stable.

yota-toyama-lz: This is fine because of the same reason as https://github.com/LayerZero-Labs/devtools/pull/1584/files#r2244025931.

packages/ovault-evm/contracts/VaultComposerSync.sol
yota-toyama-lz: @TRileySchwarz and I discussed this offline. He suggested that this doesn't seem to be necessary as we already assert the OVault (ERC-4626) is the share token (ERC-20.) But we need to double-check with the unit tests.

yota-toyama-lz (commented):
I looked at the main contract codes. So I'm submitting partial reviews on them. Some of them are already answered by @TRileySchwarz as they are resolved.

Again, I'm a blockchain newbie. So if anything doesn't make any sense, let me know!

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -0,0 +1,333 @@
...
+     * @dev This function can only be called by the contract itself (self-call restriction)
+     *      Decodes the compose message to extract SendParam and minimum message value
+     *      Routes to either deposit or redeem flow based on the input OFT token type
+     * @param _oftIn The OFT token whose funds have been received in the lzReceive associated with this lzTx
+     * @param _composeFrom The bytes32 identifier of the compose sender
+     * @param _composeMsg The encoded message containing SendParam and minMsgValue
+     * @param _amount The amount of tokens received in the lzReceive associated with this lzTx
+     */
+    function handleCompose(
+        address _oftIn,
```
yota-toyama-lz, line 140: Will the intent be more clear when we make this `bool`?

yota-toyama-lz (commented):
I put another small comment.

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: ah so those are constants in the contracts. solidity automagically creates a function for every constant that returns the value in it. `immutables` are constants that are initialized in the constructor.  

https://github.com/LayerZero-Labs/devtools/blob/feat/ovault-evm/packages/ovault-evm/contracts/VaultComposerSync.sol#L35-L43

so if we have a non-constant variable then that would be camel cased

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: the vault composer does a few things, lets assume that the input token is an asset (USDT0) and the output token is a share (yUSDT0). and lets assume that the user wants to convert the asset to a share and get the tokens on the same network that they sent the transaction from - the sending chain is say optimism and the vault chain is arbitrum.

the composer does the following:
0. assume the `composeMsg` is valid on arbitrum and sent from the user on optimism with final recipient as the same user
1. it converts asset to share
2. does and `oft.send(...args)` to send the shares to the user on optimism

now that `oft.send` needs to take in a layerzero fee for message passing. the composer shouldnt pay out of its pocket (it could but that would be a gas abstracted composer), instead the user pays for it. and they do so by paying on source (optimism) via compose value options.

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: yep theyre selectors/signatures. when contracts revert with `errors` or `events` they always have the function signature as the first 4bytes. you can uniquely identify any function (errors and events are special functions) with their signature which iirc is the first 4 bytes of the 32byte keccack hash.

so if you were to run a transaction and you just get hit with a generic message like "revert due to  `0x27b5aea9.....`"  its now easy to lookup the first 4bytes to find the error and then you can do things with `cast` like 

```
$ cast 4byte
4byte         4b   4   -- Get the function signatures for the given selector from https://openchain.xyz                  
4byte-decode  4bd  4d  -- Decode ABI-encoded calldata using https://openchain.xyz                                        
4byte-event   4be  4e  -- Get the event signature for a given topic 0 from https://openchain.xyz 
```  

to decode the message

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: yep you're right it doesnt need to. in an earlier implementation of this i had created a ERC4626Adapter where the ERC20 was not the ERC4626, and it seems that i didnt remove that line when we went back to ERC4626

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: yep its changed in a new PR https://github.com/LayerZero-Labs/devtools/pull/1639/commits/18c45535ec213c80ff81688849dfb539705c2d36

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: not really since it is possible for someone to deploy their own endpoint and have an OFT on that endpoint-mesh while the other OFT is on layerzero-endpoint-mesh

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: you could definitely just re-throw `InsufficientMsgValue` but then in the case where we have more errors where we do not want to try a refund the `if branch` becomes kinda ugly. now you can just go 

`if (a || b || c) {/*same code*/}`

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: maybe but then youre converting an address to bool on the input and then going back from a bool to address which is so its just extra logic which is gonna cost more gas

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: This `SendParam` defines:
1. dstEid
2. receiver
3. minAmountLD
4. composeMsg (if any) 
5. extraOptions (if any) 

`amountLD` gets replaced with the vault out. if the user messes up any of the above then its like sending your tokens to the wrong address and its on them. front end builders are aware of this and should assert against what the user is trying to do 


also we expose `SendParam` on all our send functions and front ends handle that 
```
function send(
        SendParam calldata _sendParam,
        MessagingFee calldata _fee,
        address _refundAddress
    )
```
from `OFTCore.sol`

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: true but since its a evm `call` and not `send` its not critical but it is a UX improvement

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: sorry uh wdym by "caller"?

yota-toyama-lz: It's the `lzCompose` function which is calling this `_refund` function. We already parse the "from" address there.

packages/ovault-evm/contracts/VaultComposerSync.sol
yota-toyama-lz: > going back from a bool to address which is so its just extra logic which is gonna cost more gas

In this function, we are using it only as a boolean, I think. There is no need to convert it back to an address.

shankars99: oh yeah. hmm would have to think about this

examples/ovault-evm/script/SendScript.s.sol
shankars99: good catch

examples/ovault-evm/script/SendScript.s.sol
shankars99: no real reason lol. thanks

examples/ovault-evm/script/SendScript.s.sol
shankars99: thanks

examples/ovault-evm/script/SendScript.s.sol
shankars99: nope and removed

examples/ovault-evm/script/SendScript.s.sol
shankars99: nah it was fudged. cleaned up

examples/ovault-evm/script/SendScript.s.sol
shankars99: for testing failed transactions. shouldnt really be used in prod by anyone unless they want to test a failure state

examples/ovault-evm/script/SendScript.s.sol
shankars99: yep and removed

packages/ovault-evm/.gitignore
shankars99: i commented them out because i didnt want to commit in artifacts until right before package release

examples/ovault-evm/script/ExecutorCompose302.s.sol
shankars99: theyre from a live transaction. ive removed them on the latest and am using an actual tx hash as an example

examples/ovault-evm/script/ExecutorCompose302.s.sol
shankars99: iirc that executor doesnt have `compose302` on it

EWCunha (dismissed):

TRileySchwarz (approved):

TRileySchwarz (approved):

PR #1585, opened 2025-06-20
[INT-295] Feature Move Oapp example move vm decoding

AlexanderLiteplo:
Beefing up the move oapp example with useful parameter encoding sent from evm-send.ts

Adding some utils on move side to assist people in parsing data. I think its useful to have amore built out move example here because most partners come form evm land and have no idea what they're doing in move vm land. The current implementation was built for a partner.

### DIFF
--- a/examples/oapp-aptos-move/contracts/MyOApp.sol
+++ b/examples/oapp-aptos-move/contracts/MyOApp.sol
@@ -7,7 +7,9 @@ import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
 contract MyOApp is OApp, OAppOptionsType3 {
     /// @notice Last string received from any remote chain
-    string public lastMessage;
+    address public address1;
+    address public address2;
+    uint256 public num;
 
     /// @notice The only Message Type in use for this OApp: sending an arbitrary string.
     /// Different message types can be assigned different enforced options per destination endpoint ID.
@@ -109,11 +111,9 @@ contract MyOApp is OApp, OAppOptionsType3 {
         // 1. Decode the incoming bytes into a string
         //    You can use abi.decode, abi.decodePacked, or directly splice bytes
         //    if you know the format of your data structures
-        string memory _message = abi.decode(_message, (string));
+        (address1, address2, num) = abi.decode(_message, (address, address, uint256));
 
         // 2. Apply your custom logic. In this example, store it in `lastMessage`.
-        lastMessage = _message;
-
         // 3. (Optional) Trigger further on-chain actions.
         //    e.g., emit an event, mint tokens, call another contract, etc.
         //    emit MessageReceived(_origin.srcEid, _message);
--- a/examples/oapp-aptos-move/scripts/aptos-move-send.ts
+++ b/examples/oapp-aptos-move/scripts/aptos-move-send.ts
@@ -62,11 +62,16 @@ async function send() {
     const extraOptions = options.toBytes()
 
     // Prepare the message - ABI encode it to match Solidity contract expectations
-    const message = 'Hello, EVM!'
-    console.log('Sending message:', message)
+    const address1 = '0x' // EVM 20 byte address
+    const address2 = '0x' // EVM 20 byte address
+    const num = ethers.BigNumber.from('0')
+    console.log('Sending message:', address1, address2, num)
 
     // ABI encode the string to match what the Solidity contract expects
-    const abiEncodedMessage = ethers.utils.defaultAbiCoder.encode(['string'], [message])
+    const abiEncodedMessage = ethers.utils.defaultAbiCoder.encode(
+        ['address', 'address', 'uint256'],
+        [address1, address2, num]
+    )
     console.log('ABI encoded message:', abiEncodedMessage)
 
     // Convert to bytes array for Aptos
--- a/examples/oapp-aptos-move/scripts/evm-send.ts
+++ b/examples/oapp-aptos-move/scripts/evm-send.ts
@@ -1,6 +1,5 @@
 import { ethers } from 'ethers'
 
-import { makeBytes32 } from '@layerzerolabs/devtools'
 import { EndpointId, getNetworkForChainId } from '@layerzerolabs/lz-definitions'
 import { Options } from '@layerzerolabs/lz-v2-utilities'
 import 'dotenv/config'
@@ -32,14 +31,11 @@ async function main() {
     const aptosMoveEid = EndpointId.APTOS_V2_TESTNET
 
     // Fill in the addresses and number to send to Aptos/Movement
-    const address1 = ''
-    const address2 = ''
-    const num = ethers.BigNumber.from('')
+    const address1 = '0x' // Aptos 32 byte address
+    const address2 = '0x' // Aptos 32 byte address
+    const num = ethers.BigNumber.from('0')
 
-    const encodedMessage = ethers.utils.solidityPack(
-        ['bytes32', 'bytes32', 'uint256'],
-        [makeBytes32(address1), makeBytes32(address2), num]
-    )
+    const encodedMessage = ethers.utils.solidityPack(['bytes32', 'bytes32', 'uint256'], [address1, address2, num])
 
     const hexString = ethers.utils.hexlify(encodedMessage)
 
--- a/examples/oapp-aptos-move/package.json
+++ b/examples/oapp-aptos-move/package.json
@@ -35,7 +35,6 @@
     "@aptos-labs/ts-sdk": "^1.33.1",
     "@babel/core": "^7.23.9",
     "@jest/globals": "^29.7.0",
-    "@layerzerolabs/devtools": "^1.0.0",
     "@layerzerolabs/devtools-extensible-cli": "^0.0.7",
     "@layerzerolabs/devtools-move": "^1.0.9",
     "@layerzerolabs/eslint-config-next": "~2.3.39",
--- a/examples/oapp-aptos-move/scripts/evm-get-received-message.ts
+++ b/examples/oapp-aptos-move/scripts/evm-get-received-message.ts
@@ -1,12 +1,16 @@
 import { ethers } from 'ethers'
 
 /**
- * A utility script to verify cross-chain message delivery by checking the last received message.
- * The lastMessage is updated each time a message is successfully received by the OApp,
+ * A utility script to verify cross-chain message delivery by checking the last received data.
+ * The address1, address2, and num fields are updated each time a message is successfully received by the OApp,
  * providing a simple way to confirm that cross-chain communication is working as expected.
  */
 async function main() {
-    const abi = ['function lastMessage() view returns (string)']
+    const abi = [
+        'function address1() view returns (address)',
+        'function address2() view returns (address)',
+        'function num() view returns (uint256)',
+    ]
 
     const contractAddress = 'your-EVM-OApp-contract-address'
 
@@ -14,8 +18,14 @@ async function main() {
 
     const contract = new ethers.Contract(contractAddress, abi, provider)
 
-    const lastMessage = await contract.lastMessage()
-    console.log('Last received message:', lastMessage)
+    const address1 = await contract.address1()
+    const address2 = await contract.address2()
+    const num = await contract.num()
+
+    console.log('Last received data:')
+    console.log('  Address 1:', address1)
+    console.log('  Address 2:', address2)
+    console.log('  Number:', num.toString())
 }
 
 main()
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -42,7 +42,7 @@ importers:
         version: 17.1.0(eslint-plugin-import@2.29.1)(eslint-plugin-n@16.6.2)(eslint-plugin-promise@6.1.1)(eslint@8.57.1)
       eslint-plugin-import:
         specifier: ^2.29.1
-        version: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint-import-resolver-typescript@3.6.1)(eslint@8.57.1)
+        version: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint@8.57.1)
       eslint-plugin-jest:
         specifier: ^27.6.3
         version: 27.6.3(@typescript-eslint/eslint-plugin@7.7.1)(eslint@8.57.1)(typescript@5.5.3)
@@ -623,9 +623,6 @@ importers:
       '@jest/globals':
         specifier: ^29.7.0
         version: 29.7.0
-      '@layerzerolabs/devtools':
-        specifier: ^1.0.0
-        version: link:../../packages/devtools
       '@layerzerolabs/devtools-extensible-cli':
         specifier: ^0.0.7
         version: link:../../packages/devtools-extensible-cli
@@ -9063,13 +9060,13 @@ packages:
     resolution: {integrity: sha512-WlBSy47LGPILdrNgzPiRtQf/hAY62IN37ncUsQwcr8T7cyX1HZREx2qljuXpvduLDAKn5otsm0XIqHuCRUHEFg==}
     dependencies:
       '@typescript-eslint/eslint-plugin': 7.7.1(@typescript-eslint/parser@7.7.1)(eslint@8.57.1)(typescript@5.5.3)
-      '@typescript-eslint/parser': 7.7.1(eslint@8.57.1)(typescript@5.5.3)
+      '@typescript-eslint/parser': 7.7.1(eslint@8.57.0)(typescript@5.5.3)
       eslint: 8.57.1
       eslint-config-prettier: 9.1.0(eslint@8.57.1)
       eslint-import-resolver-typescript: 3.6.1(@typescript-eslint/parser@7.7.1)(eslint-plugin-import@2.29.1)(eslint@8.57.1)
       eslint-plugin-autofix: 2.2.0(eslint@8.57.1)
       eslint-plugin-compat: 4.2.0(eslint@8.57.1)
-      eslint-plugin-import: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint-import-resolver-typescript@3.6.1)(eslint@8.57.1)
+      eslint-plugin-import: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint@8.57.0)
       eslint-plugin-prettier: 5.1.3(eslint-config-prettier@9.1.0)(eslint@8.57.1)(prettier@3.2.5)
       eslint-plugin-unused-imports: 3.2.0(@typescript-eslint/eslint-plugin@7.7.1)(eslint@8.57.1)
       prettier: 3.2.5
@@ -13972,6 +13969,27 @@ packages:
       - supports-color
     dev: true
 
+  /@typescript-eslint/parser@7.7.1(eslint@8.57.0)(typescript@5.5.3):
+    resolution: {integrity: sha512-vmPzBOOtz48F6JAGVS/kZYk4EkXao6iGrD838sp1w3NQQC0W8ry/q641KU4PrG7AKNAf56NOcR8GOpH8l9FPCw==}
+    engines: {node: ^18.18.0 || >=20.0.0}
+    peerDependencies:
+      eslint: ^8.56.0
+      typescript: '*'
+    peerDependenciesMeta:
+      typescript:
+        optional: true
+    dependencies:
+      '@typescript-eslint/scope-manager': 7.7.1
+      '@typescript-eslint/types': 7.7.1
+      '@typescript-eslint/typescript-estree': 7.7.1(typescript@5.5.3)
+      '@typescript-eslint/visitor-keys': 7.7.1
+      debug: 4.3.7
+      eslint: 8.57.0
+      typescript: 5.5.3
+    transitivePeerDependencies:
+      - supports-color
+    dev: true
+
   /@typescript-eslint/parser@7.7.1(eslint@8.57.1)(typescript@5.5.3):
     resolution: {integrity: sha512-vmPzBOOtz48F6JAGVS/kZYk4EkXao6iGrD838sp1w3NQQC0W8ry/q641KU4PrG7AKNAf56NOcR8GOpH8l9FPCw==}
     engines: {node: ^18.18.0 || >=20.0.0}
@@ -16764,7 +16782,7 @@ packages:
       eslint-plugin-promise: ^6.0.0
     dependencies:
       eslint: 8.57.1
-      eslint-plugin-import: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint-import-resolver-typescript@3.6.1)(eslint@8.57.1)
+      eslint-plugin-import: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint@8.57.1)
       eslint-plugin-n: 16.6.2(eslint@8.57.1)
       eslint-plugin-promise: 6.1.1(eslint@8.57.1)
     dev: true
@@ -16790,7 +16808,7 @@ packages:
       enhanced-resolve: 5.16.0
       eslint: 8.57.1
       eslint-module-utils: 2.8.1(@typescript-eslint/parser@7.7.1)(eslint-import-resolver-typescript@3.6.1)(eslint@8.57.1)
-      eslint-plugin-import: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint-import-resolver-typescript@3.6.1)(eslint@8.57.1)
+      eslint-plugin-import: 2.29.1(@typescript-eslint/parser@7.7.1)(eslint@8.57.0)
       fast-glob: 3.3.2
       get-tsconfig: 4.7.3
       is-core-module: 2.13.1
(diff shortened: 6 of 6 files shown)


examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -2,7 +2,9 @@ import { ethers } from 'ethers'
 
 import { EndpointId } from '@layerzerolabs/lz-definitions'
 import { Options } from '@layerzerolabs/lz-v2-utilities'
+
 import 'dotenv/config'
+import { makeBytes32 } from '../../../packages/devtools/dist'
```
shankars99: umm this definitely needs to be changed to global resolution and not direct path

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
...
+    const aptosEid = EndpointId.APTOS_V2_TESTNET // or EndpointId.MOVEMENT_V2_TESTNET
+
+    // Example addresses and number to encode
+    const address1 = '0x1234567890123456789012345678901234567890'
+    const address2 = '0x9876543210987654321098765432109876543210'
+    const number = ethers.BigNumber.from('123456789012345678901234567890')
+
+    const encodedMessage = ethers.utils.solidityPack(
+        ['bytes32', 'bytes32', 'uint256'],
+        [makeBytes32(address1), makeBytes32(address2), ethers.BigNumber.from(number)]
```
shankars99: `number` is already of type `ethers.BigNumber` from line 38  
```suggestion
        [makeBytes32(address1), makeBytes32(address2), number]
```

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
...
 
-    // Destination endpoint ID for Aptos
-    const aptosEid = EndpointId.APTOS_V2_TESTNET
+    // Destination endpoint ID for Aptos/Movement
+    const aptosEid = EndpointId.APTOS_V2_TESTNET // or EndpointId.MOVEMENT_V2_TESTNET
+
+    // Example addresses and number to encode
+    const address1 = '0x1234567890123456789012345678901234567890'
+    const address2 = '0x9876543210987654321098765432109876543210'
+    const number = ethers.BigNumber.from('123456789012345678901234567890')
```
shankars99: since `number` is a typescript type too. maybe call this `num`? (assertion to make sure the user doesnt pass in something like "1.2345" would be good too cuz you take an input as as `string` and then convert to `number`)

```suggestion
    const num = ethers.BigNumber.from('123456789012345678901234567890')
```

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
...
+    )
+
+    const hexString = ethers.utils.hexlify(encodedMessage)
 
-    // Message to send
-    const message = 'Hello Aptos!'
+    console.log('Encoded message:', hexString)
+    console.log('Address1:', address1)
+    console.log('Address2:', address2)
+    console.log('Number:', number.toString())
```
shankars99: You don't need to convert `number` to `number.toString()` since `console.log()` automagically does it for you 

```suggestion
    console.log('Number:', number)
```

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
...
-        .toHex()
-        .toString()
+    const options = Options.newOptions().addExecutorLzReceiveOption(200000, 0).toHex().toString()
 
     try {
         // Get quote
-        const [nativeFee] = await myOApp.quote(aptosEid, message, options, false)
-        console.log(`Quote for message: ${nativeFee} ETH`)
+        const [nativeFee] = await myOApp.quote(aptosEid, hexString, options, false)
+        console.log(`Quote for message: ${ethers.utils.formatEther(nativeFee)} ETH`)
```
shankars99: the native token is not _guaranteed_ to be `ETH` - BSC, Polygon, etc have their own native tokens

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
...
 
         // Send message
-        const tx = await myOApp.send(aptosEid, message, options, {
+        const tx = await myOApp.send(aptosEid, hexString, options, {
             value: nativeFee.toString(),
         })
 
-        console.log('Sending message to Aptos...')
+        console.log('Sending encoded message to Aptos/Movement...')
```
shankars99: 1. Sending encoded message to Aptos/Movement makes it sound like the transaction is being sent to aptos or movement  
2. can use a chainId or eid mapping to find if the network is `aptos` or `movement`

```suggestion
        console.log('Sending encoded message from <network> to eidToNetwork[moveEid]...')
```

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
     // Create contract instance
     const myOApp = new ethers.Contract(contractAddress, ABI, wallet)
 
-    // Destination endpoint ID for Aptos
-    const aptosEid = EndpointId.APTOS_V2_TESTNET
+    // Destination endpoint ID for Aptos/Movement
+    const aptosEid = EndpointId.APTOS_V2_TESTNET // or EndpointId.MOVEMENT_V2_TESTNET
```
shankars99: naming this `moveEid` so that its generic across the aptos-move networks
```suggestion
    const moveEid = EndpointId.APTOS_V2_TESTNET // or EndpointId.MOVEMENT_V2_TESTNET
```

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -27,29 +29,40 @@ async function main() {
...
         // Send message
-        const tx = await myOApp.send(aptosEid, message, options, {
+        const tx = await myOApp.send(aptosEid, hexString, options, {
             value: nativeFee.toString(),
         })
 
-        console.log('Sending message to Aptos...')
+        console.log('Sending encoded message to Aptos/Movement...')
         const receipt = await tx.wait()
         console.log(`Transaction hash: ${receipt?.transactionHash}`)
```
shankars99: would be nice to also print out a lzScan link here

examples/oapp-aptos-move/scripts/evm-send.ts
shankars99, line 1: why not accept arguments for `eid` and `number`? the addresses can stay the same since they're less likely to be changed but the `eid` and `number`can vary and its easier to change (and notice that you changed) the command than update a file

examples/oapp-aptos-move/sources/utils.move
```diff
@@ -0,0 +1,89 @@
...
+
+    const ASCII_ZERO: u8 = 48;
+    const ASCII_NINE: u8 = 57;
+    const ASCII_UPPERCASE_A: u8 = 65;
+    const ASCII_UPPERCASE_F: u8 = 70;
+    const ASCII_LOWERCASE_A: u8 = 97;
+    const ASCII_LOWERCASE_F: u8 = 102;
+    
+    const EINVALID_HEX_CHARACTER: u64 = 1;
+    const EINVALID_ADDRESS_LENGTH: u64 = 2;
```
shankars99: constants at top of the file is convention, except for `error_codes` since its less likely to be referenced

shankars99:
also `pnpm changeset` pls

examples/oapp-aptos-move/scripts/evm-send.ts
AlexanderLiteplo: good catch thanks

examples/oapp-aptos-move/scripts/evm-send.ts
AlexanderLiteplo: good catch thanks

examples/oapp-aptos-move/scripts/evm-send.ts
AlexanderLiteplo: It's supposed to just be script that people can play around with and not a robust cli command. I just want people to be able to get started with the example and quickly try out sending some messages. In this case addresses are just as likely to change as numbers because someone could be using it to log transactions.

AlexanderLiteplo: If we make it a full on script with commands I think that's a bit overkill. I'd rather have this be a sandbox for people to get started and have an idea of how to use OApps. However this is different from what we have on other examples so maybe I should simplify it and make it just a string.

My Idea is that most people are new to Move VM and don't know how to get started on it.

shankars99: if its supposed to be like an example or sandbox script then id prolly comment out the magic numbers and instead leave them in as empty strings so that people know what to enter into the variable but dont have something pre-added that they just run (they shouldnt but theres always a chance that someone would on accident as they assume the script works differently)

AlexanderLiteplo: k fixed

examples/oapp-aptos-move/contracts/MyOApp.sol
```diff
@@ -1,34 +1,28 @@
-// SPDX-License-Identifier: MIT
-
+// SPDX-License-Identifier: UNLICENSED
```
abaltes-lz: has this been by legal? From what I understand Unlicensed basically means that this provides no permission for outside entities to use this code except by specific agreement

examples/oapp-aptos-move/contracts/MyOApp.sol
```diff
@@ -1,34 +1,28 @@
...
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
 contract MyOApp is OApp, OAppOptionsType3 {
-    constructor(address _endpoint, address _delegate) OApp(_endpoint, _delegate) Ownable(_delegate) {}
+    /// @notice Last string received from any remote chain
+    string public lastMessage;
 
-    string public data = "Nothing received yet.";
-    uint256 public counter = 0;
+    /// @notice Msg type for sending a string, for use in OAppOptionsType3 as an enforced option
```
abaltes-lz: I'd emphasize that there is only one message type for this OApp, and that the string is arbitrary (technically all messages send a string). You could say something like, "The only Message Type in use for this OApp: sending an arbitrary string. Different message types can be assigned different enforced options per dst eid"

examples/oapp-aptos-move/contracts/MyOApp.sol
```diff
@@ -38,35 +32,89 @@ contract MyOApp is OApp, OAppOptionsType3 {
...
+    //   • Only the LayerZero Endpoint can call this method
+    //   • The sender is a registered peer (peers[srcEid] == origin.sender)
+    // ──────────────────────────────────────────────────────────────────────────────
+
+    /// @notice Invoked by OAppReceiver when EndpointV2.lzReceive is called
+    /// @dev   _origin    Metadata (source chain, sender address, nonce)
+    /// @dev   _guid      Global unique ID for tracking this message
+    /// @param _message   ABI-encoded bytes (the string we sent earlier)
+    /// @dev   _executor  Executor address that delivered the message
+    /// @dev   _extraData Additional data from the Executor (unused here)
```
abaltes-lz:
```suggestion
    /// @dev   _extraData Additional data from the Executor (unused by the LayerZero executor)
```

examples/oapp-aptos-move/scripts/aptos-move-send.ts
```diff
@@ -29,24 +30,25 @@ import { Options } from '@layerzerolabs/lz-v2-utilities'
 dotenv.config()
 
 // Validate required environment variables
-if (!process.env.APTOS_PRIVATE_KEY || !process.env.ACCOUNT_ADDRESS) {
-    throw new Error('Please set APTOS_PRIVATE_KEY and ACCOUNT_ADDRESS in your .env file')
+if (!process.env.APTOS_PRIVATE_KEY || !process.env.APTOS_ACCOUNT_ADDRESS) {
```
abaltes-lz: this can go after the assignments and switched to 
if (!APTOS_PRIVATE_KEY || !APTOS_ACCOUNT_ADDRESS)

examples/oapp-aptos-move/scripts/aptos-move-send.ts
```diff
@@ -29,24 +30,25 @@ import { Options } from '@layerzerolabs/lz-v2-utilities'
...
 
 // Configuration constants
 const APTOS_PRIVATE_KEY = process.env.APTOS_PRIVATE_KEY
-const ACCOUNT_ADDRESS = process.env.ACCOUNT_ADDRESS
+const APTOS_ACCOUNT_ADDRESS = process.env.APTOS_ACCOUNT_ADDRESS
 
 // OApp configuration
-const OAPP_ADDRESS = '' // Set your OApp's address on Aptos
-const REMOTE_EID = EndpointId.BSC_V2_TESTNET // Destination chain endpoint ID
+const OAPP_ADDRESS = '<your-Aptos-Move-OApp-address>'
```
abaltes-lz, line 42: any reason not to use environment variable or config file for this? might be worth having a config.ts that just sets the oapp address, network, eids for the aptos and evm sends

examples/oapp-aptos-move/scripts/evm-send.ts
```diff
@@ -21,37 +22,51 @@ async function main() {
...
+    const aptosMoveEid = EndpointId.APTOS_V2_TESTNET
 
-    // Message to send
-    const message = 'Hello Aptos!'
+    // Fill in the addresses and number to send to Aptos/Movement
+    const address1 = ''
+    const address2 = ''
+    const num = ethers.BigNumber.from('')
+
+    const encodedMessage = ethers.utils.solidityPack(
```
abaltes-lz: I don't think it's ideal to encode a message offchain that will be decoded on-chain

examples/oapp-aptos-move/sources/oapp.move
```diff
@@ -24,13 +32,63 @@ module oapp::oapp {
...
+            counter: 0,
+            raw_message: vector::empty()
+        });
     }
 
-    #[view]
-    public fun get_counter(): u64 acquires Counter {
-        borrow_global<Counter>(@oapp).value
+    public fun parse_message(message: vector<u8>): (address, address, u256) {
+        let string_length = (
```
abaltes-lz: I would make a comment to explain the message format - as a follow-up to other comment, if decoding on the receive side, the send side should be also be encoding this message.

examples/oapp-aptos-move/sources/oapp.move
```diff
@@ -24,13 +32,63 @@ module oapp::oapp {
...
+            (*vector::borrow(&message, 60) as u64) << 24 |
```
abaltes-lz: consider using endpoint_common::serde for deserialization since this is already a transient dependency for oapp

examples/oapp-aptos-move/sources/utils.move
```diff
@@ -0,0 +1,89 @@
...
+    const ASCII_ZERO: u8 = 48;
+    const ASCII_NINE: u8 = 57;
+    const ASCII_UPPERCASE_A: u8 = 65;
+    const ASCII_UPPERCASE_F: u8 = 70;
+    const ASCII_LOWERCASE_A: u8 = 97;
+    const ASCII_LOWERCASE_F: u8 = 102;
+
+    /// Converts a vector of bytes to a UTF-8 string
+    /// Will abort if the bytes are not valid UTF-8
+    public fun bytes_to_string(bytes: vector<u8>): String {
```
abaltes-lz: just use the string::utf8 and string::bytes in place where needed

examples/oapp-aptos-move/sources/utils.move
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// Converts a hex string (without 0x prefix) to a vector of bytes
+    /// Example: "48656c6c6f" -> b"Hello"
+    /// Automatically pads odd-length strings with a leading zero
+    public fun hex_string_to_bytes(hex_str: String): vector<u8> {
+        let hex_bytes = string::bytes(&hex_str);
+        let len = vector::length(hex_bytes);
+        
+        // If the hex string is odd length, pad with leading zero
```
abaltes-lz: this should probably just assert and fail if the vector is the wrong length, unless there is a legitimate reason for it to be an odd length

examples/oapp-aptos-move/sources/utils.move
```diff
@@ -0,0 +1,89 @@
...
+            let padded = vector::empty<u8>();
+            vector::push_back(&mut padded, ASCII_ZERO);
+            vector::append(&mut padded, *hex_bytes);
+            padded
+        } else {
+            *hex_bytes
+        };
+        
+        let padded_len = vector::length(&padded_hex);
+        let result = vector::empty<u8>();
```
abaltes-lz: in the core protocol, we've done vector[] instead of vector::empty() as a common practice. Also, no need to put the <u8> which is inferred

examples/oapp-aptos-move/sources/utils.move
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Converts a UTF-8 string to a vector of bytes
+    public fun string_to_bytes(str: String): vector<u8> {
+        *string::bytes(&str)
+    }
+
+    /// Converts a hex string (without 0x prefix) to a vector of bytes
+    /// Example: "48656c6c6f" -> b"Hello"
+    /// Automatically pads odd-length strings with a leading zero
+    public fun hex_string_to_bytes(hex_str: String): vector<u8> {
```
abaltes-lz, line 15: generally, we wouldn't be sending a hex string over the wire. is this necessary?

examples/oapp-aptos-move/sources/utils.move
```diff
@@ -0,0 +1,89 @@
...
+            char - ASCII_ZERO
+        } else if (char >= ASCII_UPPERCASE_A && char <= ASCII_UPPERCASE_F) {
+            char - ASCII_UPPERCASE_A + 10
+        } else if (char >= ASCII_LOWERCASE_A && char <= ASCII_LOWERCASE_F) {
+            char - ASCII_LOWERCASE_A + 10
+        } else {
+            abort EINVALID_HEX_CHARACTER
+        }
+    }
+
```
abaltes-lz, line 65: should have an ERROR CODES header

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: Great question. I copied that directly from the EVM example assuming it is correct. I will ask internally to clarify.

examples/oapp-aptos-move/scripts/aptos-move-send.ts
AlexanderLiteplo: Its really just a personal preference thing. We don't usually offer this type of script in our devtools. I've just added it because I think it would be useful to have some scripts where people can play around and get an idea of how to call our contracts.

I think a config.ts file is overkill as this isn't supposed to be someone's workhorse. Just a simple hello world style script to get people started.

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: @krak can you weigh in on this as you wrote the comments

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: switched back

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: I'll just switch all occurrences of this and then tag krak in Slack

examples/oapp-aptos-move/sources/utils.move
AlexanderLiteplo: in MyOApp.sol we do this:

```

    // ──────────────────────────────────────────────────────────────────────────────
    // 1. Send business logic
    //
    // Example: send a simple string to a remote chain. Replace this with your
    // own state-update logic, then encode whatever data your application needs.
    // ──────────────────────────────────────────────────────────────────────────────

    /// @notice Send a string to a remote OApp on another chain
    /// @param _dstEid   Destination Endpoint ID (uint32)
    /// @param _message  The string to send
    /// @param _options  Execution options for gas on the destination (bytes)
    function send(uint32 _dstEid, string calldata _message, bytes calldata _options) external payable {
        // 1. (Optional) Update any local state here.
        //    e.g., record that a message was "sent":
        //    sentCount += 1;

        // 2. Encode any data structures you wish to send into bytes
        //    You can use abi.encode, abi.encodePacked, or directly splice bytes
        //    if you know the format of your data structures
        bytes memory _message = abi.encode(_message);

        // 3. Call OAppSender._lzSend to package and dispatch the cross-chain message
        //    - _dstEid:   remote chain's Endpoint ID
        //    - _message:  ABI-encoded string
        //    - _options:  combined execution options (enforced + caller-provided)
        //    - MessagingFee(msg.value, 0): pay all gas as native token; no ZRO
        //    - payable(msg.sender): refund excess gas to caller
        //
        //    combineOptions (from OAppOptionsType3) merges enforced options set by the contract owner
        //    with any additional execution options provided by the caller
        _lzSend(
            _dstEid,
            _message,
            combineOptions(_dstEid, SEND, _options),
            MessagingFee(msg.value, 0),
            payable(msg.sender)
        );
    }
    ```

Comment (AlexanderLiteplo) [reply examples/oapp-aptos-move/sources/utils.move:15]: I'm building this entire example around the existing OApp.sol so it will integrate into existing examples

Comment (AlexanderLiteplo) [reply examples/oapp-aptos-move/sources/utils.move:15]: @St0rmBr3w what do you think?

File: examples/oapp-aptos-move/sources/utils.move (hunk 19 above)
Comment (AlexanderLiteplo) [reply examples/oapp-aptos-move/sources/utils.move]: fixed

File: examples/oapp-aptos-move/scripts/evm-send.ts (hunk 15 above)
Comment (AlexanderLiteplo) [reply examples/oapp-aptos-move/scripts/evm-send.ts]: That is common practice in our examples. Also if you encode a message on one chain then decode it on a totally different VM in my opinion its no different from offchain encoding to onchain decoding.

@St0rmBr3w can you comment on this?

File: examples/oapp-aptos-move/sources/oapp.move (hunk 16 above)
Comment (AlexanderLiteplo) [reply examples/oapp-aptos-move/sources/oapp.move]: comment added, discussion continued on other message

File: examples/oapp-aptos-move/contracts/MyOApp.sol [hunk 23]
```diff
@@ -1,34 +1,31 @@
...
-import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
-import { OApp, MessagingFee, Origin } from "@layerzerolabs/oapp-evm/contracts/oapp/OApp.sol";
-import { MessagingReceipt } from "@layerzerolabs/oapp-evm/contracts/oapp/OAppSender.sol";
+import { OApp, Origin, MessagingFee } from "@layerzerolabs/oapp-evm/contracts/oapp/OApp.sol";
 import { OAppOptionsType3 } from "@layerzerolabs/oapp-evm/contracts/oapp/libs/OAppOptionsType3.sol";
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 
 contract MyOApp is OApp, OAppOptionsType3 {
-    constructor(address _endpoint, address _delegate) OApp(_endpoint, _delegate) Ownable(_delegate) {}
+    /// @notice Last string received from any remote chain
```
DanL0, line 9: this comment seems inaccurate

it mentions "last string received"

while below there are three variables: address, address, uint256 and none of them is string

maybe better to change comment to "last data received"?

examples/oapp-aptos-move/contracts/MyOApp.sol
```diff
@@ -38,35 +35,87 @@ contract MyOApp is OApp, OAppOptionsType3 {
...
+        //    - MessagingFee(msg.value, 0): pay all gas as native token; no ZRO
+        //    - payable(msg.sender): refund excess gas to caller
+        //
+        //    combineOptions (from OAppOptionsType3) merges enforced options set by the contract owner
+        //    with any additional execution options provided by the caller
+        _lzSend(
+            _dstEid,
+            _message,
+            combineOptions(_dstEid, SEND, _options),
+            MessagingFee(msg.value, 0),
```
DanL0, line 84: I think this is an anti pattern to force users to use native gas for paying fee and hardcoding it in the code.

This makes it impossible in future to pay with ZRO token, I don't think we want to encourage users to follow this pattern and disable it in their OApps they are building on top of this example.

Instead, send method should accept MessagingFee struct, like OFTCore: https://github.com/LayerZero-Labs/devtools/blob/main/packages/oft-evm/contracts/OFTCore.sol#L177

examples/oapp-aptos-move/contracts/MyOApp.sol
```diff
@@ -38,35 +35,87 @@ contract MyOApp is OApp, OAppOptionsType3 {
...
-        bytes calldata /*payload*/,
+        bytes calldata _message,
         address /*_executor*/,
         bytes calldata /*_extraData*/
     ) internal override {
-        counter += 1;
+        // 1. Decode the incoming bytes into a string
+        //    You can use abi.decode, abi.decodePacked, or directly splice bytes
+        //    if you know the format of your data structures
+        (address1, address2, num) = abi.decode(_message, (address, address, uint256));
```
DanL0, line 114: does it work if you are currently encoding a "string" on send:
```
bytes memory _message = abi.encode(_message);
```

and on receive you are decoding "address, address, uint"?:
```
(address1, address2, num) = abi.decode(_message, (address, address, uint256));
```

I did a Foundry small test and I think if you currently do abi.encode offchain, pass it onchain and then do abi.encode again in Solidity before sending it results in wrong data:
```
function test_msg_passing() public virtual {
        bytes memory encoded = abi.encode(string(abi.encode(address(1), address(2), 100)));
        console.logBytes(encoded);

        (address address1, address address2, uint256 num) = abi.decode(encoded, (address, address, uint256));
        console.log(address1);
        console.log(address2);
        console.log(num);
    }
```

result:
```
Logs:
  0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000064
  0x0000000000000000000000000000000000000020
  0x0000000000000000000000000000000000000060
  1
```

Expected:
```
0x000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000064
  0x0000000000000000000000000000000000000001
  0x0000000000000000000000000000000000000002
  100
```

Could you create a test case in Foundry for sending messages from EVM to EVM using this OApp Solidity code and double check if the encoding and decoding logic is consistent and works?

examples/oapp-aptos-move/sources/utils.move
St0rmBr3w: It may be worth just using a simpler data structure in the long-run, the string passing OApp is abi.encoded on EVM, which results in a complex data structure on non-EVM chains to unpack.

If you `abi.encode("Hello World")`, you will get the following byte structure:

```solidity
// The function above will return the following raw bytes value.
0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000b48656c6c6f20576f726c64000000000000000000000000000000000000000000

// We can split this into words that are 32 bytes long to get:
0x0000000000000000000000000000000000000000000000000000000000000020  // offset
  000000000000000000000000000000000000000000000000000000000000000b   // length 
  48656c6c6f20576f726c64000000000000000000000000000000000000000000     // string
```
  
While "Hello World" makes sense for EVM chains, given the general complexity of implementing decoding for this message structure, I think it'd be better to have it just pass a uint value of some kind, and set a variable.

St0rmBr3w: Nazreen got around this by changing the encoding format from the standard `abi.encode`, but I think this is generally more distracting from the point of the example: to show users how to simply encode a data structure and decode it on a non-EVM chain.

https://github.com/LayerZero-Labs/devtools/blob/8641261d98d23bca325d19b93e8aed566d37718e/examples/oapp-solana/contracts/MyOApp.sol#L31

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: This is taken directly from the EVM OApp example (that devrel wrote) I'm just updating these files with their implementation. If you want me to change it across all examples I can do that, but it should be a separate PR.

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: This is taken directly from the EVM OApp example (that devrel wrote) I'm just updating these files with their implementation. If you want me to change it across all examples I can do that, but it should be a separate PR.

examples/oapp-aptos-move/contracts/MyOApp.sol
AlexanderLiteplo: Sounds good. Putting this on pause until sla-tracker work is finished.

shankars99:

PR #1591, opened 2025-06-25
INT-402: NativeOFTAdapterUpgradeable

ravinagill15:
## In this PR:
- Implemented NativeOFTAdapterUpgradeable, essentially the same as [NativeOFTAdapter](https://github.com/LayerZero-Labs/devtools/blob/main/packages/oft-evm/contracts/NativeOFTAdapter.sol), but using `OFTCoreUpgradeable`
- Added tests and mocks as needed

### DIFF
--- a/packages/oft-evm-upgradeable/contracts/oft/NativeOFTAdapterUpgradeable.sol
+++ b/packages/oft-evm-upgradeable/contracts/oft/NativeOFTAdapterUpgradeable.sol
@@ -1,0 +1,195 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.22;
+
+import { MessagingFee, MessagingReceipt, OFTFeeDetail, OFTLimit, OFTReceipt, SendParam } from "@layerzerolabs/oft-evm/contracts/OFTCore.sol";
+import { IOFT, OFTCoreUpgradeable } from "./OFTCoreUpgradeable.sol";
+
+/**
+ *
+ * @title NativeOFTAdapterUpgradeable
+ * @dev NativeOFTAdapterUpgradeable is a contract that adapts native currency to the OFT functionality.
+ *
+ * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
+ * unless you make a NON-default implementation of OFT, which needs to be done very carefully.
+ * @dev WARNING: The default NativeOFTAdapterUpgradeable implementation assumes LOSSLESS transfers, ie. 1 native in, 1 native out.
+ */
+abstract contract NativeOFTAdapterUpgradeable is OFTCoreUpgradeable {
+
+    error IncorrectMessageValue(uint256 provided, uint256 required);
+    error CreditFailed(address to, uint256 amountLD, bytes revertData);
+
+    /**
+     * @param _localDecimals The decimals of the native on the local chain (this chain). 18 on ETH.
+     * @param _lzEndpoint The LayerZero endpoint address.
+     * @param _delegate The delegate capable of making OApp configurations inside of the endpoint.
+     */
+    constructor(
+        uint8 _localDecimals,
+        address _lzEndpoint,
+        address _delegate
+    ) OFTCoreUpgradeable(_localDecimals, _lzEndpoint) {}
+
+    /**
+     * @dev Initializes the NativeOFTAdapter with the provided delegate.
+     * @param _delegate The delegate capable of making OApp configurations inside of the endpoint.
+     *
+     * @dev The delegate typically should be set as the owner of the contract.
+     * @dev Ownable is not initialized here on purpose. It should be initialized in the child contract to
+     * accommodate the different version of Ownable.
+     */
+    function __NativeOFTAdapter_init(address _delegate) internal onlyInitializing {
+        __OFTCore_init(_delegate);
+    }
+
+    function __NativeOFTAdapter_init_unchained() internal onlyInitializing {}
+
+    /**
+     * @dev Returns the address of the native token
+     * @return The address of the native token.
+     */
+    function token() public pure returns (address) {
+        return address(0);
+    }
+
+    /**
+     * @notice Indicates whether the OFT contract requires approval of the 'token()' to send.
+     * @return bool indicating whether approval of underlying token implementation is required.
+     *
+     * @dev In the case of default NativeOFTAdapter, approval is not required.
+     */
+    function approvalRequired() external pure virtual returns (bool) {
+        return false;
+    }
+
+    /**
+     * @dev Executes the send operation while ensuring the correct amount of native is sent.
+     * @param _sendParam The parameters for the send operation.
+     * @param _fee The calculated fee for the send() operation.
+     *      - nativeFee: The native fee.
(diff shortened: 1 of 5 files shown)


packages/oft-evm-upgradeable/test/OFTFeeUpgradeable.t.sol
```diff
@@ -356,4 +356,44 @@ contract OFTFeeTest is OFTTest {
...
+        assertEq(IERC20(cOFTAdapter.token()).balanceOf(userC), initialBalance);
+
+        assertEq(aOFT.token(), address(aOFT));
+        assertEq(bOFT.token(), address(bOFT));
+        assertEq(cOFTAdapter.token(), address(cERC20Mock));
+
+        // Skip native adapter assertions in fee test suite
+    }
+
+    function test_native_oft_adapter_debit() public virtual override {
```
DanL0, line 380: I'm not a fan of this solution for testing. I think it leaks problems too much to the test suite that derives from it.

It seems each time we will use OFT.t.sol as a base we will have to manually skip native OFT adapter tests? 

I think we should instead create a new test file specifically for NativeOFTAdapter or NativeOFTAdapterUpgradeable so we don't have to call vm.skip(true) in any of test suites that derives from OFT.t.sol.

New test file can derive from OFT.t.sol.

packages/oft-evm-upgradeable/test/mocks/NativeOFTAdapterUpgradeableMock.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.22;
+
+import { NativeOFTAdapterUpgradeable } from "../../contracts/oft/NativeOFTAdapterUpgradeable.sol";
+
+contract NativeOFTAdapterUpgradeableMock is NativeOFTAdapterUpgradeable {
+    constructor(
+        uint8 _localDecimals,
+        address _lzEndpoint
+    ) NativeOFTAdapterUpgradeable(_localDecimals, _lzEndpoint, address(0)) {}
```
DanL0, line 10: disableInitializers in constructor of the mock

packages/oft-evm-upgradeable/test/OFT.t.sol
```diff
@@ -76,11 +85,20 @@ contract OFTTest is TestHelperOz5 {
             )
         );
 
+        dNativeOFTAdapter = NativeOFTAdapterUpgradeableMock(
```
DanL0, line 88: Please use separate file for native OFT adapter tests or something that doesn't have to be vm.skipped if something derives from the test suite

packages/oft-evm-upgradeable/test/lib/NativeOFTAdapterUpgradeableMockCodec.sol
```diff
@@ -0,0 +1,15 @@
...
+pragma solidity ^0.8.0;
+
+import { NativeOFTAdapterUpgradeable } from "../../contracts/oft/NativeOFTAdapterUpgradeable.sol";
+import { NativeOFTAdapterUpgradeableMock } from "../mocks/NativeOFTAdapterUpgradeableMock.sol";
+
+// @title NativeOFTAdapterUpgradeableMockCodec
+// @notice Codec to convert NativeOFTAdapterUpgradeable to NativeOFTAdapterUpgradeableMock in a consistent, readable manner.
+// @dev For testing purposes only.
+library NativeOFTAdapterUpgradeableMockCodec {
+    function asNativeOFTAdapterUpgradeableMock(NativeOFTAdapterUpgradeable _oft) internal pure returns (NativeOFTAdapterUpgradeableMock) {
```
DanL0, line 12: is this used anywhere?

packages/oft-evm-upgradeable/test/OFTFeeUpgradeable.t.sol
shankars99: what we need is a base test framework for ugradeable and then there is a new test file that tests ONLY that OApp.

reference: https://github.com/LayerZero-Labs/devtools/tree/feat/ovault-evm/packages/ovault-evm/test/composer

which creates a base, and then you implement Unit, E2E, Upgradeable Unit, etc

packages/oft-evm-upgradeable/test/OFT.t.sol
```diff
@@ -120,9 +139,12 @@ contract OFTTest is TestHelperOz5 {
         assertEq(aOFT.token(), address(aOFT));
         assertEq(bOFT.token(), address(bOFT));
         assertEq(cOFTAdapter.token(), address(cERC20Mock));
+        assertEq(dNativeOFTAdapter.token(), address(0));
+
+        assertEq(dNativeOFTAdapter.approvalRequired(), false);
     }
 
-    function test_oftVersion() public view {
```
shankars99, line 125: why is this change required?

packages/oft-evm-upgradeable/contracts/oft/NativeOFTAdapterUpgradeable.sol
```diff
@@ -0,0 +1,195 @@
...
+        SendParam calldata _sendParam,
+        MessagingFee calldata _fee,
+        address _refundAddress
+    ) public payable virtual override returns (MessagingReceipt memory msgReceipt, OFTReceipt memory oftReceipt) {
+        // @dev Ensure the native funds in msg.value are exactly enough to cover the fees and amount to send (with dust removed).
+        // @dev This will revert if the _sendParam.amountLD contains any dust
+        uint256 requiredMsgValue = _fee.nativeFee + _removeDust(_sendParam.amountLD);
+        if (msg.value != requiredMsgValue) {
+            revert IncorrectMessageValue(msg.value, requiredMsgValue);
+        }
```
shankars99, line 90: this expects that the user uses a `sendParam.amountLD` that would result in no dust. why cant the user send extra native and it get refunded to them later at the refund address?

shankars99 (changes requested):
not a fan of the `send()`

packages/oft-evm-upgradeable/test/OFTFeeUpgradeable.t.sol
DanL0: Looks like a clean separation @shankars99 👍

packages/oft-evm-upgradeable/contracts/oft/NativeOFTAdapterUpgradeable.sol
DanL0: this is how NativeOFTAdapter (non-upgradeable) operates already, I don't think we should change this in upgradeable version. we should change in both if we want to change it

in my legacy implementation I have also allowed refunds, here you need to pass very precise amounts which I think can be a little bit worse UX during gas spikes: revert instead of refund

shankars99: yeah we should change it in both places. revert if not exact is worse UX

ravinagill15:
Closing this PR as Plume has stated it is not required to have upgradeable contracts

PR #1592, opened 2025-06-26
fix bug in cli version check

Tlazypanda:
Fixes this bug which wrongly calculates the CLI version for move-packages. 

Acc. to prev logic v7.5.0 was older than v6.0.1
image


DanL0:
Hello, this is good find!

I have opened new PR: https://github.com/LayerZero-Labs/devtools/pull/1593 with the fix + tests + changesets and we will continue tracking this in the new PR.

PR #1593, opened 2025-06-26
INT-406: fix move CLI version checks

DanL0:
Based on: https://github.com/LayerZero-Labs/devtools/pull/1592

The version check for the CLI was wrong, as flagged by the original PR creator it was showing that 7.5.0 isn't greater or equal than 6.0.2.

This PR fixes the problem and also adds tests and enables running them in CI.

### DIFF
--- a/.changeset/wet-colts-peel.md
+++ b/.changeset/wet-colts-peel.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/devtools-move": patch
+---
+
+fix move CLI version checks
--- a/packages/devtools-move/jest/config.test.ts
+++ b/packages/devtools-move/jest/config.test.ts
@@ -1,0 +1,36 @@
+import { expect } from 'chai'
+import { isVersionGreaterOrEqualTo, isVersionLessThanOrEqualTo } from '../tasks/move/utils/config'
+
+describe('tasks/move/utils/config', () => {
+    describe('isVersionGreaterOrEqualTo', () => {
+        it('works', () => {
+            expect(isVersionGreaterOrEqualTo('6.0.1', '6.0.1')).to.equal(true)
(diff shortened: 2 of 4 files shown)


shankars99:
@AlexanderLiteplo and I had a discussion about this. 

So the entire stack has been tested at `6.0.1` - oft, oapp, endpoint, libraries, etc. So we don't know if using `v7` would break anything or even if a compile and test work the bytecode is different and it doesnt play well with our systems. 

As such we think instead of just allowing anyone over the version of `6.0.1` to build, we guarantee the working at `6.0.1` and anything over that we info prompt them something along the lines of "something might be different and we cant guarantee equivalence or provide support if a re-write is required to be compliant with a higher version"

DanL0:
Got it. I'm assuming we can close this PR then and we can open new PR for adding warning @shankars99 ?

AlexanderLiteplo:
Good catch Dan, I think we should merge this as that is indeed a bug.

AlexanderLiteplo:
We can add a warning in a new PR or just add a warning to this one.

AlexanderLiteplo:
I recently tested with version over 7 and everything worked

AlexanderLiteplo (dismissed):
THis should be merged, and then we can put a warning within the CLI later. For now the Readme reccomends 6.0.1 which is enough of a warning. But currently there is a blocked partner so lets get this merged asap.

AlexanderLiteplo:
mvm going to just quickly add a warning when its not 6.0.1

AlexanderLiteplo:
image

Added warning

cc: @shankars99 @DanL0

AlexanderLiteplo:
added operation cancel option
image

shankars99 (approved):

AlexanderLiteplo (approved):

PR #1595, opened 2025-06-26
INT-402: mint/burn native oft adapter upgradeable

ravinagill15:
## In this PR:

- Implementing mint/burn native oft adapter upgradeable in devtools instead of orbit-adapter repo to avoid open zeppelin versioning issues

### Todo
- tests


ravinagill15:
@St0rmBr3w Yeah that's a great point since this is a ready to use contract. I marked it as abstract to follow the pattern of the other existing upgradeable implementations we have like [OFTAdapterUpgradeable](https://github.com/LayerZero-Labs/devtools/blob/main/packages/oft-evm-upgradeable/contracts/oft/OFTAdapterUpgradeable.sol)

Regardless, this PR will be closed because Plume has stated upgradeable contracts are not a hard requirement for them. Furthermore, since no liquidity will be locked inside a mint/burn contract, a new mint/burn contract can easily be deployed and wired into the mesh

PR #1596, opened 2025-06-30
docs: example READMEs specs + examples/oft README revamp

nazreen:
## This PR

- introduces a new specs document for Examples, which will initially contain the structure for READMEs
- revamps `examples/oft/README/md` based on the new specs

### DIFF
--- a/docs/EXAMPLES_DEVELOPMENT.md
+++ b/docs/EXAMPLES_DEVELOPMENT.md
@@ -1,0 +1,90 @@
+This document is intended for the maintainers of the examples that are in `/examples` in this repo. It is also meant as a guide for coding agents for the purposes of reviewing or editing.
+
+Currently, this document will only detail the structure for the READMEs of the examples.
+
+## Structure for examples' READMEs
+
+1. **Header**
+   - Goal: Branding + promote docs site + entrypoint
+   - Contents: LayerZero logo + links to docs and dev site
+
+2. **Example Title**
+   - Goal: What the example will teach
+   - Contents: Title and 1–2 sentence description (possibly goal-oriented)
+
+3. **Prerequisite Knowledge**
+   - Goal: What to understand before running the example
+   - Contents: Short list (≤3 items) like OApp, OFT
+
+4. **Requirements**
+   - Goal: What needs to be installed
+   - Contents: Tools + exact versions
+
+5. **Scaffold this example**
+   - Goal: How to init the example
+   - Contents: `npx create-lz-oapp@latest --example <example>`
+
+6. **Helper Tasks (inline notice)**
+   - Goal: Know that helper tasks exist
+   - Contents: Statement + link to helper section
+
+7. **Setup**
+   - Goal: What to configure before running
+   - Contents: .env instructions, deployer account setup
+
+8. **Build**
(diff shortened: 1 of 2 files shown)


examples/oft-upgradeable/deploy/MyOFTUpgradeableMock.ts
St0rmBr3w, line 1: If you're changing OFTUpgradeable deploy scripts, you need a CHANGESET.

docs/EXAMPLES_SPECS.md
St0rmBr3w, line 1: I would change the title of this file - it doesn't clearly explain that this is a prompt for creating the README for the `./examples`

I would also get more specific with the type of information you want to see from the LLM.

docs/EXAMPLES_SPECS.md
```diff
@@ -0,0 +1,99 @@
...
+   - Goal: What is this example about
+   - Contents: Brief explanation if needed; otherwise rely on title + prerequisites
+
+6. **Requirements**
+   - Goal: What needs to be installed
+   - Contents: List tools + version numbers; consider noting testnet token needs
+
+7. **Scaffold this example**
+   - Goal: How to init the example
+   - Contents: `pnpm dlx create-lz-oapp@latest --example <name>`
```
St0rmBr3w: This only works if the example is public, have the model check `./packages/create-lz-oapp/src/config/` to see if the example is feature flagged and include it in the install instructions.

docs/EXAMPLES_SPECS.md
```diff
@@ -0,0 +1,99 @@
...
+    - Goal: How to deploy contracts/programs/modules
+    - Contents: Deploy command + minting instructions (if applicable)
+
+12. **Enable Messaging**
+    - Goal: How to set up OApps for use
+    - Contents: LZ config, init step, wiring step
+
+13. **Sending Message/OFT/ONFT**
+    - Goal: How to trigger cross-chain actions
+    - Contents: Send command(s) for both/all directions
```
St0rmBr3w: Define what send commands actually are.

St0rmBr3w (changes requested):
smol comments, overall looking better!

docs/EXAMPLES_SPECS.md
nazreen: @St0rmBr3w but the intention was not for this to be a file for prompting exclusively. The idea was this is for humans and coding agents, as started clearly in the first two sentences of the markdown.

> I would also get more specific with the type of information you want to see from the LLM.

on this, sure, I think regardless of human/agents, some of the points can be elaborated further. 

But from actually referencing the specs, in its given state, it's already performing well, I recommend you to try it on any of the other READMEs that are not yet revamped.

docs/EXAMPLES_SPECS.md
nazreen: ok that makes sense. will do.

10xKelly (approved):

examples/oft-upgradeable/deploy/MyOFTUpgradeableMock.ts
nazreen: pushed

PR #1600, opened 2025-07-03
audit patches + improvement

shankars99:
Fixes to the audit report


packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -159,53 +151,77 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
     /// @dev Always possible unless the lzCompose() fails due to an Out-Of-Gas panic
     function refund(bytes32 _guid, bytes calldata _extraOptions) external payable nonReentrant {
+        if (failedGuidState(_guid) != FailedState.CanOnlyRefund) revert CanNotRefund(_guid);
+
         FailedMessage memory failedMessage = failedMessages[_guid];
+        delete failedMessages[_guid];
+
         SendParam memory refundSendParam = failedMessage.refundSendParam;
-        if (failedGuidState(_guid) != FailedState.CanOnlyRefund) revert CanNotRefund(_guid);
+        address refundOft = failedMessage.oft;
```
gpersoon: @shankars99 it this ok?
Previousely this was `failedMessage.refundOFT` and now `failedMessage.oft` is used

In the case of `FailedState.CanOnlyRefund`: `failedMessage.oft == address(0))`, see:

https://github.com/LayerZero-Labs/devtools/blob/f4f5703eb7cd2b7280d9f69ce2c76ad7b8b3a073/packages/ovault-evm/contracts/OVaultComposer.sol#L262-L264

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -154,9 +154,9 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
     /// @dev Permissionless function to send back the message to the source chain
     /// @dev Always possible unless the lzCompose() fails due to an Out-Of-Gas panic
     function refund(bytes32 _guid, bytes calldata _extraOptions) external payable nonReentrant {
-        if (failedGuidState(_guid) != FailedState.CanOnlyRefund) revert CanNotRefund(_guid);
-
         FailedMessage memory failedMessage = failedMessages[_guid];
+        if (failedMessage.refundOFT != address(0)) revert CanNotRefund(_guid);
+
```
gpersoon: see https://github.com/LayerZero-Labs/devtools/pull/1600/commits/f4f5703eb7cd2b7280d9f69ce2c76ad7b8b3a073#r2184625897

also if refundOFT!=0 then failedMessage.oft could still be 0

so L163 probably should be:
```diff
-address refundOft = failedMessage.oft;
+address refundOft = failedMessage.refundOFT;
```

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -216,6 +232,21 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
+    }
+
+    function _executeHubTransfer(address _oft, address _receiver, uint256 _amountLD) internal {
+        IERC20 token = IERC20(IOFT(_oft).token());
+        token.transfer(_receiver, _amountLD);
+        if (msg.value > 0) {
+            (bool sent, ) = _receiver.call{ value: msg.value }("");
+            require(sent, "Failed to send Ether");
+        }
+        emit SentOnHub(_receiver, _oft, _amountLD);
```
gpersoon: emits are handled different on `send()` versus the other functions.

Via `send()` you can get two emits (when sendParam.dstEid == HUB_EID):
- emit SentOnHub(_receiver, _oft, _amountLD);
- emit Sent(_guid, oft);

Via for example `retry()` you just get (when sendParam.dstEid == HUB_EID):
- emit SentOnHub(_receiver, _oft, _amountLD);

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -154,9 +154,9 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
     /// @dev Permissionless function to send back the message to the source chain
     /// @dev Always possible unless the lzCompose() fails due to an Out-Of-Gas panic
     function refund(bytes32 _guid, bytes calldata _extraOptions) external payable nonReentrant {
-        if (failedGuidState(_guid) != FailedState.CanOnlyRefund) revert CanNotRefund(_guid);
-
         FailedMessage memory failedMessage = failedMessages[_guid];
+        if (failedMessage.refundOFT != address(0)) revert CanNotRefund(_guid);
```
gpersoon: Would be good to add a comment here, because this is now possible via two states.

Maybe better to do this, because then you keep using `_failedGuidState()`:
```solidity
FailedState f = _failedGuidState(failedMessage);
if (  (f != CanOnlyRefund) && (f != CanRetryWithSwapOrRefund) ) revert CanNotRefund(_guid);
```

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: oops that should be `.refundOFT`

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: yep it has to be `refundOFT` since this is the `refund` function lol. typo on my part. thanks for finding this

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: true. this pattern is cleaner

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: they now emit (in addition to a possible SentOnHub):
retry - Retried()
refund - Refunded()
retryWithSwap - RetriedWithSwap() instead of Sent() so that its unique to the operation

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -149,93 +156,110 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
         }
+    }
+
+    function sendFailedMessage(
+        bytes32 _guid,
+        address _oft,
+        SendParam memory _sendParam,
+        uint256 _prePaidMsgValue,
+        address _refundOverpayAddress
+    ) external payable {
```
gpersoon, line 254: probably good to also make this also `nonReentrant`, same as `send()`

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -149,93 +156,110 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
+            this.sendFailedMessage{ value: msg.value }(
+                _guid,
+                failedMessage.refundOFT,
+                failedMessage.refundSendParam,
+                failedMessage.msgValue,
+                REFUND_OVERPAY_ADDRESS
+            )
+        {
+            emit Refunded(_guid, failedMessage.refundOFT);
+        } catch {
```
gpersoon: So make sure this msg.value isn't lost:
```solidity
failedMessages[_guid].msgValue += msg.value
```

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -149,93 +156,110 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
-        if (sendParam.dstEid == HUB_EID) {
-            _executeHubTransfer(retryOFT, sendParam.to.bytes32ToAddress(), sendParam.amountLD);
-        } else {
-            sendParam.extraOptions = _extraOptions;
-            _send(retryOFT, sendParam);
+        if (removeExtraOptions) {
+            sendParam.extraOptions = "";
+            prePaidMsgValue = 0;
+
+            payable(REFUND_OVERPAY_ADDRESS).transfer(prePaidMsgValue);
```
gpersoon: i think this should be, otherwise you are sending 0:
```diff
-prePaidMsgValue = 0;
 payable(REFUND_OVERPAY_ADDRESS).transfer(prePaidMsgValue);
+prePaidMsgValue = 0;
```

gpersoon: why not also use `.call()`, like `_send()` does?

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -149,93 +156,110 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
         }
 
-        emit Retried(_guid, retryOFT);
+        try
+            this.sendFailedMessage{ value: msg.value }(_guid, failedMessage.oft, sendParam, prePaidMsgValue, tx.origin)
+        {
+            emit Retried(_guid, failedMessage.oft);
+        } catch {
```
gpersoon: should update the msg.value calculation, also taking into account the amount send to REFUND_OVERPAY_ADDRESS
```solidity
failedMessages[_guid].msgValue = prePaidMsgValue + msg.value;
```

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -149,93 +156,110 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
+                this.sendFailedMessage{ value: msg.value }(
+                    _guid,
+                    failedMessage.oft,
+                    failedMessage.sendParam,
+                    failedMessage.msgValue,
+                    tx.origin
+                )
+            {
+                emit Retried(_guid, failedMessage.oft);
+            } catch {
```
gpersoon, line 241: should do:
```solidity
failedMessages[_guid].msgValue += msg.value
```

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: making this `nonReentrant` would not play nice with the external calls made from retry/refunds (which have the `nonReentrant` modifiers) right?
```
try
    this.sendFailedMessage{ value: msg.value }(
        _guid,
        failedMessage.oft,
        failedMessage.sendParam,
        failedMessage.msgValue,
        tx.origin
    )       
```

it being `onlySelf` and external with the only functions that can call it being:
1. `refund()`
2. `retry()`
3. `retryWithSwap()` 

makes it fine?

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: oh yeah, good find. thanks

shankars99: done, thanks

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: i think on `refund()` we can just `revert()` in the catch since it being atomic is alright (there's no partial state that we can reach like with `retryWithSwap()`)

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: reverting at `retry` - no partial state required

packages/ovault-evm/contracts/OVaultComposer.sol
gpersoon: If you are going to revert there is no need for a try catch

packages/ovault-evm/contracts/OVaultComposer.sol
gpersoon: Yes thats true, should be good indeed

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: ah true, changed

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -126,116 +152,197 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
+                failedMessages[_guid].msgValue += msg.value;
+                emit SendFailed(_guid, failedMessage.oft);
+            }
```
gpersoon, line 244: when `skipRetry == true` something should also be done with msg.value:
either revert or add it to `failedMessages[_guid].msgValue`

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -126,116 +152,197 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
+    /// @dev Helper to check if the target OFT does not have a peer set for the destination chain OR if our target chain is not the same as the HUB chain
     function _isInvalidPeer(address _oft, uint32 _dstEid) internal view returns (bool) {
         return _dstEid != HUB_EID && IOAppCore(_oft).peers(_dstEid) == bytes32(0);
     }
 
-    /// @dev Helper to view the state of a failed message
-    function failedGuidState(bytes32 _guid) public view returns (FailedState) {
-        FailedMessage memory failedMessage = failedMessages[_guid];
+    function _removeDust(address _oft, uint256 _amount) internal view returns (uint256) {
+        uint256 decimalConversionRate = _oft == address(ASSET_OFT)
```
gpersoon, line 324: If is difficult to reason about which oft is which:
- `lzCompose - _refundOFT` is the oft that is received and `lzCompose - oft` is the one that is send
- in (_)executeOVaultActionWithSlippageCheck this is called oft, but is then one that is received
- in _checkSlippage and _removeDust: oft is also that one that is received

Suggestion: use more consistent naming in all functions (maybe oftReceived and oftConverted)

packages/ovault-evm/contracts/OVaultComposer.sol
```diff
@@ -126,116 +152,197 @@ contract OVaultComposer is IOVaultComposer, ReentrancyGuard {
...
+
+        uint256 srcAmount = failedMessage.refundSendParam.amountLD;
+
+        uint256 vaultAmount = _executeOVaultActionWithSlippageCheck(
+            failedMessage.refundOFT,
+            srcAmount,
+            failedMessage.sendParam.minAmountLD
+        );
+
+        failedMessage.sendParam.amountLD = vaultAmount;
```
gpersoon: This isn't persisted in storage (e.g. failedMessages[_guid]), so this is lost when `skipRetry==true` or `this.sendFailedMessage()` reverts.
This function can't be executed again because `failedMessages[_guid].refundOFT = address(0);` so the funds are lost

shankars99: ah yes

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: a `retry` seems to be cleaner

packages/ovault-evm/contracts/OVaultComposer.sol
shankars99: yep `executeOVaultActionWithSlippageCheck` and `_removeDust` take in refundOFT.
but `_checkSlippage` takes in the targetOFT.

updating to `refundOFT` and `oft`

or do you think `vaultInOFT` and `vaultOutOFT` would be better?

PR #1605, opened 2025-07-08
feat: add slim testhelper

St0rmBr3w:


packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,324 @@
...
+            if (_executorOptionExists(options, ExecutorOptions.OPTION_TYPE_LZRECEIVE)) {
+                this.lzReceive(packetBytes, options);
+            }
+            
+            // Handle native drops
+            if (_executorOptionExists(options, ExecutorOptions.OPTION_TYPE_NATIVE_DROP)) {
+                (uint256 amount, bytes32 receiver) = _parseExecutorNativeDropOption(options);
+                address receiverAddress = address(uint160(uint256(receiver)));
+                (bool success, ) = receiverAddress.call{value: amount}("");
+                require(success, "Native drop failed");
```
ravinagill15: Perhaps we should include the receiver address in the error here

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,324 @@
...
+        }
+        
+        while (numberOfPackets > 0) {
+            numberOfPackets--;
+            bytes32 guid = queue.popBack();
+            bytes memory packetBytes = packets[guid];
+            this.assertGuid(packetBytes, guid);
+            this.validatePacket(packetBytes, _resolvedPayload);
+
+            bytes memory options = optionsLookup[guid];
```
ravinagill15, line 347: `packets` and `optionsLookup` never get deleted after being used, which might cause memory leaks during fuzz testing or long-running tests. Perhaps we can delete the entry for that `guid` in `packets` and `optionsLookup` after delivery?

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,324 @@
...
+        DoubleEndedQueue.Bytes32Deque storage queue = packetsQueue[_dstEid][_dstAddress];
+        uint256 pendingPacketsSize = queue.length();
+        uint256 numberOfPackets;
+        if (_packetAmount == 0) {
+            numberOfPackets = queue.length();
+        } else {
+            numberOfPackets = pendingPacketsSize > _packetAmount ? _packetAmount : pendingPacketsSize;
+        }
+        
+        while (numberOfPackets > 0) {
```
ravinagill15: I think to be safe we could update this to be

```suggestion
        while (numberOfPackets > 0 && queue.length() > 0) {
```

Since we are popping from the queue during the while loop

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,324 @@
...
+     * @notice Schedules a packet for delivery
+     */
+    function schedulePacket(bytes calldata _packetBytes, bytes calldata _options) public {
+        uint32 dstEid = _packetBytes.dstEid();
+        bytes32 dstAddress = _packetBytes.receiver();
+        DoubleEndedQueue.Bytes32Deque storage queue = packetsQueue[dstEid][dstAddress];
+        bytes32 guid = _packetBytes.guid();
+        queue.pushFront(guid);
+        packets[guid] = _packetBytes;
+        optionsLookup[guid] = _options;
```
ravinagill15: Should we be checking for duplicate guids in the queue?

St0rmBr3w: That shouldn't happen since the GUID is constructed using the normal LZ GUID generation.

ravinagill15 (dismissed):
🚀

examples/oft/test/foundry/MyOFT.t.sol
```diff
@@ -43,7 +43,7 @@ contract MyOFTTest is TestHelperOz5 {
         vm.deal(userB, 1000 ether);
 
         super.setUp();
-        setUpEndpoints(2, LibraryType.UltraLightNode);
+        setUpEndpoints(2);
```
shankars99:
```suggestion
        setUpEndpoints(2); // setUpEndpoints(2, LibraryType.UltraLightNode); also works 
```
not a fan of breaking the interface, since we do `setUpEndpoints(a,/*b*/)` it might be good to comment that you can also stick to the old interface and ONLY update the `SlimLzTestHelper`import to use it

packages/test-devtools-evm-foundry/contracts/mocks/SlimSimpleMessageLibMock.sol
```diff
@@ -0,0 +1,98 @@
...
+        // No-op - packets are always valid in testing
+    }
+
+    // Required interface functions with minimal implementation
+    function messageLibType() external pure returns (MessageLibType) {
+        return MessageLibType.Send;
+    }
+
+    function version() external pure returns (uint64 major, uint8 minor, uint8 endpointVersion) {
+        return (1, 0, 2);
```
shankars99:
```suggestion
        return (0, 0, 0);
```

since its a custom impl?

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+    
+    /// @dev Thrown when packet validation fails
+    /// @param guid The GUID of the packet that failed validation
+    error SlimLzTestHelper_PacketValidationFailed(bytes32 guid);
+    
+    /// @dev Thrown when attempting to schedule a packet with duplicate GUID
+    /// @param guid The duplicate GUID that was detected
+    error SlimLzTestHelper_DuplicateGuid(bytes32 guid);
+    using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
+    using PacketV1Codec for bytes;
```
shankars99: nit: move to top of file

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    // Core mappings
+    mapping(uint32 => mapping(bytes32 => DoubleEndedQueue.Bytes32Deque)) packetsQueue;
+    mapping(bytes32 => bytes) packets;
+    mapping(bytes32 => bytes) optionsLookup;
+    mapping(uint32 => address) endpoints;
+
+    // Constants
+    uint128 public executorValueCap = 0.1 ether;
```
shankars99, line 89:
```suggestion
    uint128 public constant EXECUTOR_VALUE_CAP = 0.1 ether;
```

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+        _setUpUlnOptions();
+    }
+
+    /**
+     * @dev set executorValueCap if more than 0.1 ether is necessary
+     * @param _valueCap amount executor can pass as msg.value to lzReceive()
+     */
+    function setExecutorValueCap(uint128 _valueCap) public {
+        executorValueCap = _valueCap;
+    }
```
shankars99: Line 81 says that `executorValueCap` is a constant

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+    function endpoint() external view returns (ILayerZeroEndpointV2 iEndpoint);
+}
+
+/**
+ * @title SlimLzTestHelper
+ * @notice Lightweight helper contract for basic LayerZero OApp testing without compile size issues.
+ * @dev Maintains backward compatibility with TestHelperOz5 for basic testing scenarios.
+ * @dev For advanced features (DVN, Executor, workers), use TestHelperOz5 instead.
+ */
+contract SlimLzTestHelper is Test, OptionsHelper {
```
shankars99, line 38: would be nice to add fork test support into this

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+     */
+    function setExecutorValueCap(uint128 _valueCap) public {
+        executorValueCap = _valueCap;
+    }
+
+    /**
+     * @notice Sets up endpoints - maintains backward compatibility
+     * @param _endpointNum Number of endpoints to create
+     */
+    function setUpEndpoints(uint8 _endpointNum, LibraryType /* _libraryType */) public {
```
shankars99:
```suggestion
    function setUpEndpoints(uint8 _endpointNum) public {
```

this one, being the base implementation only takes in `_endpointNum`

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+            }
+        }
+    }
+
+    /**
+     * @notice Overload for backward compatibility
+     */
+    function setUpEndpoints(uint8 _endpointNum) public {
+        setUpEndpoints(_endpointNum, LibraryType.SimpleMessageLib);
+    }
```
shankars99:
```suggestion
    function setUpEndpoints(uint8 _endpointNum, LibraryType /* _libraryType */) public {
        setUpEndpoints(_endpointNum);
    }
```

this is a wrapper so it takes in `LibraryType` but subcalls `setUpEndpoints(uint8)` which is the implementation

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+     */
+    function _deployOApp(bytes memory _oappBytecode, bytes memory _constructorArgs) internal returns (address addr) {
+        bytes memory bytecode = bytes.concat(abi.encodePacked(_oappBytecode), _constructorArgs);
+        assembly {
+            addr := create(0, add(bytecode, 0x20), mload(bytecode))
+            if iszero(extcodesize(addr)) {
+                revert(0, 0)
+            }
+        }
+    }
```
shankars99: nit: would re-order them in terms of when the function would be called in a test lifecycle, so 

1. _deployOapp()
2. setupOapps()
3. wireOapps()
etc

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+
+    function addressToBytes32(address _addr) internal pure returns (bytes32) {
+        return bytes32(uint256(uint160(_addr)));
+    }
+
+    receive() external payable {}
+}
+
+// Import ExecutorOptions for compatibility
+import { ExecutorOptions } from "@layerzerolabs/lz-evm-messagelib-v2/contracts/libs/ExecutorOptions.sol";
```
shankars99: umm whys the import at the bottom of the file lol

packages/test-devtools-evm-foundry/contracts/SlimLzTestHelper.sol
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    function registerEndpoint(EndpointV2 endpoint) public {
+        endpoints[endpoint.eid()] = address(endpoint);
+    }
+
+    function addressToBytes32(address _addr) internal pure returns (bytes32) {
+        return bytes32(uint256(uint160(_addr)));
+    }
```
shankars99, line 541: is there a package where this can be re-used?

packages/test-devtools-evm-foundry/contracts/mocks/EndpointV2Simple.sol
```diff
@@ -0,0 +1,160 @@
...
+    event PacketSent(bytes encodedPayload, bytes options, address sendLibrary);
+    event PacketDelivered(Origin origin, address receiver);
+    
+    constructor(uint32 _eid) {
+        eid = _eid;
+    }
+
+    /// @notice Simplified quote - just returns a fixed fee for testing
+    function quote(MessagingParams calldata, address) external pure returns (MessagingFee memory) {
+        return MessagingFee({ nativeFee: 0.001 ether, lzTokenFee: 0 });
```
shankars99: nit: dont think this is a good idea, since options are something that should be tested locally. 

moreover can use `MessagingParams.dstEid` and `MessagingParams.message` to create something that would not be the same for every network and is a factor of the message length (something as simple as `nativeFee: 0.001 ether + dstEid + message.length`)

would also use `MessagingParam.payInLzToken` to flip between `nativeFee` and `lzTokenFee` for future proofing - a small change that would save as future PR when this is eventually merged in

shankars99 (changes requested):
Would also be good to make all the no-op function `virtual` so that someone, if they want can override it and have a custom EndpointV2Simple or something. 

will re-review after any changes

shankars99 (approved):
lgtm

PR #1621, opened 2025-07-18
[INT-500] Aptos bigint export fix

AlexanderLiteplo:


packages/devtools-move/tasks/move/utils/utils.ts
```diff
@@ -385,7 +385,13 @@ function formatInitiaArgumentValue(arg: any): any {
 
 function formatArgumentValue(arg: any): any {
     if (Array.isArray(arg)) {
-        return arg.map((item: any) => item)
+        return arg.map((item: any) => formatArgumentValue(item))
```
ItsAdel, line 388: Not too familiar with what type its formatting through but this code handles nested arrays correctly because it is recursive. However, if it receives an object, it doesn't check any of the properties inside it. If its intended then I can approve

AlexanderLiteplo: This just covers the case where we have an array of bigints - before we would return the argument as is (just the bigint) now we format those as strings. A bit confusing to use recursion here but we aren't expecting to have nested arrays so it makes the code a lot cleaner instead of having to add an if statement inside this mapping.

tldr: its intentional

ItsAdel (approved):
lgtm!

EWCunha (approved):
Alex, the programing rockstar!

PR #1626, opened 2025-07-21
hyperliquid - composer  v1 with extensions

shankars99:
Despite the branch name this will be our first stable version release of the hyperliquid composer

### Changes:
1. Adding a recovery extension that is permissioned to retrieve arbitrary amounts of `HYPE` and `ERC20` from `HyperCore` to `HyperEVM` and the trusted `RECOVERY_ADDRESS`
2. Since `USDC` does not natively exist on HyperEVM the funds can be transferred to a target address on HyperCore
3. Target ERC20 and HYPE transfers on HyperEVM
4. Simplifying code by replacing replacing `validate_message()`+`validate_msg_or_refund()` with `decode_message()`+`refund()`

### Recovery extension:
This is an extension to the `HyperliquidComposer` that lets a dedicated (and trusted) address, configured on deployment to pull funds out of the composer on `hypercore` or `hyperevm`. There are a few situations where this maybe triggered
1. Failed `composer->receiver` transfer on hypercore which results in funds being stuck on the composer's address on hypercore
2. Failed precompile interaction on the composer leading to locked funds

### Simplifying composeMessage decoding
Earlier the `composeMsg` was `abi.encode(address)/encodePacked(address)`. This design was to allow us to transfer the tokens to the `sender` on hyperevm if we had a failed decoding of `composeMsg`, in the case the sender was non-evm then the funds would be locked forever. 

We are changing this to accept `abi.encode(uint256 minMsgValue, address _receiver)` and if this fails we have a new function `refund(bytes32 _guid)` that lets anyone permissionlessly send funds back to the sender on the source chain. We can do this because the OFT on hyperevm appends onto the compose message:
1. `srcEid`
2. `amountLD`
And the source OFT adds the `msg.sender` before the `composeMsg`

The triplet of `(sender, srcEid, amountLD)` is enough information to trigger a refund. This operation would rely on `enforcedOptions` since we do not want to allow arbitrary options (attacked can use `NativeDrop` and consume some of the stored `msg.value`)

The refund transaction consumes the `msg.value` cached on the `lzCompose(...args)` call and also accepts in additional `msg.value`. Extra `msg.value` is sent to a dedicated address `REFUND_ADDRESS` so users should be sure to not overpay, this is to prevent complicated operations where MEV bots can take the overpayment if the `lzCompose.value` the user paid for is greater than the `oft.send()`. Since adding in nativeDrop options into the composeMsg makes compute difficult, we are avoiding that and users can only recover their `ERC20`   

2 Other PRs have been merged into this https://github.com/LayerZero-Labs/devtools/pull/1683 into https://github.com/LayerZero-Labs/devtools/pull/1681 into this branch. Please reference them.


examples/oft-hyperliquid/deploy/MyHyperLiquidComposer_Recoverable.ts
```diff
@@ -0,0 +1,121 @@
...
+const contractName_oft = 'MyOFT'
+const contractName_composer = 'HyperLiquidComposer_Recoverable'
+const recoveryAddress = '0x0000000000000000000000000000000000000000'
+
+const deploy: DeployFunction = async (hre) => {
+    if (recoveryAddress == '0x0000000000000000000000000000000000000000') {
+        throw new Error(
+            'Recovery address is not defined. Please set the recoveryAddress variable in the deploy script.'
+        )
+    }
```
ravinagill15: Might be better to have avoid editing the deploy scripts, and maybe add the recovery address as a named account within the `hardhat.config.ts`? This would hopefully minimize the number of edits needed to get this example deployed since hardhat config usually needs to be updated anyway with networks and their rpcs

examples/oft-hyperliquid/contracts/extensions/MyHyperLiquidComposer_Recoverable.sol
```diff
@@ -0,0 +1,34 @@
...
+    /// @param _hlIndexId The HyperLiquid core spot's index value
+    /// @param _assetDecimalDiff The difference in decimals between the HyperEVM's ERC20 and the HyperLiquid HIP-1 token
+    ///                 (i.e. 18 decimals on evm and 6 on HyperLiquid would be 18 - 6 = 12)
+    /// @param _recoveryAddress The address to which funds can be recovered.
+    constructor(
+        address _oft,
+        uint64 _hlIndexId,
+        int8 _assetDecimalDiff,
+        address _recoveryAddress
+    ) HyperLiquidComposer(_oft, _hlIndexId, _assetDecimalDiff) RecoverableComposer(_recoveryAddress) {}
```
ravinagill15: I wonder if we should validate that the recovery address is not the zero address within the contract constructor too? I see it is validated in the deploy script, but since it is an important role, might be good to enforce here too

examples/oft-hyperliquid/test/foundry/HyperLiquidComposer.t.sol
```diff
@@ -23,24 +22,23 @@ contract HyperLiquidComposerTest is HyperliquidBaseTest {
     }
 
     function test_deployment() public view {
-        IHyperAsset memory hypeAsset = hyperLiquidComposer.getHypeAsset();
-        assertEq(hypeAsset.assetBridgeAddress, 0x2222222222222222222222222222222222222222);
+        assertEq(hyperLiquidComposer.NATIVE_ASSET_BRIDGE(), 0x2222222222222222222222222222222222222222);
```
ravinagill15, line 25: nit: let's place `0x2222222222222222222222222222222222222222` into a constant variable like `EXPECTED_NATIVE_ASSET_BRIDGE`

examples/oft-hyperliquid/test/foundry/HyperLiquidComposer.t.sol
```diff
@@ -23,24 +22,23 @@ contract HyperLiquidComposerTest is HyperliquidBaseTest {
...
     }
 
     function test_hypeIndexByChainId_testnet() public {
         string memory rpcUrl = "https://rpc.hyperliquid-testnet.xyz/evm";
+
         try vm.envString("RPC_URL_HYPERLIQUID_TESTNET") returns (string memory _rpcUrl) {
             rpcUrl = _rpcUrl;
         } catch {
-            console.log("Using default mainnet RPC URL");
+            console.log("Using default testnet RPC URL");
```
ravinagill15, line 41: Not necessary for the current PR, but we can set the default rpc to the lz testnet proxy rpc referenced as an environment variable, then update github secrets for devtools. This would be more resilient than a public rpc for tests

examples/oft-hyperliquid/test/foundry/ComposerCodec/ComposeMessage.t.sol
```diff
@@ -22,11 +20,10 @@ contract ComposeMessageTest is HyperliquidBaseTest {
         super.setUp();
     }
 
-    function test_validateAndDecodeMessage_EncodingVariants(address _receiver, bool _useEncodePacked) public view {
+    function test_validateAndDecodeMessage(address _receiver) public view {
```
ravinagill15, line 23: We are only testing happy path here, perhaps we should add tests for when receiver is missing, or is the zero address

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
tinom9, line 1: There are many dust references in this file that should be removed.

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
```diff
@@ -1,214 +1,133 @@
...
 
-    /// @notice Converts an asset bridge address to a core index id
-    /// @notice This function is called by the HyperLiquidComposer contract
-    ///
-    /// @param _assetBridgeAddress The asset bridge address to convert
-    ///
-    /// @return _coreIndexId The core index id
+    /**
+     * @notice Converts an asset bridge address to a core index id
+     * @notice This function is called by the HyperLiquidComposer contract
```
tinom9: It's not actually called.

```suggestion
```

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
```diff
@@ -1,214 +1,133 @@
...
+     * @param _assetBridgeAddress The asset bridge address to convert
+     * @return _coreIndexId The core index id
+     */
     function into_tokenId(address _assetBridgeAddress) internal pure returns (uint256) {
```
tinom9: This should be returning an `uint64`?

packages/hyperliquid-composer/contracts/HyperLiquidCore.sol
```diff
@@ -0,0 +1,66 @@
...
+    address internal constant CORE_USER_EXISTS_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000000810;
+
+    /// @dev Pre-computed headers for gas efficiency
+    /// @dev bytes1 CORE_WRITER_VERSION = 0x01;
+    /// @dev bytes3 SPOT_SEND_ACTION_ID = 0x000006;
+    bytes4 public constant SPOT_SEND_HEADER = 0x01000006; // Pre-computed concatenation
+
+    function spotBalance(address user, uint64 token) public view returns (SpotBalance memory) {
+        (bool success, bytes memory result) = SPOT_BALANCE_PRECOMPILE_ADDRESS.staticcall(abi.encode(user, token));
+        require(success, "SpotBalance precompile call failed");
```
tinom9, line 49: Why revert strings instead of custom errors in this file?

examples/oft-hyperliquid/test/foundry/ComposerCodec/TypeConversion.t.sol
```diff
@@ -2,14 +2,16 @@
...
-import { IHyperAsset, IHyperAssetAmount } from "@layerzerolabs/hyperliquid-composer/contracts/interfaces/IHyperLiquidComposerCore.sol";
+import { IHyperAssetAmount } from "@layerzerolabs/hyperliquid-composer/contracts/interfaces/IHyperLiquidComposer.sol";
 
 import { Test, console } from "forge-std/Test.sol";
 
 contract TypeConversionTest is Test {
     int8 MIN_DECIMAL_DIFF = -2;
     int8 MAX_DECIMAL_DIFF = 8;
 
+    uint64 MAX_TRANSFER_PER_TX = type(uint64).max;
```
ravinagill15: [MAX_DECIMAL_DIFF in HyperLiquidComposer is 18](https://github.com/LayerZero-Labs/devtools/blob/3e0b7d1a1859748a74d298d94db3ab30c124b4a4/packages/hyperliquid-composer/contracts/HyperLiquidComposer.sol#L37) -- should we change that here as well?

examples/oft-hyperliquid/test/foundry/HyperLiquidComposer.t.sol
shankars99: thats what we do rn

examples/oft-hyperliquid/test/foundry/ComposerCodec/ComposeMessage.t.sol
shankars99: more testing is under packages. examples has a subset of those

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
```diff
@@ -1,214 +1,133 @@
...
-    /// @param _coreIndexId The core index id to convert
-    ///
-    /// @return _assetBridgeAddress The asset bridge address
+    /**
+     * @notice Converts a core index id to an asset bridge address
+     * @notice This function is called by the HyperLiquidComposer contract
+     * @param _coreIndexId The core index id to convert
+     * @return _assetBridgeAddress The asset bridge address
+     */
     function into_assetBridgeAddress(uint256 _coreIndexId) internal pure returns (address) {
```
tinom9: `_coreIndexId` should be a `uint64`?

examples/oft-hyperliquid/test/foundry/ComposerCodec/TypeConversion.t.sol
shankars99: we should yes

examples/oft-hyperliquid/test/foundry/ComposerCodec/ComposeMessage.t.sol
shankars99: do you think we should copy all of them over? i thought that they're core composer tests so developers arent really gonna be testing against them when they build an application

examples/oft-hyperliquid/test/foundry/HyperLiquidComposer.t.sol
shankars99: i wanted that to be clear to test against that value. didnt want 2 indirections in the assert if that makes sense?

packages/hyperliquid-composer/contracts/HyperLiquidComposer.sol
```diff
@@ -1,243 +1,298 @@
...
-            bytes memory errMsg = completeRefund(_err, _executor);
-            emit ErrorMessage(errMsg);
-            // Pre-emptive return after the refund
-            return;
-        }
+            uint256 minGas = msg.value > 0 ? MIN_GAS_WITH_VALUE() : MIN_GAS();
+
+            /// @dev Gas check before executing hypercore precompile operations. Can be retried from the endpoint with sufficient gas.
+            /// @dev Contracts would need to called with more gas than this to account for the code above this line.
+            if (gasleft() < minGas) revert InsufficientGas(gasleft(), minGas);
```
tinom9, line 111: Are we taking into account the 63/64 gas forwarding rule?

examples/oft-hyperliquid/contracts/extensions/MyHyperLiquidComposer_Recoverable.sol
shankars99: yes we should yes

examples/oft-hyperliquid/deploy/MyHyperLiquidComposer_Recoverable.ts
shankars99: good idea

packages/hyperliquid-composer/contracts/HyperLiquidComposer.sol
```diff
@@ -1,243 +1,298 @@
...
+     * @param _refundAddress The address to refund tokens to
+     * @param _amountLD The amount of ERC20 tokens to refund
+     */
+    function _refundToHyperEvm(address _refundAddress, uint256 _amountLD) internal virtual {
+        if (msg.value != 0) {
+            (bool success1, ) = _refundAddress.call{ value: msg.value }("");
+            if (!success1) {
+                (bool success2, ) = tx.origin.call{ value: msg.value }("");
+                if (!success2) revert NativeTransferFailed(msg.value);
             }
```
tinom9, line 249: A bit weird to be sending ETH to a non-user address. I'd prefer just reverting and letting the user call `refundToSrc`.

examples/oft-hyperliquid/contracts/extensions/MyHyperliquidComposer_FeeToken.sol
```diff
@@ -0,0 +1,30 @@
...
+/// @dev This Extension is for OFTs like USDT0 that can be used to pay activation fee for users on hypercore.
+///
+/// @dev This contract does NOT refund dust to the receiver because we do not expect any due to truncation of sharedDecimals.
+///      In the off-chance that you have dust you would have to implement dust refunds to the receiver in:
+///      `_transferERC20ToHyperCore` and `_transferNativeToHyperCore`
+///
+/// @dev Disclaimer: If your token's evm total supply exceeds your asset bridge's balance when scaled to EVM, it is possible
+///      that the composer will not be able to send the tokens to the receiver address on hypercore due to bridge consumption.
+///      Tokens would instead be returned to the sender address on HyperEVM. Front-end handling is recommended.
+contract MyHyperliquidComposer_FeeToken is FeeToken {
```
tinom9:
```suggestion
contract MyHyperLiquidComposer_FeeToken is FeeToken {
```

Filename should probably also be changed.

packages/hyperliquid-composer/contracts/extensions/FeeToken.sol
```diff
@@ -0,0 +1,53 @@
...
+     * @param _to The address to check
+     * @param _coreAmount The core amount to transfer
+     * @return The final core amount to transfer (same as _coreAmount in default impl)
+     */
+    function _getFinalCoreAmount(address _to, uint64 _coreAmount) internal view override returns (uint64) {
+        if (coreUserExists(_to).exists) return _coreAmount;
+
+        uint64 fee = activationFee();
+        if (_coreAmount < fee) revert InsufficientCoreAmountForActivation();
+        return _coreAmount - fee;
```
tinom9: What happens when `_coreAmount == fee`?

Should probably document it.

packages/hyperliquid-composer/contracts/HyperLiquidCore.sol
shankars99: thats from hyperliquid. they dont have packages and isntead give you code to copy paste

packages/hyperliquid-composer/contracts/HyperLiquidComposer.sol
shankars99: no this is to simulate enforcedOptions since we do not set them for compose messages

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
shankars99: being type safe yes it should

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
shankars99: being type safe yes it should

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
shankars99: removed

packages/hyperliquid-composer/contracts/library/HyperLiquidComposerCodec.sol
shankars99: cleaned

packages/hyperliquid-composer/contracts/extensions/FeeToken.sol
shankars99: ah good one. if its 0 then the tx fails. will change it to `<=`

examples/oft-hyperliquid/contracts/extensions/MyHyperliquidComposer_FeeToken.sol
shankars99: done

tinom9 (approved):

ravinagill15 (approved):

PR #1629, opened 2025-07-24
ovault atomic synchronous composer

shankars99:


packages/ovault-evm/contracts/SynchronousVaultComposer.sol
```diff
@@ -0,0 +1,218 @@
...
+        bytes32 _composeFrom,
+        bytes memory _composeMsg,
+        uint256 _amount
+    ) external payable {
+        // Can only be called by self
+        if (msg.sender != address(this)) revert OnlySelf(msg.sender);
+
+        // TODO comment why we do this
+        (SendParam memory sendParam, uint256 minMsgValue) = abi.decode(_composeMsg, (SendParam, uint256));
+        if (msg.value < minMsgValue) revert InvalidMsgValue(minMsgValue, msg.value); /// @dev only happens on lzCompose
```
St0rmBr3w: Is this reasonable to include? I know that the non-EVM implementation is out of scope, but I think relying on abi.decode forces the non-EVM implementation to adhere to the solidity ABI format for custom structs, which might not be ideal.

packages/ovault-evm/contracts/SynchronousVaultComposer.sol
```diff
@@ -0,0 +1,218 @@
...
+
+        if (_oft == ASSET_OFT) {
+            _depositAndSend(_composeFrom, _amount, sendParam, tx.origin);
+        } else {
+            _redeemAndSend(_composeFrom, _amount, sendParam, tx.origin);
+        }
+    }
```
St0rmBr3w: Just confirming this is expected, this will always refund the Executor. Would it be better to let the deployer of the composer set this?

St0rmBr3w: Maybe even a malicious actor would want to frontrun to steal gas? Unsure if this is best.

packages/ovault-evm/test/synchronous-vault/OVaultComposer_Base.t.sol
```diff
@@ -124,17 +121,37 @@ contract OVaultComposerBaseTest is TestHelperOz5 {
...
+            amountWithOffset = _amount - modOffset;
+        } else {
+            // If offset is positive, we can safely add it
+            amountWithOffset = _amount + uint128(_offset);
+        }
+        return _removeDust(amountWithOffset);
+    }
+
+    function _removeDust(uint256 _amount) internal pure returns (uint256, uint256) {
+        uint256 dust = _amount % 10 ** 12;
```
St0rmBr3w: Would it be better to fuzz this with the decimalConversionRate assumption like we do in OFT.t.sol?

packages/ovault-evm/contracts/SynchronousVaultComposer.sol
shankars99: why not? its a composeMsg which is almost always generated off-chain, you can create the composeMsg as bytes with ethers and encode it into whatever you want

packages/ovault-evm/contracts/SynchronousVaultComposer.sol
shankars99: yep its to the executor

shankars99: You can't unless you wanna pay the minMsgValue

PR #1637, opened 2025-07-28
DEVREL-634 Fix: enable metadata tools to append custom executors

St0rmBr3w:


DanL0 (approved):

packages/metadata-tools/src/config-metadata.ts
```diff
@@ -96,9 +125,16 @@ export async function translatePathwayToConfig(
...
+        // Use custom executor, resolving name to address if needed
+        sourceExecutor = executorNameToAddress(customExecutor, sourceLZDeployment.chainKey, metadata)
+    } else {
+        // Use default executor from deployment
+        sourceExecutor = isSolanaDeployment(sourceLZDeployment)
+            ? sourceLZDeployment.executor?.pda
+            : sourceLZDeployment.executor?.address
+    }
 
     if (!sourceExecutor) {
```
AlexanderLiteplo: from my understanding this should be triggered if executorNameToAddress does not find the custom executor

packages/metadata-tools/src/config-metadata.ts
```diff
@@ -70,6 +70,34 @@ export function DVNsToAddresses(dvns: string[], chainKey: string, metadata: IMet
     return dvnAddresses.sort()
 }
 
+export function executorNameToAddress(executorName: string, chainKey: string, metadata: IMetadata): string {
```
AlexanderLiteplo: I think this function name does not make a lot of sense because we are just returning the name if nothing is found. Also, why not return null or undefined instead of just the name in the case if not found in custom executors?

packages/metadata-tools/src/config-metadata.ts
```diff
@@ -177,9 +213,16 @@ export async function translatePathwayToConfig(
     }
 
     if (destinationToSourceConfirmations) {
-        const destinationExecutor = isSolanaDeployment(destinationLZDeployment)
-            ? destinationLZDeployment.executor?.pda
-            : destinationLZDeployment.executor?.address
+        let destinationExecutor: string | undefined
+        if (customExecutor) {
+            // Use custom executor, resolving name to address if needed
```
AlexanderLiteplo: nit - code duplication. potentially we should extract a method with a name that makes it more readable or do a larger refactor of the file because it looks like other stuff can be extracted too

packages/metadata-tools/src/types.ts
```diff
@@ -8,6 +8,7 @@ export type TwoWayConfig = [
     [string[], [string[], number] | []], // [requiredDVNs, [optionalDVNs, threshold]]
     [number, number | undefined], // [srcToDstConfirmations, dstToSrcConfirmations]
     [OAppEnforcedOption[] | undefined, OAppEnforcedOption[] | undefined], // [enforcedOptionsSrcToDst, enforcedOptionsDstToSrc]
+    string?, // customExecutor - optional custom executor for send config
```
AlexanderLiteplo: we can remove comments here if do something like this:

```
export type TwoWayConfig = [
  srcContract: OmniPointHardhat,
  dstContract: OmniPointHardhat,
  dvnConfig: DVNConfig,
  confirmations: Confirmations,
  enforcedOptions: EnforcedOptions,
  customExecutor?: string
];
```

AlexanderLiteplo: much cleaner

ItsAdel (approved):

PR #1638, opened 2025-07-28
hlp/sdk - activated users and core spot balances 

shankars99:
Using `spotClearinghouseState` for 
### Display activated accounts 
```bash
npx @layerzerolabs/hyperliquid-composer is-account-activated \
    --user <0x> \
    --network {testnet | mainnet} \
    [--log-level {info | verbose}]
```
prints `true` or `false` 

function `isAccountActivated(...args) -> boolean`  for api users

### Print out user balances as a table
 ```bash
npx @layerzerolabs/hyperliquid-composer get-core-balances \ 
    --user <0x> \
    [--show-zero {false | true}] \ 
    --network {testnet | mainnet} \
    [--log-level {info | verbose}]
```

prints a table of balances 

function `getCoreBalances(...args) -> SpotBalancesResponse` for api users

### CODE REVIEW

File: packages/hyperliquid-composer/src/types/operations.ts [hunk 1]
```diff
@@ -56,3 +56,10 @@ export interface SpotDeployAction extends BaseExchangeRequest {
               registerHyperliquidity: RegisterHyperliquidity
           }
 }
+
+export interface SpotClearinghouseState extends BaseExchangeRequest {
```
ravinagill15, line 60: nit: do we want to use camel case for this -- `SpotClearingHouseState`?

packages/hyperliquid-composer/src/io/cli-output-formatter.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (!showZeroBalances) {
+        filteredBalances = balances.balances.filter((balance) => parseFloat(balance.total) > 0)
+    }
+
+    filteredBalances.sort((a, b) => parseFloat(b.total) - parseFloat(a.total))
+
+    for (const balance of filteredBalances) {
+        const total = parseFloat(balance.total)
+        const hold = parseFloat(balance.hold)
+        const entryNtl = parseFloat(balance.entryNtl)
```
ravinagill15, line 30: Should we validate these are numbers before parsing?

packages/hyperliquid-composer/src/io/cli-output-formatter.ts
```diff
@@ -0,0 +1,46 @@
...
+
+        table.push([
+            balance.coin,
+            balance.token.toString(),
+            total === 0 ? '0.0' : total.toFixed(8).replace(/\.?0+$/, ''),
+            hold === 0 ? '0.0' : hold.toFixed(8).replace(/\.?0+$/, ''),
+            entryNtl === 0 ? '0.0' : entryNtl.toFixed(6).replace(/\.?0+$/, ''),
```
ravinagill15, line 37: nit: we can have a helper for the regex here

packages/hyperliquid-composer/src/io/cli-output-formatter.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    const table = new Table({
+        head: ['Coin', 'Token', 'Total', 'Hold', 'Entry Ntl'],
+        colAligns: ['left', 'right', 'right', 'right', 'right'],
+    })
+
+    let filteredBalances = balances.balances
+
+    if (!showZeroBalances) {
+        filteredBalances = balances.balances.filter((balance) => parseFloat(balance.total) > 0)
```
ravinagill15, line 18: should we return early if `filteredBalances` is empty?

ravinagill15 (commented):
Looks great! just some nits

packages/hyperliquid-composer/src/types/operations.ts
shankars99: https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint/spot#retrieve-a-users-token-balances uses `spotClearinghouseState`

packages/hyperliquid-composer/src/io/cli-output-formatter.ts
shankars99: their api's definition tells us that is it a number, if they make changes then several things will break. so its not really required

packages/hyperliquid-composer/src/io/cli-output-formatter.ts
shankars99: true but creating a function for it just increases the number of lines vs it being inlined, also imo makes it easier to read?

packages/hyperliquid-composer/src/io/cli-output-formatter.ts
shankars99: yeah we should

ravinagill15 (approved):

packages/hyperliquid-composer/src/commands/account-state.ts
```diff
@@ -0,0 +1,27 @@
+import { setDefaultLogLevel } from '@layerzerolabs/io-devtools'
+
+import { spotClearinghouseState } from '@/operations'
+import { SpotBalancesResponse } from '@/types'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export async function isAccountActivated(args: any): Promise<boolean> {
```
AlexanderLiteplo, line 7: Args should be unpacked into a defined type so that we know its contents

packages/hyperliquid-composer/src/operations/infoMisc.ts
```diff
@@ -0,0 +1,17 @@
...
+    isTestnet: boolean,
+    logLevel: string
+): Promise<SpotBalancesResponse> {
+    const action: SpotClearinghouseState['action'] = {
+        type: 'spotClearinghouseState',
+        user: user,
+    }
+
+    const hyperliquidClient = new HyperliquidClient(isTestnet, logLevel)
+    const response = await hyperliquidClient.submitHyperliquidAction('/info', null, action)
```
AlexanderLiteplo, line 15: string /info should be given a meaningful name

AlexanderLiteplo: also null should be given a name or this function should just have typed params

packages/hyperliquid-composer/src/cli.ts
```diff
@@ -26,6 +30,29 @@ program
...
+    .requiredOption('-u, --user <0x>', 'User address')
+    .requiredOption('-n, --network <network>', 'Network (mainnet/testnet)')
+    .option('-l, --log-level <level>', 'Log level', LogLevel.info)
+    .action(async (options) => {
+        const res = await isAccountActivated(options)
+        console.log(`Account activated: ${res}`)
+    })
+
+program
+    .command('get-core-balances')
```
AlexanderLiteplo, line 45: what is a core balance? can you add documentation around that or give it a better name?

packages/hyperliquid-composer/HYPERLIQUID.README.md
```diff
@@ -324,6 +324,25 @@ npx @layerzerolabs/hyperliquid-composer core-spot \
     [--log-level {info | verbose}]
 ```
 
+#### Get Core Spot balances
```
AlexanderLiteplo, line 327: a breif explanation on whata core spot balance is would be helpful

packages/hyperliquid-composer/src/commands/account-state.ts
```diff
@@ -0,0 +1,27 @@
...
+    setDefaultLogLevel(args.logLevel)
+
+    const user = args.user
+    const isTestnet = args.network === 'testnet'
+
+    const userBalances = await spotClearinghouseState(user, isTestnet, args.logLevel)
+
+    return userBalances.balances.length === 0 ? false : true
```
AlexanderLiteplo, line 15: are you sure balances will be defined?

packages/hyperliquid-composer/src/types/parser.ts
```diff
@@ -96,3 +96,15 @@ export interface SpotDeployStates {
     states: DeployState[]
     gasAuction: GasAuction
 }
+
+export interface SpotBalance {
```
shankars99, line 100: hyperliquid's return json lol

packages/hyperliquid-composer/src/cli.ts
shankars99: its in the readme

packages/hyperliquid-composer/HYPERLIQUID.README.md
shankars99: like you can scroll up theres a lot of places where it talks about it

packages/hyperliquid-composer/src/commands/account-state.ts
shankars99: yeah its the primary key

packages/hyperliquid-composer/src/operations/infoMisc.ts
shankars99: im following the hyperliquid python sdk's usage where they pass in "/info" instead of something like INFO which is literally just `/info`.

wdym "have typed params"

packages/hyperliquid-composer/src/commands/account-state.ts
shankars99: typing for every api call is unnecessary work

packages/hyperliquid-composer/src/operations/infoMisc.ts
AlexanderLiteplo: sorry not typed params I meant destructured parameters:

```typescript
function createUser({ name, email, age }: { name: string; email: string; age: number }) {
    return `User: ${name}, ${email}, ${age}`;
}

// Must use an object with named properties
createUser({ name: "John", email: "john@example.com", age: 25 });
```

packages/hyperliquid-composer/src/types/parser.ts
AlexanderLiteplo: can you add a comment explaining what they are?

shankars99: best i can do it link them to hyperliquid's api https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint which is also not too useful

packages/hyperliquid-composer/src/operations/infoMisc.ts
shankars99: im against this since there are only 3 arguments and are defined across all the places where its used, there are also no optional args and are few in number of args. 

the first is always an endpoint url 
the second is wallet or null 
the third is the action. 

having destructured args here is just overkill and makes stuff more unreadable

AlexanderLiteplo: How does that make it unreadable? This pattern is specifically for readability. You have a bunch of magic values right now which are unreadable.

packages/hyperliquid-composer/src/types/parser.ts
AlexanderLiteplo: why can you not add a one line explanation of each value? do you not know what they are?

packages/hyperliquid-composer/src/operations/infoMisc.ts
shankars99: its a bad pattern for:
1. fewer args 
2. no optional args

AlexanderLiteplo: '/info' should be name infoEndpointUrlSuffix or something. otherwise I look at this and have no idea what is going on

packages/hyperliquid-composer/src/types/parser.ts
shankars99: honestly no, i dont know what each of them means

packages/hyperliquid-composer/src/operations/infoMisc.ts
shankars99: disagree cuz its wordier

AlexanderLiteplo: Same with null. I look at this and cannot tell what is being set to null. Also passing in null for a wallet does not make sense unless you define the name. EMPTY_WALLET = null, or something like that. Then I understand why we are passing null and do not have to read docs of the code you are interacting with

AlexanderLiteplo: If that is too wordy then get creative and make a succint name. Wordier is better than no name at all

shankars99: > do not have to read docs 

sir?

packages/hyperliquid-composer/src/cli.ts
AlexanderLiteplo: where? none of the readmes in this pr define it

packages/hyperliquid-composer/src/operations/infoMisc.ts
shankars99: a name isn't a requirement for every variable. 

open to suggestions but idt its possible to get something under "/info" or "null"

packages/hyperliquid-composer/src/cli.ts
shankars99: https://github.com/LayerZero-Labs/devtools/blob/main/packages/hyperliquid-composer/HYPERLIQUID.README.md#token-standards

it being a token tells you that it has a balance. imo thats a reference anyone can make

AlexanderLiteplo (approved):

St0rmBr3w (approved):
LGTM!

PR #1639, opened 2025-07-29
ovault composer audit resolution - ottersec

shankars99:


packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -272,23 +272,23 @@ contract VaultComposerSync is IVaultComposerSync, ReentrancyGuard {
...
-     * @param _oft The OFT contract address to quote
+     * @param _targetOFT The OFT contract address to quote
      * @param _vaultInAmount The amount of tokens to send to the vault
      * @param _sendParam The parameters for the send operation
      * @return MessagingFee The estimated fee for the send operation
      * @dev This function can be overridden to implement custom quoting logic
      */
     function quoteSend(
-        address _oft,
+        address _targetOFT,
```
TRileySchwarz, line 283: Need to also update the interface param value for this corresponding change

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -272,23 +272,23 @@ contract VaultComposerSync is IVaultComposerSync, ReentrancyGuard {
...
         uint256 _vaultInAmount,
         SendParam memory _sendParam
     ) external view virtual returns (MessagingFee memory) {
-        if (_oft == ASSET_OFT) {
-            _sendParam.amountLD = VAULT.previewDeposit(_vaultInAmount);
-        } else {
+        if (_targetOFT == ASSET_OFT) {
```
TRileySchwarz, line 289: Would add a small comment something like this:

// @dev If we are quoting the asset OFT, that means we are inputting shares and the amountLD passed into quote should be denominated in assets (and vice versa)

TRileySchwarz: Split comment into multiple lines, too long for linting

PR #1644, opened 2025-08-03
fix getSendLibrary throwing in happy path

nazreen:
Previously, sendLibrary would not throw, but just return null, when the defaultSendLibrary was not found. Wiring was allowed to continue. This PR catches the specific new error and allows wiring to continue.

## Proof of Test

https://testnet.layerzeroscan.com/tx/2cmxP2GGyumksbVAism2QFouvLmoNhz9vpKh65fcij9uJ2LLCJ7AVy5muCyRCfVSxNk5ourYmSqw394tvK5xAAsS
https://testnet.layerzeroscan.com/tx/0xc07afa6cfad3b5b53def8dadfec4a7fd75b8c87cc44491bc8fbe5d4afdcf665c

### DIFF
--- a/packages/protocol-devtools-solana/src/endpointv2/sdk.ts
+++ b/packages/protocol-devtools-solana/src/endpointv2/sdk.ts
@@ -201,7 +201,21 @@ export class EndpointV2 extends OmniSDK implements IEndpointV2 {
         this.logger.debug(`Checking default send library for eid ${dstEid} (${eidLabel}) and address ${sender}`)
 
         const config = await mapError(
-            () => this.program.getSendLibrary(this.connection, new PublicKey(sender), dstEid),
+            async () => {
+                try {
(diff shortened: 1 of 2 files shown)


packages/protocol-devtools-solana/src/endpointv2/sdk.ts
```diff
@@ -548,6 +562,11 @@ export class EndpointV2 extends OmniSDK implements IEndpointV2 {
         throw new TypeError(`isRegisteredLibrary() not implemented on Solana Endpoint SDK`)
     }
 
+    @AsyncRetriable()
+    async isBlockedLibrary(): Promise<boolean> {
```
nazreen, line 566: @ziming-zung is this fine to include? it otherwise wouldn't build, complaining about how IEndpointV2 needs `isBlockedLibrary`

PR #1646, opened 2025-08-04
DEVREL-650 feat: ovault evm example setup

St0rmBr3w:
### In this PR:

- adds better handling and documentation for sending tokens to the vault

Screenshot 2025-08-04 at 12 49 04 PM


examples/ovault-evm/contracts/MyERC4626.sol
```diff
@@ -2,12 +2,14 @@
...
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { ERC4626 } from "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
 
-import { OVault } from "@layerzerolabs/ovault-evm/contracts/OVault.sol";
 import { OFTAdapter } from "@layerzerolabs/oft-evm/contracts/OFTAdapter.sol";
 
-contract MyOVault is OVault {
-    constructor(string memory name, string memory symbol, address asset) OVault(name, symbol, asset) {}
+contract MyERC4626 is ERC4626 {
```
shankars99, line 27: @St0rmBr3w what do you think is the best way to bubble up ERC4626 issues? like the inflation attack? 
ex: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol#L11-L47

examples/ovault-evm/contracts/MyERC4626.sol
```diff
@@ -2,12 +2,14 @@
...
+    constructor(string memory _name, string memory _symbol, IERC20 _asset) ERC20(_name, _symbol) ERC4626(_asset) {}
 }
 
 contract MyShareOFTAdapter is OFTAdapter {
```
shankars99, line 44: do you think we should keep this here? or in its own file?

shankars99:
```suggestion
/// @dev The share token MUST be an OFT adapter (lockbox style). 
/// @dev A mint-burn adapter would not work since it affects `ShareERC20::totalSupply()`
contract MyShareOFTAdapter is OFTAdapter {
```

examples/ovault-evm/contracts/MyOVaultComposer.sol
```diff
@@ -1,13 +1,12 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.22;
 
-import { OVaultComposer } from "@layerzerolabs/ovault-evm/contracts/OVaultComposer.sol";
+import { VaultComposerSync } from "@layerzerolabs/ovault-evm/contracts/VaultComposerSync.sol";
 
-contract MyOVaultComposer is OVaultComposer {
+contract MyOVaultComposer is VaultComposerSync {
```
shankars99, line 10:
```suggestion
contract MySyncVaultComposer is VaultComposerSync {
```

this changes `/deploy` -> update contract name

examples/ovault-evm/deploy/MyERC4626.ts
```diff
@@ -4,15 +4,13 @@ import { type DeployFunction } from 'hardhat-deploy/types'
 
 import { assetOFTContractName } from './MyAssetOFT'
 
-export const ovaultContractName = 'MyOVault'
-const tokenName = 'MyMockShareOFT'
-const tokenSymbol = 'MockShare'
+export const ovaultContractName = 'MyERC4626'
+const tokenName = 'MyShareOFT'
+const tokenSymbol = 'SHARE'
```
shankars99:
```suggestion
const export tokenName = 'MyShareOFT'
const export tokenSymbol = 'SHARE'
```

since `asset` and `share` are re-used across different files. i dont like the user having to re-define these things. what do you think about creating a new file here called `config.ts` where we add

`assetToken, shareToken, etc` as types `{name: string, symbol: string}` and pass the objects to these 3 deploy scripts?

examples/ovault-evm/deploy/MyShareOFT.ts
```diff
@@ -2,11 +2,11 @@ import assert from 'assert'
...
 import { type DeployFunction } from 'hardhat-deploy/types'
 
-import { ovaultContractName } from './MyOVault'
+import { ovaultContractName } from './MyERC4626'
 
 const shareOFTContractName = 'MyShareOFT'
-const tokenName = 'MyMockShareOFT'
-const tokenSymbol = 'MockShare'
+const tokenName = 'MyShareOFT'
+const tokenSymbol = 'SHARE'
```
shankars99:
```suggestion
import { tokenName, tokenSymbol } from './MyERC4626'
```

examples/ovault-evm/deploy/MyShareOFT.ts
```diff
@@ -2,11 +2,11 @@ import assert from 'assert'
...
 
 const deploy: DeployFunction = async (hre) => {
```
shankars99, line 9: todo: add docs on how this makes sure that youre not on the same network that has the `ERC4626`

shankars99 (commented):
todo: 
delete `ExecutorCompose302.s.sol` and update `SendScript.s.sol` 

 i have not gone through the tasks and readme. will do it in a bit

examples/ovault-evm/contracts/MyERC4626.sol
St0rmBr3w: We can leave a comment, but ultimately teams should understand the tokens they're implementing. I think best practice is NOT to reimplement the entire comment, but leave some security consideration, for example:

```solidity
/**
 * @title MyERC4626
 * @notice ERC4626 tokenized vault implementation for cross-chain vault operations
 * @dev SECURITY CONSIDERATIONS:
 *      - Donation/inflation attacks on empty or low-liquidity vaults
 *      - Share price manipulation via large donations before first deposit
 *      - Slippage during deposit/redeem operations in low-liquidity conditions  
 *      - First depositor advantage scenarios
 *      
 *      See OpenZeppelin ERC4626 documentation for full risk analysis:
 *      https://docs.openzeppelin.com/contracts/4.x/erc4626#inflation-attack
 *      
 *      MITIGATIONS:
 *      - OpenZeppelin v4.9+ includes virtual assets/shares to mitigate inflation attacks
 *      - Deployers should consider initial deposits to prevent manipulation
 *      
 * @author LayerZero Labs
 */
contract MyERC4626 is ERC4626 {
    /**
     * @notice Creates a new ERC4626 vault
     * @dev Initializes the vault with virtual assets/shares protection against inflation attacks
     * @param _name The name of the vault token
     * @param _symbol The symbol of the vault token  
     * @param _asset The underlying asset that the vault accepts
     */
    constructor(string memory _name, string memory _symbol, IERC20 _asset) ERC20(_name, _symbol) ERC4626(_asset) {}
}
```

examples/ovault-evm/contracts/MyOVaultComposer.sol
St0rmBr3w: does this name change matter?

examples/ovault-evm/deploy/MyShareOFT.ts
St0rmBr3w: @shankars99 we can just force them to define this in hardhat.config.ts the same way we do for OFTAdapter and add a check that way?

examples/ovault-evm/contracts/MyOVaultComposer.sol
shankars99: other vault composers that we put out may not use the same constructor interface no?

examples/ovault-evm/contracts/MyERC4626.sol
shankars99: yeah thats good

examples/ovault-evm/contracts/MyOVaultComposer.sol
St0rmBr3w: Wondering what best practice is in that case, I was thinking just keeping the term OVaultComposer consistent across implementations since they all play the same role, albeit via different impl.

examples/ovault-evm/type-extensions.ts
```diff
@@ -0,0 +1,39 @@
...
+     */
+    isHubChain: boolean
+    /**
+     * Asset token configuration (name and symbol)
+     */
+    assetToken: TokenConfig
+    /**
+     * Share token configuration (name and symbol)
+     */
+    shareToken: TokenConfig
```
shankars99: prolly shouldnt force both of them on each right? 
since the asset mesh may not matter - ex: they're using pre-deployed USDT but are deploying SHARE OFTs on all other networks.

St0rmBr3w: ah fair enough.

examples/ovault-evm/contracts/MyOVaultComposer.sol
shankars99: I think we can come back to this later too tbh

EWCunha (approved):

PR #1651, opened 2025-08-06
[INT-589] Fixing animation bug

AlexanderLiteplo:
Cause: unmount method needed to be called along with clear method. Previously only clear was needed due to the simplicity of the loading animation.

see slack for demo video of fix


packages/io-devtools/src/swag/printer.tsx
```diff
@@ -48,16 +48,11 @@ export const printRecord = (data: RecordData): void =>
  *
  * @see {@link printRecord}
  *
- * @param {RecordData[]} data Array of rows
+ * @param {TableRow} data Array of rows
  * @returns {void}
  */
-export const printRecords = (data: RecordData[]): void => {
-  const instance = render(<RecordList data={data} />, {
-    // Prevent clearing the terminal to preserve history
```
ItsAdel, line 56: seems like we want `debug:true` if it prevents clearing the terminal

packages/io-devtools/src/swag/printer.tsx
```diff
@@ -48,16 +48,11 @@ export const printRecord = (data: RecordData): void =>
  *
  * @see {@link printRecord}
  *
- * @param {RecordData[]} data Array of rows
+ * @param {TableRow} data Array of rows
```
ItsAdel, line 51: Should be RecordData[] still

packages/io-devtools/src/swag/components/penguin.tsx
```diff
@@ -72,33 +72,36 @@ export const PenguinSpinner: React.FC<PenguinSpinnerProps> = ({
...
     return () => {
-      mountedRef.current = false;
       if (intervalRef.current) {
         clearInterval(intervalRef.current);
         intervalRef.current = null;
       }
     };
+  }, [isActive]);
+
+  useEffect(() => {
```
ItsAdel: this code is running when component unmounts in the react lifecycle. Its state is about to be destroyed anyway so this is redundant. and could lead to warnings when attempting to change the state of a component which is unmounting

packages/io-devtools/src/swag/printer.tsx
AlexanderLiteplo: was causing other issues

packages/io-devtools/src/swag/printer.tsx
AlexanderLiteplo: reverting this change to its previous state (I made the change to recordData[])

ItsAdel (approved):
lgtm!

EWCunha (approved):
Awesome! You rock!

PR #1656, opened 2025-08-08
ovault - example improvements and deploy

shankars99:
Few changes to the deploy steps and state tracking `vaultChain`

1. hh config only marks if a network is a vaultChain 
2. validation against multiple vaultChains
3. remove replication of `ShareOFT` and `AssetOFT` against all networks. So now the relationship between the type of OFT and the OFT content (name, symbol, contractName) are pinned
4. create a framework that can be easily adapted to multiple vaults (you just change type-extensions from bool to string. like vaultName and then make a mapping keyed on the vaultName)
5. try fetching the contract from deployments if not provided by the user 
6. handle cases where the dev either is either deploying their own AssetOFT or using a pre-deployed one (in this case the only time they supply it is on the vaultChain)
7. allow the developer to ONLY deploy the vault if they want and resume from that state


examples/ovault-evm/deploy/tokenConfig.ts
St0rmBr3w, line 1: Not a fan of a separate token config file, imo less is more and they already need to manage:

- hardhat.config.ts
- layerzero.config.ts

Let's consolidate changes to either/or.

examples/ovault-evm/deploy/MyOVaultComposer.ts
```diff
@@ -0,0 +1,48 @@
...
+    // Validate that vault contracts are deployed only on hub chain
+    if (!isVaultChain(hre.network.config)) {
+        throw new Error(
+            `Vault contracts can only be deployed on hub chain. Network '${hre.network.name}' is configured as spoke chain (isHubChain: false). Please deploy on a hub chain.`
+        )
+    }
+
+    // If ONLY_VAULT=1 is passed then the share OFT adapter deployment is skipped
+    // That means we may not have a share OFT adapter to deploy to
+    if (onlyDeployVault()) return
```
St0rmBr3w: I don't see the need for this, anyone who's deploying the vault likely has to deploy the ShareOFTAdapter. The counterpoint is that if they need separate deployments, why not create a separate deployment file and contract for MyShareOFTAdapter?

shankars99: they do its just an additional feature kept behind a flag

examples/ovault-evm/deploy/tokenConfig.ts
shankars99: hardhat config shouldnt since its a hardhat networks file and lz config doesnt need all the additional information such as token name, etc. 

instead of hardcoding it in every deploy file its better to share it around

PR #1659, opened 2025-08-08
Refactor deploy scripts

TRileySchwarz:


examples/ovault-evm/devtools/deployConfig.ts
```diff
@@ -1,43 +1,39 @@
...
+export const DEPLOYMENT_CONFIG = {
+    // Hub chain configuration
+    hub: {
+        eid: EndpointId.ARBSEP_V2_TESTNET,
+        contracts: {
+            vault: 'MyERC4626',
+            shareAdapter: 'MyShareOFTAdapter',
+            composer: 'MyOVaultComposer',
+        },
+    },
```
St0rmBr3w: this uses different semantics than the other two fields.

hub, asset, oft, doesn't make sense.

Should change hub to ovault since the vault is implicitly:

- ShareOFTAdapter
- Composer
- ERC4626 Vault

shankars99: oh yeah i agree

examples/ovault-evm/devtools/deployConfig.ts
```diff
@@ -1,43 +1,37 @@
...
-const shareConfig: OFTConfigShare = {
-    oft: shareOFT,
-    networks: shareMesh,
-}
-
-export { assetConfig, shareConfig, oVaultConfig }
+export const DEPLOYMENT_CONFIG = {
+    // Hub chain configuration
+    hub: {
+        eid: EndpointId.ARBSEP_V2_TESTNET,
```
St0rmBr3w, line 6: this should be kept consistent with the other params.

Either change `eid` to `chains` and have it revert if `chains.length > 1` or replace `chains` with `eids`.

examples/ovault-evm/devtools/deployConfig.ts
```diff
@@ -1,43 +1,37 @@
...
-
-const shareConfig: OFTConfigShare = {
-    oft: shareOFT,
-    networks: shareMesh,
-}
-
-export { assetConfig, shareConfig, oVaultConfig }
+export const DEPLOYMENT_CONFIG = {
+    // Hub chain configuration
+    hub: {
```
St0rmBr3w: `hub` doesn't make sense in this context, since we're configuring contract meshes, so it should be the "contract" we're deploying.

In this case since the OVault system is `ShareOFTAdapter` + `ERC4626` + `Composer`, this should be renamed to `ovault`.

examples/ovault-evm/devtools/deployConfig.ts
```diff
@@ -1,43 +1,37 @@
...
+        eid: EndpointId.ARBSEP_V2_TESTNET,
+        contracts: {
+            vault: 'MyERC4626',
```
St0rmBr3w, line 8: this needs a metadata field to define the share token's `name` and `symbol`.

PR #1663, opened 2025-08-12
docs: update link to simple config page

nazreen:
new simple config section is created in docs
links in example READMEs need to be updated


nazreen:
@10xKelly @St0rmBr3w now that the docs PR has been merged, we should merge this in too. A redirect was applied, but best we update the path here too.

PR #1689, opened 2025-08-26
IP-1274 feat(oft-solana): conditional ata value

nazreen:
## Motivation

Read: https://docs.layerzero.network/v2/developers/solana/oft/account#conditional-msgvalue-for-ata-creation

## Changes
- no longer recommending `value` in enforced options when sending to Solana
- introduce conditional setting of value based on existence of Solana recipient's Associated Token Account existence AND  based on whether there's already enforcedOptions.value set 

## Proof of Tests

First send - https://testnet.layerzeroscan.com/tx/0x028261e4cf2de488e3f73facf35ab5bd7ad1b6718116fe23d2c7a6eb90b9032f
(has value)

Subsequent send - https://testnet.layerzeroscan.com/tx/0x5ae75e635504975b97520601374ee9b1fae915489f204cecd98b0730cb2cdbd9
(no value)

---

EDIT: Enhanced to support Token2022 token accounts

Token2022 Mint: https://solscan.io/account/Dn6hECYyQQmeA7ZuWVQcYFdD5gjS1FEeKhpxu7YyohnL?cluster=devnet
Created with Permanent Delegate extension.
Created OFT in MABA mode.
Example token account: 4RzxArQWnhfAmS4j6Wv5vUvaAs3VLznqRXyCFFUCiSU4
Token account size: 170 bytes

## Proof of Tests

Test recipient: 8yToDJqkD31aikucY1sPvCmKFAPQt6ZVZLyyUUH6M36N

First send to test recipient: https://testnet.layerzeroscan.com/tx/0xe5ce462c7f40145ab41378f02f6fc58c4b2e781e27878da73a9f6b40f242226e
**Value provided**

Second send to test recipient: https://testnet.layerzeroscan.com/tx/0x50831f7116589f037a4e08a39d11343795151cdeef73fe1c6bc8fe8c3de3269b
**No value provided**

### DIFF
--- a/examples/onft721-zksync/hardhat.config.ts
+++ b/examples/onft721-zksync/hardhat.config.ts
@@ -49,6 +49,7 @@ const config: HardhatUserConfig = {
                 version: '0.8.24',
                 eraVersion: '1.0.0', //optional. Compile contracts with EraVM compiler
                 settings: {
+                    evmVersion: 'cancun',
                     optimizer: {
                         enabled: true,
                         runs: 200,
--- a/examples/onft721/hardhat.config.ts
+++ b/examples/onft721/hardhat.config.ts
@@ -45,6 +45,7 @@ const config: HardhatUserConfig = {
             {
                 version: '0.8.24',
                 settings: {
+                    evmVersion: 'cancun',
                     optimizer: {
                         enabled: true,
                         runs: 200,


examples/oft-solana/tasks/common/sendOFT.ts
```diff
@@ -65,13 +69,36 @@ task('lz:oft:send', 'Sends OFT tokens cross‐chain from any supported chain')
...
+            conditionalValue = await getConditionalValueForSendToSolana({
+                eid: args.dstEid,
+                recipient: args.to,
+                mint: solanaDeployment.mint,
+            })
+        }
+
+        // throw if user specified extraOptions and conditionalValue is non-zero
+        if (args.extraOptions && conditionalValue > 0) {
+            throw new Error('extraOptions and conditionalValue cannot be set at the same time')
```
St0rmBr3w: Why don't you just sum the extraOptions with the options determined by the ATA?

Then you can logger.caution that dstEid Solana uses conditional logic already for extraOptions?

nazreen:
Converted to draft. We need to coordinate with Stargate as currently OFTs on Stargate are required to have enforcedOptions.

examples/oft-solana/tasks/common/sendOFT.ts
nazreen: Actually, how often do we expect that devs will pass extraOptions strings as a param?

nazreen:
Note: I will be amending this PR to first check enforcedOptions before adding value

nazreen:
Changes from original implementation

- add check for enforcedOptions
- `sendOFT.ts` now no longer responsible for attaching the value, but only helping evaluate what minimum value is needed lzReceive if/when sending to Solana
- make the sendOFT backwards compatible (also works if enfocedOptions.value is set for send to Solana)
- removed the comment `"// Setting value both in the SOLANA_ENFORCED_OPTIONS and in the conditionalValue block below will result in redundant value being sent"` as now the logic ensures no redundant value is added if enforced options or extra options already include the minimum value
- before: throw if extraOptions and conditionalValue are both defined. now: no throwing. logic will help determine whether additional value is needed after combining enforcedOptions and extraOptions (if provided).
- `conditionalValue` renamed to `minimumLzReceiveValue`
- `getConditionalValueForSendToSolana` renamed to `getMinimumValueForSendToSolana`

nazreen:
note: I'll amend the PR with one last change: get token account size for Token2022 dynamically. will do this tomorrow.

EDIT: this is done

nazreen:
Currently blocked. We'll only merge after Stargate drops the requirement of having enforcedOptions.value for sends to Solana

nazreen:
Note:

the solidity version in the 2 ONFT examples were bumped from 0.8.22 to 0.8.24 due to the following:
In OZ v5.6.0+, ERC721.sol, Strings.sol, and Bytes.sol all bumped their pragma to ^0.8.24 (Strings now imports Bytes, and ERC721 imports Strings). The CI creates fresh projects without a lockfile, so ^5.0.2 resolves to 5.6.1.
Only onft721 is affected because ERC721.sol bumped to ^0.8.24 while ERC20.sol stayed at ^0.8.20 — so the OFT examples (which use ERC20, not ERC721) still compile fine with 0.8.22.

PR #1690, opened 2025-08-27
DEVREL-531 rename retryPayload to retryMessage

nazreen:
## Changes
- renamed retryPayload to retryMessage
 - also: changed `computeUnits` to `withComputeUnitLimit` and `withPriorityFee` to `withComputeUnitPrice` to align the param names to how they are in the `solana` CLI
 
 ## Proof of Test
 
 https://solscan.io/tx/5QpyGaUtsLGrbZAtuQgYwZhwDrLrjLJNXMLh4STth2yytMSjZhKejXrzt8vNQa5UdQ9uXnhvCZ2TowmTeJhNNrbK?cluster=devnet

### DIFF
--- a/examples/oft-solana/layerzero.config.ts
+++ b/examples/oft-solana/layerzero.config.ts
@@ -7,9 +7,9 @@ import { getOftStoreAddress } from './tasks/solana'
 
 // Note:  Do not use address for EVM OmniPointHardhat contracts.  Contracts are loaded using hardhat-deploy.
 // If you do use an address, ensure artifacts exists.
-const sepoliaContract: OmniPointHardhat = {
-    eid: EndpointId.SEPOLIA_V2_TESTNET,
-    contractName: 'MyOFT',
+const arbitrumContract: OmniPointHardhat = {
+    eid: EndpointId.ARBSEP_V2_TESTNET,
+    contractName: 'MyOFT', // Note: change this to your production contract name
 }
 
 const solanaContract: OmniPointHardhat = {
@@ -51,7 +51,7 @@ export default async function () {
     // if you declare A,B there's no need to declare B,A
     const connections = await generateConnectionsConfig([
         [
-            sepoliaContract, // Chain A contract
+            arbitrumContract, // Chain A contract
             solanaContract, // Chain B contract
             [['LayerZero Labs'], []], // [ requiredDVN[], [ optionalDVN[], threshold ] ]
             [15, 32], // [A to B confirmations, B to A confirmations]
@@ -60,7 +60,7 @@ export default async function () {
     ])
 
     return {
-        contracts: [{ contract: sepoliaContract }, { contract: solanaContract }],
+        contracts: [{ contract: arbitrumContract }, { contract: solanaContract }],
         connections,
     }
 }
--- a/.changeset/gorgeous-steaks-press.md
+++ b/.changeset/gorgeous-steaks-press.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/oft-solana-example": patch
+---
+
+fix oft-solana's debug script when peers are checked
--- a/.changeset/great-mugs-glow.md
+++ b/.changeset/great-mugs-glow.md
@@ -1,0 +1,24 @@
+---
+"@layerzerolabs/oft-adapter-aptos-move-example": patch
+"@layerzerolabs/mint-burn-oft-adapter-example": patch
+"@layerzerolabs/native-oft-adapter-example": patch
+"@layerzerolabs/oft-adapter-initia-example": patch
+"@layerzerolabs/oapp-aptos-example": patch
+"@layerzerolabs/oft-hyperliquid-example": patch
+"@layerzerolabs/oft-upgradeable-example": patch
+"@layerzerolabs/oft-aptos-move-example": patch
+"@layerzerolabs/view-pure-read-example": patch
+"@layerzerolabs/uniswap-read-example": patch
+"@layerzerolabs/oapp-solana-example": patch
+"@layerzerolabs/oft-adapter-example": patch
+"@layerzerolabs/oft-initia-example": patch
+"@layerzerolabs/oapp-read-example": patch
+"@layerzerolabs/omni-call-example": patch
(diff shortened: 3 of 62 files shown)


examples/oft-solana/tasks/solana/retryPayload.ts
```diff
@@ -68,12 +68,9 @@ task('lz:oft:solana:retry-payload', 'Retry a stored payload on Solana')
                     nonce: nonce.toString(),
                     srcEid,
                     sender: makeBytes32(sender),
-                    dstEid,
```
nazreen, line 71: context on why dstEid is also deleted

the type for lzReceive's third param:

```
packet: Pick<Packet, "message" | "sender" | "srcEid" | "guid" | "receiver" | "nonce">,
```

dstEid is not needed

examples/oft-solana/tasks/solana/retryPayload.ts
```diff
@@ -68,12 +68,9 @@ task('lz:oft:solana:retry-payload', 'Retry a stored payload on Solana')
...
                     receiver,
-                    payload: '', // unused;  just added to satisfy typing
```
nazreen, line 73: there are no more typing errors/warnings

examples/oft-solana/layerzero.config.ts
```diff
@@ -40,7 +40,8 @@ const SOLANA_ENFORCED_OPTIONS: OAppEnforcedOption[] = [
         msgType: 1,
         optionType: ExecutorOptionType.LZ_RECEIVE,
         gas: CU_LIMIT,
-        value: SPL_TOKEN_ACCOUNT_RENT_VALUE,
+        // value: SPL_TOKEN_ACCOUNT_RENT_VALUE,
+        value: 0,
```
ravinagill15: I think we should update the `Elaboration on `value` when sending OFTs to Solana:` comment above to indicate why we have set value to 0 here (which I believe is that it is only needed when sending to an uninitialized account for the first time, and all subsequent calls would unnecessarily send funds)

examples/oft-solana/tasks/solana/retryPayload.ts
```diff
@@ -43,15 +44,11 @@ task('lz:oft:solana:retry-payload', 'Retry a stored payload on Solana')
...
-            computeUnits,
+            message,
             lamports,
-            withPriorityFee,
+            withComputeUnitLimit,
+            withComputeUnitPrice,
         }: Args) => {
-            if (!process.env.SOLANA_PRIVATE_KEY) {
-                throw new Error('SOLANA_PRIVATE_KEY is not defined in the environment variables.')
-            }
```
ravinagill15, line 53: We might need this since this env variable is [being referenced further down in this task](https://github.com/LayerZero-Labs/devtools/blob/6bd1cfc1edec48543f3b18586fd72b2056fdfa26/examples/oft-solana/tasks/solana/retryMessage.ts#L86)

examples/oft-solana/tasks/solana/retryMessage.ts
ravinagill15, line 1: Should we also update the `retryPayload` tasks within the oapp and lzapp-migration examples?

ravinagill15: Not necessary for the current PR, but in a future PR I think it would be best to move these tasks to the devtools-solana package to reduce redundancy and minimize code that needs to be updated when changes like this occur

ravinagill15 (commented):
Overall looks good, left a few thoughts/questions

Let's also add a changeset

examples/oft-solana/layerzero.config.ts
nazreen: @ravinagill15 thanks for the spot. I need to undo this message as it's unintended change. This PR should not be modifying anything in the LZ config. This change was there as I needed to simulate a message that would pass verification but can't be executed immediately (in this case, due to insufficient value)

examples/oft-solana/tasks/solana/retryMessage.ts
nazreen: > but in a future PR I think it would be best to move these tasks to the devtools-solana package to reduce redundancy and minimize code

@ravinagill15 agreed. I think we'll be doing that in the new devtools though. This was just a quick patch to fix a misnomer.

examples/oft-solana/tasks/solana/retryPayload.ts
nazreen: thanks for the spot again! 

There should not be any direct references to env vars related to Solana keypairs/private keys as it should be handled by 

https://github.com/LayerZero-Labs/devtools/blob/main/examples/oft-solana/tasks/solana/index.ts#L48C14-L49

I've pushed a commit to ensure that the right signer variable is used now: https://github.com/LayerZero-Labs/devtools/pull/1690/commits/789f0157be39e41094d6c41d013eb6b6346e9a0c

nazreen: Proof of Test after the change: https://solscan.io/tx/4AP94EUeiN8eqojcjVkd6J9BGhgFAJzUWmEk9naWiffJZUnYQrziAwq1ZnY2wJVJJqXu8CYB8Sg9bqZ2szuqFJMd?cluster=devnet

examples/oft-solana/tasks/solana/retryMessage.ts
nazreen: > Should we also update the retryPayload tasks within the oapp and lzapp-migration examples?

Done, @ravinagill15 , thank you!

.changeset/gorgeous-kings-grow.md
ravinagill15, line 1: How come we have 2 changesets? I think we can mention all updated packages in a single changeset

ravinagill15:
Also just noticed the failing builds, let's ensure they pass before merging

.changeset/gorgeous-kings-grow.md
nazreen: I don't think this matters. The 2 changesets cover diff packages and do not repeat. 

But to answer your question, there are 2 because the first one was made when only the script in oft-solana was updated.

DanL0 (approved):

PR #1714, opened 2025-09-09
feat: generate and verify lockfiles for example packages

tinom9:
### Description

- Generate and verify `pnpm` V6 lockfiles for example packages installed with `create-lz-oapp`.
  - `pnpm` installation for versions 8, 9, and 10 is available through generated lockfiles:
    - Versions 8 and 9 use generated `pnpm` V6 lockfiles.
    - Version 10 needs to use `pnpm` V9 lockfiles, so the installation process converts the generated V6 lockfile into V9 and then installs it. This has been chosen instead of maintaining both V6 and V9 lockfiles in the repo.
  - `pnpm` installation is still available without lockfile support, which should be considered unsafe.
  - `npm` and `bun` installations should be considered unsafe as well as no lockfiles are supported yet for them.

### DIFF
--- a/packages/oft-evm-upgradeable/CHANGELOG.md
+++ b/packages/oft-evm-upgradeable/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @layerzerolabs/oft-evm-upgradeable
 
+## 3.2.0
+
+### Minor Changes
+
+- 7af2d6e: Adding NativeOFTAdapterUpgradeable
+
 ## 3.1.0
 
 ### Minor Changes
--- a/packages/oft-evm-upgradeable/contracts/oft/NativeOFTAdapterUpgradeable.sol
+++ b/packages/oft-evm-upgradeable/contracts/oft/NativeOFTAdapterUpgradeable.sol
@@ -1,0 +1,141 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.22;
+
+import { MessagingFee, MessagingReceipt, OFTCoreUpgradeable, OFTFeeDetail, OFTLimit, OFTReceipt, SendParam } from "./OFTCoreUpgradeable.sol";
+
+/**
+ * @title NativeOFTAdapterUpgradeable Contract
+ * @dev NativeOFTAdapterUpgradeable is a contract that adapts native currency to the OFT functionality.
+ *
+ * @dev WARNING: ONLY 1 of these should exist for a given global mesh,
+ * unless you make a NON-default implementation of OFT, which needs to be done very carefully.
+ * @dev WARNING: The default NativeOFTAdapterUpgradeable implementation assumes LOSSLESS transfers, ie. 1 native in, 1 native out.
+ */
+abstract contract NativeOFTAdapterUpgradeable is OFTCoreUpgradeable {
+
+    error IncorrectMessageValue(uint256 provided, uint256 required);
+    error CreditFailed(address to, uint256 amountLD, bytes revertData);
+
+    /**
+     * @dev Constructor for the NativeOFTAdapterUpgradeable contract.
+     * @param _localDecimals The decimals of the native on the local chain (this chain). 18 on ETH.
+     * @param _lzEndpoint The LayerZero endpoint address.
+     */
+    constructor(
+        uint8 _localDecimals,
+        address _lzEndpoint
+    ) OFTCoreUpgradeable(_localDecimals, _lzEndpoint) {}
+
+    /**
+     * @dev Initializes the NativeOFTAdapterUpgradeable with the provided delegate.
+     * @param _delegate The delegate capable of making OApp configurations inside of the endpoint.
+     *
+     * @dev The delegate typically should be set as the owner of the contract.
+     * @dev Ownable is not initialized here on purpose. It should be initialized in the child contract to
+     * accommodate the different version of Ownable.
+     */
+    function __NativeOFTAdapter_init(address _delegate) internal onlyInitializing {
+        __OFTCore_init(_delegate);
+    }
+
+    function __NativeOFTAdapter_init_unchained() internal onlyInitializing {}
+
+    /**
+     * @dev Returns the address of the native token
+     * @return The address of the native token.
+     */
+    function token() public pure returns (address) {
+        return address(0);
+    }
+
+    /**
+     * @notice Indicates whether the OFT contract requires approval of the 'token()' to send.
+     * @return bool indicating whether approval of underlying token implementation is required.
+     *
+     * @dev In the case of default NativeOFTAdapter, approval is not required.
+     */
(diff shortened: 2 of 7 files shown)


packages/create-lz-oapp/src/config.ts
```diff
@@ -242,23 +242,43 @@ export const getExamples = (): Example[] => {
...
-        label: 'npm',
+        label: 'npm (unsafe / no lockfile)',
+        hasLockfile: false,
     },
     {
         id: 'bun',
         executable: 'bun',
         args: ['install'],
-        label: 'bun',
+        label: 'bun (unsafe / no lockfile',
```
shankars99: nit: missing `)`

```suggestion
        label: 'bun (unsafe / no lockfile)',
```

packages/create-lz-oapp/src/config.ts
```diff
@@ -242,23 +242,43 @@ export const getExamples = (): Example[] => {
...
 
 const PACKAGE_MANAGERS: PackageManager[] = [
+    {
+        // pnpm 10+ needs to update pnpm lockfile from V6 to V9.
+        id: 'pnpm_lockfile',
+        executable: 'pnpm',
+        commands: ['pnpm install --lockfile-only --ignore-scripts', 'pnpm install --frozen-lockfile'],
+        label: 'pnpm (recommended)',
+        hasLockfile: true,
+        versionRegex: /^(1[0-9]|[2-9][0-9])\./,
```
shankars99, line 252: do we want to go that deep on version 10.xyz? 

would just `1[0-9]` be fine? also this goes until pnpm 19? shouldnt we limit this boundary to be within a tested version which rn is 10? and as this increases we can change it?

packages/create-lz-oapp/src/config.ts
tinom9: Will change.

packages/create-lz-oapp/src/config.ts
tinom9: This goes 99.

The idea is that the package manager will be the one that breaks support and raises the error, same behaviour than before. So that we don't need to explicitly add 11 if it doesn't break support.

Happy to change it if you have a strong opinion.

shankars99: sounds good

packages/create-lz-oapp/src/config.ts
```diff
@@ -242,23 +242,43 @@ export const getExamples = (): Example[] => {
...
-        label: 'pnpm (recommended)',
+        label: 'pnpm (unsafe / no lockfile)',
+        hasLockfile: false,
     },
     {
         id: 'npm',
         executable: 'npm',
         args: ['install'],
-        label: 'npm',
+        label: 'npm (unsafe / no lockfile)',
```
shankars99, line 273: should we throw a warning icon on this message? ⚠️ this one?

packages/create-lz-oapp/src/utilities/cloning.ts
```diff
@@ -79,14 +79,20 @@ export const cloneExample = async ({ example, destination }: Config) => {
 // List of files to be removed after the cloning is done
 const IGNORED_FILES = ['CHANGELOG.md', 'turbo.json']
 
+// Lockfiles from diferent package managers should be removed
+const LOCKFILES: { [lockfile: string]: PackageManager['id'][] } = {
+    'pnpm-lock.yaml': ['pnpm_lockfile', 'pnpm_lockfile_8_or_9'],
```
shankars99, line 84: should we list `npm` and `bun` on this such that if the user tries installing those we remove the other lockfiles too (`pnpm-lock`)? or i might be reading this function wrong.

also its not part the code change but throwing a console logger in line 101 could be helpful

packages/create-lz-oapp/src/config.ts
tinom9: I added a warning icon after the install happens.

Saying `unsafe` there to me it's a good indicative already of that it's better to go with the recommended way in that step, and then we say that they must verify the packages after the install is done.

packages/create-lz-oapp/src/utilities/cloning.ts
tinom9: This should be removing the lockfiles if the package manager ID is not in the list.

Therefore:
- Lockfiles should be kept for both `pnpm` versions that have a lockfile.
- Lockfiles should be deleted for `pnpm` version without lockfile, `npm`, and `bun`.

packages/create-lz-oapp/src/utilities/cloning.ts
```diff
@@ -79,14 +79,20 @@ export const cloneExample = async ({ example, destination }: Config) => {
 // List of files to be removed after the cloning is done
 const IGNORED_FILES = ['CHANGELOG.md', 'turbo.json']
 
+// Lockfiles from diferent package managers should be removed
```
ravinagill15: nit: typo

```suggestion
// Lockfiles from different package managers should be removed
```

packages/create-lz-oapp/src/config.ts
```diff
@@ -242,23 +242,43 @@ export const getExamples = (): Example[] => {
 }
 
 const PACKAGE_MANAGERS: PackageManager[] = [
+    {
+        // pnpm 10+ needs to update pnpm lockfile from V6 to V9.
+        id: 'pnpm_lockfile',
+        executable: 'pnpm',
+        commands: ['pnpm install --lockfile-only --ignore-scripts', 'pnpm install --frozen-lockfile'],
```
ihor-romanchenko-lzl, line 249: Why do we need 2 separate commands here? Why not a single `pnpm install` ?

package.json
```diff
@@ -8,6 +8,8 @@
     "dev": "$npm_execpath turbo run dev",
     "lint": "$npm_execpath turbo run lint",
     "lint:fix": "$npm_execpath turbo run lint:fix",
+    "lockfiles:generate": "node scripts/generate-lockfiles.mjs",
```
ihor-romanchenko-lzl, line 11: Why do we commi the lockfiles if we can generate them?

packages/create-lz-oapp/src/config.ts
tinom9: Changed.

packages/create-lz-oapp/src/config.ts
tinom9: First to migrate V6 lockfile into V9, second one to install.

packages/create-lz-oapp/src/utilities/cloning.ts
tinom9: Good eye, changed!

package.json
tinom9: To have version control, etc...

PR #1719, opened 2025-09-14
fix: composeMsg in debug send script

shankars99:
composeMsg is `abi.encode(SendParam, uint256 minMsgValue)` where `minMsgValue` is the `composeValue`


shankars99:
have a few more changes to add in docs wise and a new command to enable pairs against other quote tokens

DanL0 (approved):

PR #1723, opened 2025-09-17
chore: events for tracking deposit and redeem

shankars99:
Adding in `Deposited` and `Redeemed` events that emit:
 - Sender as `bytes32` (`srcChain.sender` if `lzCompose` or `hubChain.sender` if `proxyAndSend`)
 - Vault in amt (asset/share)
 - Recipient as `bytes32` (`srcChain.sender` if `lzCompose` or `hubChain.sender` if `proxyAndSend`)
 - Vault out amt (share/asset)

Events are emitted in `_depositAndSend(...args)` and `_redeemAndSend(...args)` which are used by `lzCompose Txs` AND `proxyAndSend Txs`

Todo: add artifacts and changeset


packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -247,6 +248,7 @@ contract VaultComposerSync is IVaultComposerSync, ReentrancyGuard {
         _sendParam.minAmountLD = 0;
 
         _send(ASSET_OFT, _sendParam, _refundAddress);
+        emit Redeemed(_redeemer, _shareAmount, _sendParam.to, assetAmount, _sendParam.dstEid);
```
0xsiddharthks: nit: maybe lets have the asset amount and share amount be swapped in positions to be consistent in both events?

shankars99:
```
    event Deposited(bytes32 sender, uint256 assetAmt, bytes32 recipient, uint256 shareAmt, uint32 dstEid); // 0x1267eb8b
    event Redeemed(bytes32 sender, uint256 shareAmt, bytes32 recipient, uint256 assetAmt, uint32 dstEid); // 0x53b1e2cb
```

i defined `Deposited` and `Redeemed` like this `(sender,vaultInToken,receiver,vaultOutToken,dstEid)` on Redeem the `vaultInToken` is the `share` and the `vaultOutToken` is the `asset`.

i can make the two uniform like `sender,assetAmt,receiver,shareAmt,dstEid` too

0xsiddharthks (approved):
LGTM! `dstEid` should allow us to differentiate between `ABB` and `BBA`

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -11,10 +11,13 @@ interface IVaultComposerSync is IOAppComposer {
...
     event Refunded(bytes32 indexed guid); // 0xfe509803
 
+    event Deposited(bytes32 sender, bytes32 recipient, uint256 assetAmt, uint256 shareAmt, uint32 dstEid); // 0x65f3a003
+    event Redeemed(bytes32 sender, bytes32 recipient, uint256 shareAmt, uint256 assetAmt, uint32 dstEid); // 0x53b1e2cb
+
     /// ========================== Error Messages =====================================
     error ShareOFTNotAdapter(address shareOFT); // 0xfc1514ae
     error ShareTokenNotVault(address shareERC20, address vault); // 0x0e178ab6
-    error AssetTokenNotVaultAsset(address assetERC20, address vaultAsset); // 0xba9d665f
+    error AssetTokenNotVaultAsset(address assetERC20, address vaultAsset); // 0x338bea04
```
tinom9:
```suggestion
    error AssetTokenNotVaultAsset(address assetERC20, address vaultAsset); // 0xba9d665f
```

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -11,10 +11,13 @@ interface IVaultComposerSync is IOAppComposer {
     event Sent(bytes32 indexed guid); // 0x27b5aea9
     event Refunded(bytes32 indexed guid); // 0xfe509803
 
+    event Deposited(bytes32 sender, bytes32 recipient, uint256 assetAmt, uint256 shareAmt, uint32 dstEid); // 0x65f3a003
+    event Redeemed(bytes32 sender, bytes32 recipient, uint256 shareAmt, uint256 assetAmt, uint32 dstEid); // 0x53b1e2cb
```
tinom9:
```suggestion
    event Redeemed(bytes32 sender, bytes32 recipient, uint256 shareAmt, uint256 assetAmt, uint32 dstEid); // 0x338bea04
```

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -247,6 +248,7 @@ contract VaultComposerSync is IVaultComposerSync, ReentrancyGuard {
         _sendParam.minAmountLD = 0;
```
tinom9, line 248: I know in practice this is not going to be a problem (composers will be deployed on trustable OFTs), and that this is not relevant in this PR, but given that we receive `_sendParam` from the final user, I don't see the harm on letting `_sendParam.minAmountLD` remain a user-input value.

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
```diff
@@ -11,10 +11,13 @@ interface IVaultComposerSync is IOAppComposer {
     event Sent(bytes32 indexed guid); // 0x27b5aea9
     event Refunded(bytes32 indexed guid); // 0xfe509803
 
+    event Deposited(bytes32 sender, bytes32 recipient, uint256 assetAmt, uint256 shareAmt, uint32 dstEid); // 0x65f3a003
```
tinom9: Supernit, but I'd consider grouping `recipient` and `dstEid` together, as they both represent the same thing: a GUID of the recipient. Both for `Deposited` and `Redeemed`

tinom9 (changes requested):

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: updated in https://github.com/LayerZero-Labs/devtools/pull/1723/commits/60233b61f04da2dccdd81d6d57606d00a7a5c375

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: updated in https://github.com/LayerZero-Labs/devtools/pull/1723/commits/60233b61f04da2dccdd81d6d57606d00a7a5c375

packages/ovault-evm/contracts/interfaces/IVaultComposerSync.sol
shankars99: that does make sense

shankars99: updated in https://github.com/LayerZero-Labs/devtools/pull/1723/commits/8130219671c909070b3949f58a4cf14b90a6ffb4

cc @0xsiddharthks

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99: we have a slippage check 2 lines above that. the reason its in is cuz the lets call the amt that you get out of the vault `rawAmtLd` and now when you pass that into `OFT.send(..args)` the `debit(...args)` takes in `rawAmtLd` and `_sendParam.minAmtLd` and then:

1. `dustlessAmtLd` = removeDust(`rawAmtLd`)
2. `if dustlessAmtLd < minAmtLd => revert`

The issue here is that the front end or user may have performed a simple `previewDeposit/Redeem` and just plugged that value into the `minAmtLd` without accounting for dust. This means that `minAmtLd` _could_ be `dustlessAmtLd + dust` and will always revert.  

While we _could_ update the `_sendParam.minAmtLd` to be the dustless version by doing a bunch of contract calls, its a simple change by setting to 0 and having our own slippageCheck that we can work around a `dusty-sendParam.minAmtLd`


this is in the function's `natspec @notice` btw

tinom9: I agree on that the most important use-case of the field is the slippage from the vault redemption/deposit.

I see that the frontend has to make a more complex operation (first quote vault and then quote OFT), but this implementation effectively disables the functionality of `minAmountLD`. I think we should aim for consistency of `minAmountLD` usage and implications, especially if there is no risk.

It's very subjective though and, again, not a change from this PR.

0xsiddharthks (approved):

PR #1725, opened 2025-09-18
feat: hyperliquid register token sdk

shankars99:
The hyperliquid SDK now checks deployed tokens and then prompts the user for entering their quote tokens (USDC, USDT, etc). This checks the user's current deployments and prompts them for first-class tokens (according to me) that we support which are USDC and USDT. 

The user can also type in any quote token index. 

Deploying spotMarkets for tokens after your initial deployment costs about `500 HYPE` right now which is the gas auction.

I have therefore created 2 new sdk operations to:
1. check the current auction cost - `spot-auction-status`
2. list deployments for a given token - `list-spot-pairs`


packages/hyperliquid-composer/src/commands/core-spot-deployment.ts
```diff
@@ -95,10 +104,10 @@ export async function coreSpotDeployment(args: any): Promise<void> {
         const decimals = await contract.decimals()
         logger.verbose(`Token name: ${tokenName}, Decimals: ${decimals}`)
 
-        const weiDiff = decimals - coreSpot.weiDecimals
+        const weiDiff = Number(decimals) - coreSpot.weiDecimals
```
lzErfan: why is it `Number(decimals)` and `parseInt(tokenIndex)`

packages/hyperliquid-composer/src/types/cli-args.ts
```diff
@@ -0,0 +1,80 @@
...
+export interface PrivateKeyArgs extends BaseArgs {
+    privateKey?: string
+}
+
+export interface UserArgs extends BaseArgs {
+    user: string
+}
+
+export interface OAppArgs extends BaseArgs {
+    oappConfig?: string
```
lzErfan, line 27: If your OAppArgs doesn't have the oappConfig string, is it still an OAppArg?

shankars99: renamed to OptionalOAppArgs cuz its used in a few places

packages/hyperliquid-composer/src/commands/core-spot-deployment.ts
shankars99: using parseInt to be consistent

packages/hyperliquid-composer/src/operations/spotMeta.ts
```diff
@@ -47,3 +56,107 @@ export async function getSpotDeployState(deployerAddres: string, isTestnet: bool
...
+ * @returns Array of spot pairs that include the specified token
+ */
+export async function getSpotPairs(isTestnet: boolean, tokenIndex: number, logLevel: string): Promise<SpotPair[]> {
+    const action: BaseInfoRequest = {
+        type: 'spotMeta',
+    }
+
+    const hyperliquidClient = new HyperliquidClient(isTestnet, logLevel)
+    const response = await hyperliquidClient.submitHyperliquidAction('/info', null, action)
+    const spotMetaUniverse = response as SpotMetaUniverse
```
tinom9, line 77: Super nit.

You could pass `SpotMetaUniverse` as a type arg to `hyperliquidClient.submitHyperliquidAction`.

packages/hyperliquid-composer/src/operations/spotDeploy.ts
```diff
@@ -230,11 +230,153 @@ export async function setNoHyperliquidity(wallet: Wallet, isTestnet: boolean, to
...
+    choices.push({
+        name: 'Quit (cancel operation)',
+        value: -2, // Special value to indicate quit
+    })
+
+    // Check if all standard quote tokens are already deployed
+    const allStandardTokensDeployed =
+        existingQuoteTokens.includes(usdc_tokenId) && existingQuoteTokens.includes(usdt0_tokenId)
+
+    if (allStandardTokensDeployed && choices.length === 1) {
```
tinom9: `choices.length >= 2` is an invariant?

packages/hyperliquid-composer/src/operations/spotDeploy.ts
```diff
@@ -230,11 +230,153 @@ export async function setNoHyperliquidity(wallet: Wallet, isTestnet: boolean, to
...
+        quoteTokenName = `Token ${quoteTokenId}`
+    } else {
+        quoteTokenId = selectedQuoteToken
+        if (quoteTokenId === usdc_tokenId) {
+            quoteTokenName = 'USDC'
+        } else if (quoteTokenId === usdt0_tokenId) {
+            quoteTokenName = 'USDT0'
+        } else {
+            quoteTokenName = `Token ${quoteTokenId}`
+        }
```
tinom9, line 370: This is never reached?

packages/hyperliquid-composer/src/operations/spotMeta.ts
shankars99: uhhh that would be a lot of Generic typing no? that would be a lot of refactor

packages/hyperliquid-composer/src/operations/spotDeploy.ts
shankars99: oh true

packages/hyperliquid-composer/src/operations/spotDeploy.ts
shankars99: 4 options - 2 tokens, idk if we are gonna support more "standard" tokens in the future. this selection of "standard tokens" is a layerzero thing and not a hyperliquid protocol thing

shankars99: updaetd choices.length to 2 in https://github.com/LayerZero-Labs/devtools/pull/1725/commits/560e0817b511ce7921a7c32717b8cce5470f894d

packages/hyperliquid-composer/src/operations/spotDeploy.ts
shankars99: updated in https://github.com/LayerZero-Labs/devtools/pull/1725/commits/560e0817b511ce7921a7c32717b8cce5470f894d

packages/hyperliquid-composer/src/operations/spotMeta.ts
tinom9: In retrospective it would've been cleaner IMO, but nothing inherently wrong here and very subjective, so I'm ok with it.

shankars99: cool will think about this in the next refactor

PR #1727, opened 2025-09-18
types for the cli->internal interface

shankars99:
creating types for cli->internal sdk interface replacing `any`


packages/hyperliquid-composer/src/types/cli-args.ts
```diff
@@ -0,0 +1,80 @@
...
+export interface SpotDeployStateArgs extends TokenIndexArgs {
+    deployerAddress?: string
+}
+
+export interface GetCoreBalancesArgs extends UserQueryArgs {
+    showZero: boolean
+}
+
+// Simple command args that only extend base interfaces
+export interface GenesisArgs extends DeploymentArgs {}
```
lzErfan, line 68: Not sure we'll need all these types here but I guess it would make more sense for future proofing it

PR #1729, opened 2025-09-19
fix: skip lockfile validation during package bumping

shankars99:


tinom9:
@shankars99 we probably want `layerzero-bot` to be running the update lockfiles step.

PR #1732, opened 2025-09-21
Update README.md

0xanmol:
1. Line 127: The string message is incorrect

    - Current: 'Hello from Ethereum!'
    - Should be: 'Hello from Optimism!'
    - Reason: The command is sending from

   Optimism Sepolia, so the message should reflect that

2. Line 133: The destination EID is incorrect
    - Current: --dst-eid 40161 (Ethereum Sepolia)
    - Should be: --dst-eid 40232 (Optimism Sepolia)
    - Reason: The description says

 "Arbitrum Sepolia to Optimism Sepolia" but uses Ethereum Sepolia's EID


DanL0:
conflicts

PR #1734, opened 2025-09-23
DEVREL-787 build: set up merge queue

nazreen:
this PR ensures that the checks are triggered for a PR when it is added into a merge queue

### DIFF
--- a/.changeset/small-starfishes-crash.md
+++ b/.changeset/small-starfishes-crash.md
@@ -1,5 +1,0 @@
----
-"@layerzerolabs/oft-solana-example": patch
----
-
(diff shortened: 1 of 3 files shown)


DanL0 (dismissed):

nazreen:
reverting change to `cancel-in-progress` as it turns out

```
group: ${{ github.workflow }}-${{ github.ref }}
```
which is already in the yaml
is sufficient to ensure that CIs run from inside PRs do not use the same slot as CIs run from in the merge queue

they do not share the same group, as the `ref` would be different for each

tinom9 (approved):

PR #1737, opened 2025-09-29
fix: example lockfiles ci

tinom9:
### Description

- Update stale example lockfiles.
- Re-enable lockfile verification CI step as a new workflow.
- Add extra step to the publish packages workflow that creates an additional PR to update example lockfiles after the NPM packages have been published. Trying to update example lockfiles before changeset PR is merged to main would fail as NPM packages would not yet be published.


.github/workflows/reusable-publish.yaml
```diff
@@ -98,3 +98,48 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.LAYERZERO_BOT_GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN_PUBLISHER }}
+
+      # This should only run on the main branch after the packages are published to NPM
+      # to make sure the updated packages are available when updating the lockfiles.
+      - name: Create additional lockfile update PR
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
shankars99: i dont think this uses the `LAYERZERO_BOT_GITHUB_TOKEN` personal access token that we have. so this commit comes from a generic bot

.github/workflows/reusable-publish.yaml
```diff
@@ -98,3 +98,48 @@ jobs:
...
+          if git diff --quiet; then
+            echo "No changes to commit, skipping PR creation"
+            exit 0
+          fi
+
+          git add .
+          git commit -m "chore: automatically update example lockfiles"
+          git push --force origin "$BRANCH"
+
+          if gh pr view "$BRANCH" --json state --jq '.state' 2>/dev/null | grep -q "OPEN"; then
```
shankars99: dont think the package `gh` exists on default

.github/workflows/reusable-publish.yaml
```diff
@@ -98,3 +98,48 @@ jobs:
...
+            echo "Open PR already exists for branch $BRANCH"
+            exit 0
```
shankars99: why exit and not commit to?

.github/workflows/reusable-publish.yaml
shankars99, line 1: id like this part to be in the lockfiles part. you can chain them like 

```
on:
  workflow_run:
    workflows: ["reusable-publish"]
    types:
      - completed
```

and do some other stuff to force it when it happens on main. reusable-publish only runs on main iirc but wouldnt hurt to be super specific

.github/workflows/reusable-publish.yaml
tinom9: The idea is that the commit comes from the generic bot, but the GH PR is created by `layerzero-bot`.

This is because GH generic bot does not trigger other GH actions to avoid becoming cyclical. I believe `layerzero-bot` is there to circumvent this.

.github/workflows/reusable-publish.yaml
shankars99, line 1: also think replacing it with a pre-existing action like `https://github.com/peter-evans/create-pull-request` would make life a lot easier instead of re-inventing the wheel?

.github/workflows/reusable-publish.yaml
tinom9: I ran a test in a brand new private repo and it did exist.

.github/workflows/reusable-publish.yaml
tinom9: Commit & push is a couple lines above.

.github/workflows/reusable-publish.yaml
tinom9: I like that. Will do.

.github/workflows/reusable-publish.yaml
shankars99: true

.github/workflows/reusable-publish.yaml
tinom9: I don't think there's much boilerplate in the action step.

I like keeping it explicit and I don't love adding new dependencies with write access to our codebase.

.github/workflows/reusable-publish.yaml
shankars99: ah right that makes sense

.github/workflows/reusable-publish.yaml
shankars99: if its something kinda simple like this i think it being a popular action makes maintenance easier? im cool with either since this repo should be deprecated in a few months

tinom9: Keeping it as is then.

tinom9: Done.

shankars99 (approved):

rad-lz (approved):

PR #1717, opened 2025-09-12
DEVREL-769 feat(examples/native-oft-adapter): insert standard send task

nazreen:
## Proof of Test

https://testnet.layerzeroscan.com/tx/0x813b69c00a9be791691118200d2e960719676e4fc6e32bce65bb0c7eef9f67a8

https://testnet.layerzeroscan.com/tx/0x6ed83a3402be85bd65760a5f1245bfe728f8b71be9d0cd8bbae901eb058fd405

## Notes

The Proof of Test deployment uses only 2 chains: Arbitrum and Optimism instead of the 3 chains, for simplicity.

The change to `turbon.json` was made because lint was erroring with:

```
@layerzerolabs/native-oft-adapter-example:lint
  cache bypass, force executing cf31145bca08b98f
   WARN  Issue while reading "/__w/devtools/devtools/.npmrc". Failed to replace env in config: ${NPM_TOKEN}
  
  > @layerzerolabs/native-oft-adapter-example@0.2.8 lint /__w/devtools/devtools/examples/native-oft-adapter
  > $npm_execpath run lint:js && $npm_execpath run lint:sol
  
   WARN  Issue while reading "/__w/devtools/devtools/.npmrc". Failed to replace env in config: ${NPM_TOKEN}
  
  > @layerzerolabs/native-oft-adapter-example@0.2.8 lint:js /__w/devtools/devtools/examples/native-oft-adapter
  > eslint '**/*.{js,ts,json}' && prettier --check .
  
  
  /__w/devtools/devtools/examples/native-oft-adapter/tasks/sendEvm.ts
    7:53  error  Unable to resolve path to module '@layerzerolabs/devtools-evm-hardhat'  import/no-unresolved
    8:48  error  Unable to resolve path to module '@layerzerolabs/io-devtools'           import/no-unresolved
  
  /__w/devtools/devtools/examples/native-oft-adapter/tasks/sendOFT.ts
    4:40  error  Unable to resolve path to module '@layerzerolabs/devtools-evm-hardhat'  import/no-unresolved
  
  /__w/devtools/devtools/examples/native-oft-adapter/tasks/utils.ts
    46:71  error  Unable to resolve path to module '@layerzerolabs/io-devtools'  import/no-unresolved
  
  ✖ 4 problems (4 errors, 0 warnings)
  ```
  
even though those packages have been added to the example's package.json and the lockfile had been updated correctly. 
 
[...]

PR #1743, opened 2025-10-02
fix: ci build step after linting

tinom9:
### Description

- #1717 introduced a `build` dependency on turbo `lint` commands ([turbo.json](https://github.com/LayerZero-Labs/devtools/blob/762cfa24446ed69501d61e1fe56b9fb2472eeb97/turbo.json#L22-L29)).
- A file is being created in the `pnpm lint` step in the CI in the arm64 machine, with non-UFT-8 characters in its filename. This is `./packages/test-devtools-evm-hardhat/p@���@8` ([workflow run](https://github.com/LayerZero-Labs/devtools/actions/runs/18210429126/job/51849699705)).
- The `build` dependency before `lint` is only needed to circumvent eslint's `'import/no-unresolved'` error, which is already re-classified in other packages as a warning so that we don't need to build before linting.
- To avoid diving deeper into the created non-UTF-8 filename, we can just apply the ostrich algorithm.

   image

### DIFF
--- a/examples/native-oft-adapter/.eslintrc.js
+++ b/examples/native-oft-adapter/.eslintrc.js
@@ -7,5 +7,6 @@ module.exports = {
         // @layerzerolabs/eslint-config-next defines rules for turborepo-based projects
         // that are not relevant for this particular project
         'turbo/no-undeclared-env-vars': 'off',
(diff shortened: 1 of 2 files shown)


shankars99 (approved):

ravinagill15 (commented):
amazing 🤯 ~Let's add a changeset for the native-oft-adapter example package~ changeset not needed for linting

PR #1746, opened 2025-10-03
feat: allow multiple solana alts in oft send examples

tinom9:
### Description

- `@layerzerolabs/oft-solana-example`
  - Allow multiple Solana ALTs in oft send examples.
  - Fix `composeMsg` param not being passed forward in send OFT task.
  - Fix `composeMsg` and `extraOptions` parsing in Solana send OFT task.
- `@layerzerolabs/lzapp-migration-example`
  - Apply Solana send OFT task changes.


examples/lzapp-migration/tasks/solana/index.ts
```diff
@@ -140,27 +141,30 @@ export const getSolanaDeployment = (
...
     if (!existsSync(filePath)) {
+        DebugLogger.printWarning(KnownWarnings.SOLANA_DEPLOYMENT_NOT_FOUND)
         throw new Error(`Could not find Solana deployment file for eid ${eid} at: ${filePath}`)
     }
 
     const fileContents = readFileSync(filePath, 'utf-8')
     return JSON.parse(fileContents)
 }
 
+// TODO: move below outside of solana folder since it's generic
```
ravinagill15, line 159: Will we be addressing this TODO considering this is repo will not be maintained for much longer?

examples/lzapp-migration/package.json
```diff
@@ -29,7 +29,7 @@
     "@layerzerolabs/export-deployments": "^0.0.16",
     "@layerzerolabs/io-devtools": "^0.3.0",
     "@layerzerolabs/lz-evm-sdk-v1": "^3.0.49",
-    "@layerzerolabs/oft-v2-solana-sdk": "^3.0.64",
+    "@layerzerolabs/oft-v2-solana-sdk": "^3.0.136",
```
ravinagill15, line 32: Curious as to why this is needed -- the `^` should allow updating to all minor/patch version automatically no?

examples/lzapp-migration/tasks/common/utils.ts
```diff
@@ -47,3 +47,10 @@ export const createSdkFactory = (
 export { createSolanaSignerFactory }
 
 export { DebugLogger, KnownErrors, KnownOutputs, KnownWarnings } from '@layerzerolabs/io-devtools'
+
+export function isEmptyOptionsSolana(optionsBytes?: Uint8Array): boolean {
+    if (!optionsBytes) {
+        return true // Treat undefined or null as empty options
+    }
+    return Buffer.from(optionsBytes).toString('hex') === '0003' // empty options type 3 without 0x prefix
+}
```
ravinagill15: This is unused in `lzapp-migration/tasks/solana/sendSolana.ts`

tinom9: You're right. Removed!

examples/lzapp-migration/package.json
tinom9: 3.0.136 introduces new functionality (passing multiple ALTs).

We don't want posterior package updates to resolve to an earlier version for some reason, that's the reasoning.

examples/lzapp-migration/tasks/solana/index.ts
tinom9: Probably not.

The reason to move it forward is that it's the same code that exists in `examples/oft-solana`, so I believe the comments should also be matching.

Not that I love having repeated logic here, and the most desirable solution would be to extract it to a common OFT Solana package, but that'll not be happening.

PR #1749, opened 2025-10-06
Refactor msg value to be passed down vs. direct reference

TRileySchwarz:


packages/ovault-evm/contracts/VaultComposerSyncNative.sol
```diff
@@ -59,9 +69,9 @@ contract VaultComposerSyncNative is VaultComposerSync, IVaultComposerSyncNative
      * @param _sendParam The parameters for the send operation
      * @param _refundAddress Address to receive tokens and native on Pool failure
      */
-    function _sendRemote(address _oft, SendParam memory _sendParam, address _refundAddress) internal override {
+    function _sendRemote(address _oft, SendParam memory _sendParam, address _refundAddress, uint256 _msgValue) internal override {
         /// @dev msg.value passed in this call is used as LayerZero fee
-        uint256 msgValue = msg.value;
+        uint256 msgValue = _msgValue;
```
tinom9, line 74: This line can be removed.

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -344,35 +344,63 @@ contract VaultComposerSync is IVaultComposerSync, ReentrancyGuard {
...
+        if (shareERC20 != address(VAULT)) {
+            revert ShareTokenNotVault(shareERC20, address(VAULT));
+        }
+
+        /// @dev ShareOFT must be an OFT adapter. We can infer this by checking 'approvalRequired()'.
+        /// @dev burn() on tokens when a user sends changes totalSupply() which the asset:share ratio depends on.
+        if (!IOFT(shareOFT).approvalRequired()) revert ShareOFTNotAdapter(shareOFT);
+
+        /// @dev Approve the share adapter with the share tokens held by this contract
+        IERC20(shareERC20).approve(shareOFT, type(uint256).max);
```
tinom9:
```suggestion
        /// @dev Approve the share adapter with the share tokens held by this contract
        IERC20(shareERC20).approve(shareOFT, type(uint256).max);

        /// @dev ShareOFT must be an OFT adapter. We can infer this by checking 'approvalRequired()'.
        /// @dev burn() on tokens when a user sends changes totalSupply() which the asset:share ratio depends on.
        if (!IOFT(shareOFT).approvalRequired()) revert ShareOFTNotAdapter(shareOFT);
```

Reorder these lines so that it matches `_initializeAssetToken`.

tinom9: Or reorder the ones at `_initializeAssetToken` to keep the revert before the expensive `approve` opperation.

packages/ovault-evm/contracts/VaultComposerSync.sol
```diff
@@ -344,35 +344,63 @@ contract VaultComposerSync is IVaultComposerSync, ReentrancyGuard {
...
+        * @dev Internal function to validate the share token compatibility
+        * @dev Validate part of the constructor in an overridable function due to differences in asset and OFT token
+        * @param _shareOFT The address of the share OFT
+        * @return shareOFT The address of the share OFT
+        * @return shareERC20 The address of the share ERC20 token
+        * @notice Share token must be the vault itself
+        * @notice Share OFT must be an adapter (approvalRequired() returns true)
+     */
+    function _initializeShareToken(address _shareOFT) internal virtual returns (address shareOFT, address shareERC20) {
+        shareOFT = _shareOFT;
```
tinom9: I would avoid adding `shareOFT` as we're just returning the parameter, and I'm not seeing a case in which `shareOFT` mutates.

```suggestion
    function _initializeShareToken(address _shareOFT) internal virtual returns (address shareERC20) {
```

Very subjective though.

tinom9: Actually, since we're using `VAULT` and we're already assuming other immutable variables have been set, I'd avoid passing `_shareOFT` and just set it beforehand.

```suggestion
    function _initializeShareToken() internal virtual returns (address shareERC20) {
```

packages/ovault-evm/contracts/VaultComposerSync.sol
shankars99, line 1: natspec update for `msgValue` addition. can be done later too

packages/ovault-evm/contracts/VaultComposerSyncNative.sol
```diff
@@ -59,9 +69,9 @@ contract VaultComposerSyncNative is VaultComposerSync, IVaultComposerSyncNative
      * @param _sendParam The parameters for the send operation
      * @param _refundAddress Address to receive tokens and native on Pool failure
      */
-    function _sendRemote(address _oft, SendParam memory _sendParam, address _refundAddress) internal override {
+    function _sendRemote(address _oft, SendParam memory _sendParam, address _refundAddress, uint256 _msgValue) internal override {
         /// @dev msg.value passed in this call is used as LayerZero fee
```
shankars99:
```suggestion
        /// @dev msg.value passed in this call is used as LayerZero fee
```

few cleanups like this stale comment. also should we call this `_msgValue` ? while its the same as `msg.value` in `VaultComposerSync` its not really the same in the native case. calling it `msgFee` or something else?

packages/ovault-evm/contracts/VaultComposerSyncNative.sol
TRileySchwarz: This is required as we potentially mutate the msgValue depending on if OFT == assetOFT

packages/ovault-evm/contracts/VaultComposerSync.sol
TRileySchwarz: Done

packages/ovault-evm/contracts/VaultComposerSync.sol
TRileySchwarz: Yeah, updated these

packages/ovault-evm/contracts/VaultComposerSync.sol
TRileySchwarz: done

packages/ovault-evm/contracts/VaultComposerSyncNative.sol
TRileySchwarz: should be updated

PR #1763, opened 2025-10-16
DEVREL-833 check Solana local decimals when creating OFT

nazreen:
## Changes
- introduce two new utils: `localDecimalsToMaxSupply` (Solana) and `formatTokenAmountCompact` in `packages` (added unit tests)
- `createOFT` and `createOFTAdapter` scripts in `oft-solana` and `lzapp-migration`: added prompt to confirm that states what max supply will be based on local decimals
- introduce 'ci' flag for consistency with wire task

## Example runs

```
examples/oft-solana $ pnpm hardhat lz:oft:solana:create --eid 40168 --program-id $PROGRAM_ID --only-oft-store true --local-decimals 18 --ci
info:    Using Solana keypair from SOLANA_PRIVATE_KEY => <PUBLIC_KEY_REDACTED>

? You have chosen 18 local decimals. The maximum supply of your Solana OFT token will be 18 (~18).

examples/oft-solana $ pnpm hardhat lz:oft:solana:create --eid 40168 --program-id $PROGRAM_ID --only-oft-store true --local-decimals 10 --ci
info:    Using Solana keypair from SOLANA_PRIVATE_KEY => <PUBLIC_KEY_REDACTED>

? You have chosen 10 local decimals. The maximum supply of your Solana OFT token will be 1,844,674,407 (~1.8B). Continue ?
```

## Testing
- run `pnpm hardhat lz:oft:solana:create --eid 40168 --program-id <PROGRAM_ID> --only-oft-store true --local-decimals 7` and view the confirmation prompt.
- rerun the above with varying values for `--local-decimals` to see the varying max supply

For ease of testing you can use `export PROGRAM_ID= 9j9Wp46t7GyDSU3NVV8AmYMdXZHoiZmAzVuDSVDY2jy`

### DIFF
--- a/.changeset/selfish-timers-allow.md
+++ b/.changeset/selfish-timers-allow.md
@@ -1,0 +1,5 @@
+---
+"@layerzerolabs/verify-contract": patch
+---
+
+STABLE mainnet added to the networks
--- a/packages/verify-contract/src/common/networks.ts
+++ b/packages/verify-contract/src/common/networks.ts
@@ -410,6 +410,13 @@ export const networks: Record<string, NetworkDefinition> = {
         aliases: ['sophon-sepolia-testnet', 'sophon-testnet'],
     },
 
+    // Stable
+    stable: {
+        chainId: 988,
+        apiUrl: ETHERSCAN_V2_URL,
+        aliases: ['stable-mainnet'],
+    },
+
     // Swellchain
     swellchain: {
         chainId: 1923,


nazreen:
will be updating this PR to remove the concept of 'max recommended local decimals' and just prompt for confirmation every single time to ensure the dev is aware of the max total supply given the local decimals value

nazreen:
updated

examples/lzapp-migration/tasks/solana/utils.ts
```diff
@@ -32,3 +32,52 @@ export function silenceSolana429(connection: Connection): void {
...
+    { base: 1_000_000_000n, suffix: 'B' },
+    { base: 1_000_000n, suffix: 'M' },
+    { base: 1_000n, suffix: 'K' },
+]
+
+/**
+ * Max whole-token supply on Solana for a given localDecimals,
+ * formatted as T/B/M/K, else plain number. Rounded half up.
+ */
+export function maxSupplyHuman(localDecimals: number, precision = 1): string {
```
shankars99: nit: would call it whole tokens instead of human

examples/lzapp-migration/tasks/solana/utils.ts
```diff
@@ -32,3 +32,52 @@ export function silenceSolana429(connection: Connection): void {
...
     }) as typeof process.stderr.write
 }
+
+// Max whole-token supply on Solana (u64)
+const U64_MAX = (1n << 64n) - 1n
+const UNITS = [
+    { base: 1_000_000_000_000n, suffix: 'T' },
+    { base: 1_000_000_000n, suffix: 'B' },
+    { base: 1_000_000n, suffix: 'M' },
+    { base: 1_000n, suffix: 'K' },
```
shankars99: whats `K`?

examples/lzapp-migration/tasks/solana/utils.ts
shankars99: also whats `precision`?

examples/lzapp-migration/tasks/solana/utils.ts
```diff
@@ -32,3 +32,52 @@ export function silenceSolana429(connection: Connection): void {
...
+            }
+
+            const frac = scaled % pow
+            const fracStr = precision === 0 ? '' : frac.toString().padStart(precision, '0').replace(/0+$/, '')
+            return fracStr ? `${intPart}.${fracStr}${suffix}` : `${intPart}${suffix}`
+        }
+    }
+
+    return whole.toString()
+}
```
shankars99: definitely needs more docs

examples/lzapp-migration/tasks/solana/utils.ts
nazreen: for 'thousands' 

e.g. 100_000 -> 100K

examples/lzapp-migration/tasks/solana/utils.ts
nazreen: `precision` = number of decimal places in formatted amount

e.g. if amount is 1_234_500 (arbitrary number, doesn't map to any local decimals' max supply value) and `precision` is 0, formatted amount is `1M`
if `precision` is 1, then formatted amount is `1.2M`

I'll change the param name to something more obvious like `maxDisplayDecimals`

noted on using 'whole tokens' in function name

nazreen:
@shankars99 pushed changes to address your feedback

examples/lzapp-migration/tasks/solana/utils.ts
```diff
@@ -32,3 +32,81 @@ export function silenceSolana429(connection: Connection): void {
...
+export function localDecimalsToMaxSupply(localDecimals: number): bigint {
+    if (!Number.isInteger(localDecimals) || localDecimals < 0) {
+        throw new Error('localDecimals must be a non-negative integer')
+    }
+    const scalingFactor = 10n ** BigInt(localDecimals)
+    return U64_MAX / scalingFactor
+}
+
+/** Formats a whole-token amount into a compact human-readable form using K/M/B/T. */
+export function formatAmount(whole: bigint, maxDisplayDecimals = 1): string {
```
shankars99: why not take in localDecimals and call localDecimalsToMaxSupply() to get the maxSupply? i like the ldToMax being its own function

examples/lzapp-migration/tasks/solana/utils.ts
```diff
@@ -32,3 +32,81 @@ export function silenceSolana429(connection: Connection): void {
...
+    }
+
+    for (let i = 0; i < UNITS.length; i++) {
+        const { base, suffix } = UNITS[i]
+        if (whole >= base) {
+            const pow = 10n ** BigInt(maxDisplayDecimals)
+            let scaled = (whole * pow + base / 2n) / base
+            let intPart = scaled / pow
+
+            if (intPart >= 1000n && i === 1) {
```
shankars99: how is this possible? 
`i = 1` is billion and `1000 * billion` is `trillion` which is `i = 0` right?

nazreen: > i = 1 is billion and 1000 * billion is trillion which is i = 0 right?

Yes, that's correct. But also yeah this part is buggy as it doesn't turn 1000M into 1B (or 1000K into 1M). Will fix.

nazreen:
@shankars99 I think for debtools, correctness should be sufficient. I don't really want to spend too much time on optimizing implementations. For correctness, I've added unit tests. You can check them out and see if any cases are missed.

shankars99:
not really optimizations but it makes it easier to port these tooling to the new one. the new monorepo internal would not have all of these scripts that we have and its upto us to port them over. 

we also do not know when devtools will be out on public launch. its safer and better to keep our current implementations the best they can be even if it takes some extra time to compensate for a doomsday case

TRileySchwarz:
This formatter shouldnt show rounded numbers.

"... Solana OFT token will be 1.8B tokens..."

This is misleading and not true, should simply show the number formatted with decimals and commas to the actual maximum value.

packages/devtools-solana/src/common/numbers.ts
```diff
@@ -0,0 +1,11 @@
+// Compute the maximum whole-token supply for a Solana SPL mint given its local decimals.
+
+const U64_MAX = (BigInt(1) << BigInt(64)) - BigInt(1)
+
+export function localDecimalsToMaxSupply(localDecimals: number): bigint {
```
TRileySchwarz: This name should be changed, this doesn't return 'MaxSupply', this returns 'MaxWholeTokens' or, 'MaxTokensLessDecimals' etc.

packages/devtools/src/common/numbers.ts
```diff
@@ -0,0 +1,50 @@
...
+// Supports: K (thousand), M (million), B (billion), T (trillion).
+
+const UNITS = [
+    { base: BigInt(1000) * BigInt(1000) * BigInt(1000) * BigInt(1000), suffix: 'T' },
+    { base: BigInt(1000) * BigInt(1000) * BigInt(1000), suffix: 'B' },
+    { base: BigInt(1000) * BigInt(1000), suffix: 'M' },
+    { base: BigInt(1000), suffix: 'K' },
+]
+
+export function formatTokenAmount(whole: bigint, maxDisplayDecimals = 1): string {
```
TRileySchwarz: Why not use a builtin:

```
export const formatTokenAmount = (whole: bigint, maxDecimals = 1): string =>
    new Intl.NumberFormat('en-US', {
        notation: 'compact',
        maximumFractionDigits: maxDecimals,
        roundingMode: 'floor'
    }).format(Number(whole))
```

packages/devtools/test/formatAmount.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    it('formats billions with B suffix (no 1000M rollover)', () => {
+        expect(formatTokenAmount(BigInt(1_000_000_000))).toBe('1B')
+        expect(formatTokenAmount(BigInt(1_000_000_000))).not.toBe('1000M')
+    })
+
+    it('handles values near 1T boundary without unintended rollover', () => {
+        expect(formatTokenAmount(BigInt(999_400_000_000))).toBe('999.4B')
+        expect(formatTokenAmount(BigInt(999_600_000_000), 1)).toBe('999.6B')
+        expect(formatTokenAmount(BigInt(999_950_000_000), 1)).toBe('999.95B')
```
TRileySchwarz: Why does 1 decimal return 999.95B? Shouldn't it return 999.9B?

packages/devtools/test/formatAmount.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    })
+
+    it('trims trailing zeros in fractional part', () => {
+        expect(formatTokenAmount(BigInt(1_500_000), 2)).toBe('1.5M')
+        expect(formatTokenAmount(BigInt(12_340_000), 3)).toBe('12.34M')
+    })
+
+    it('throws on invalid precision values', () => {
+        expect(() => formatTokenAmount(BigInt(1_000), -1 as unknown as number)).toThrow()
+        expect(() => formatTokenAmount(BigInt(1_000), 7)).toThrow()
```
TRileySchwarz: Why should this throw? if I want to round to 7 decimals of precision, and it doesn't exist in that numeration, just dont include them?

examples/oft-solana/tasks/solana/createOFTAdapter.ts
```diff
@@ -63,6 +66,14 @@ task('lz:oft-adapter:solana:create', 'Creates new OFT Adapter (OFT Store PDA)')
             const mint = publicKey(mintStr)
 
             const mintPDA = await getMint(connection, new PublicKey(mintStr), undefined, new PublicKey(tokenProgramStr))
+            const mintDecimals = mintPDA.decimals
+
+            const maxSupply = formatTokenAmount(localDecimalsToMaxSupply(mintDecimals))
```
TRileySchwarz: This isnt the MaxSupply, its the maxSupplyFloor, or maxSupplyRounded etc.

examples/lzapp-migration/test/scripts/solanaUtils.script.test.ts
```diff
@@ -0,0 +1,57 @@
+import { formatTokenAmount } from '../../tasks/solana/utils'
```
TRileySchwarz: Why is this series of tests copy and pasted into lzapp-migration example? This lib function is already tested inside its local package?

nazreen: My bad. Initially the new helpers were defined locally in examples. Had a commit to clean it up, but hadn't pushed it. Just pushed.

packages/devtools-solana/src/common/numbers.ts
nazreen: You're right. Thanks, updated.

packages/devtools/test/formatAmount.test.ts
nazreen: I wanted it to not lose information so amended the helper to force include additional decimals if they are non zero. But taking into account your other comment, I think what I'll do now is print both the raw amount and this formatted amount. So this helper can be simplified (will use the builtin you suggested) since it'll be okay to display the truncated formatted amount given that the formatted (and potentially truncated) amount is just there for ease of reading. 

e.g.

Instead of 

```
...The maximum supply of your Solana OFT token will be 1.8B tokens. Continue? › (Y/n)
```

It'll print


```
...The maximum supply of your Solana OFT token will be 1,844,674,407 (~1.8B). Continue? › (Y/n)
```

packages/devtools/src/common/numbers.ts
nazreen: updated to use this suggestion

packages/devtools/test/formatAmount.test.ts
nazreen: amended per your suggestion.

examples/oft-solana/tasks/solana/createOFTAdapter.ts
nazreen: These lines have now been amended to

```
const maxSupplyRaw = localDecimalsToMaxWholeTokens(decimals)
const maxSupplyFormatted = new Intl.NumberFormat('en-US').format(maxSupplyRaw)
const maxSupplyCompacted = formatTokenAmountCompact(maxSupplyRaw)
```

packages/devtools/test/formatAmount.test.ts
TRileySchwarz: Yeah I like having both, actual and then rounded for ease of reading

examples/oft-solana/tasks/solana/createOFT.ts
```diff
@@ -223,6 +231,15 @@ task('lz:oft:solana:create', 'Mints new SPL Token and creates new OFT Store acco
 
             // EOF: Validate combination of parameters
 
+            const maxSupplyRaw = localDecimalsToMaxWholeTokens(decimals)
+            const maxSupplyFormatted = new Intl.NumberFormat('en-US').format(maxSupplyRaw)
```
TRileySchwarz: I would probably just have this formateTokenAmounts return all these values, so its not expensive or inefficient to always do the calcaultions, and that way there is one helper vs. multiple

St0rmBr3w (approved):
LGTM!

examples/oft-solana/tasks/solana/createOFTAdapter.ts
nazreen: @TRileySchwarz  can you confirm whether this thread can be resolved?

examples/oft-solana/tasks/solana/createOFT.ts
nazreen: I'm assuming you meant there should be one helper to serve these 2 lines:

```
    const maxSupplyFormatted = new Intl.NumberFormat('en-US').format(maxSupplyRaw)
    const maxSupplyCompacted = formatTokenAmountCompact(maxSupplyRaw)
```

rather than absorbing what `localDecimalsToMaxWholeTokens` does into `formateTokenAmounts` as I think these two functions do 2 separate things

nazreen: pushed:

```
export function formatTokenAmount(rawAmount: bigint, maxDisplayDecimals = 1): { full: string; compact: string } {
    if (!Number.isInteger(maxDisplayDecimals) || maxDisplayDecimals < 0) {
        throw new Error('maxDisplayDecimals must be an integer between 0 and 6')
    }
    const full = new Intl.NumberFormat('en-US').format(rawAmount)
    const compact = new Intl.NumberFormat('en-US', {
        notation: 'compact',
        maximumFractionDigits: maxDisplayDecimals,
        roundingMode: 'floor',
    }).format(Number(rawAmount))

    return {
        full,
        compact,
    }
}
```

nazreen: @St0rmBr3w had to dismiss your review to make this change, do review again when you can 🙏

examples/oft-solana/tasks/solana/utils.ts
```diff
@@ -83,3 +83,5 @@ export function silenceSolana429(connection: Connection): void {
         return origWrite(chunk, ...args)
     }) as typeof process.stderr.write
 }
+export { localDecimalsToMaxWholeTokens } from '@layerzerolabs/devtools-solana'
+export { formatTokenAmount } from '@layerzerolabs/devtools'
```
shankars99: why not just import it in `createOFT` and `createOFTAdapter`

examples/oft-solana/tasks/solana/createOFTAdapter.ts
```diff
@@ -63,6 +66,15 @@ task('lz:oft-adapter:solana:create', 'Creates new OFT Adapter (OFT Store PDA)')
...
             const mintPDA = await getMint(connection, new PublicKey(mintStr), undefined, new PublicKey(tokenProgramStr))
+            const mintDecimals = mintPDA.decimals
+
+            const maxSupplyRaw = localDecimalsToMaxWholeTokens(mintDecimals)
+            const maxSupplyFormatted = formatTokenAmount(maxSupplyRaw)
+            const maxSupplyStatement = `You have chosen ${mintDecimals} local decimals. The maximum supply of your Solana OFT token will be ${maxSupplyFormatted.full} (~${maxSupplyFormatted.compact}).\n`
+            const confirmMaxSupply = await promptToContinue(maxSupplyStatement)
+            if (!confirmMaxSupply) {
+                return
+            }
```
shankars99: do adapters on solana define the max supply?

nazreen: Not the adapters themselves, but figured it would be good to point out to the partner dev what the max supply would be given the Solana token (Mint Account) that they supply to this command.

examples/oft-solana/tasks/solana/utils.ts
nazreen: This was a remnant of the refactoring that I did as those helpers were initially defined in here. Hope it's okay if we just leave them in here first since it's a minor thing. I'd rather not dismiss the existing review by Krak.

examples/oft-solana/tasks/solana/createOFTAdapter.ts
shankars99: i would probably print out the message differently, something like "Underlying token has X decimals. The maximum...." since its not a choice that the `OFTAdapter` deployer can make

examples/oft-solana/tasks/solana/utils.ts
shankars99: yeah im fine with that, but if there are any other changes to the PR then we can make this change too

packages/devtools/test/formatAmount.test.ts
TRileySchwarz: it should not be in the "throws on invalid prevision values" test block, it should be in a test thats "NOT" supposed to throw

examples/oft-solana/tasks/solana/createOFTAdapter.ts
```diff
@@ -63,6 +66,15 @@ task('lz:oft-adapter:solana:create', 'Creates new OFT Adapter (OFT Store PDA)')
             const mint = publicKey(mintStr)
 
             const mintPDA = await getMint(connection, new PublicKey(mintStr), undefined, new PublicKey(tokenProgramStr))
+            const mintDecimals = mintPDA.decimals
+
+            const maxSupplyRaw = localDecimalsToMaxWholeTokens(mintDecimals)
+            const maxSupplyFormatted = formatTokenAmount(maxSupplyRaw)
```
TRileySchwarz: I would destructure these here, vs. indexing later

examples/oft-solana/tasks/solana/utils.ts
nazreen: implemented

examples/oft-solana/tasks/solana/createOFTAdapter.ts
nazreen: implemented

examples/oft-solana/tasks/solana/createOFTAdapter.ts
nazreen: implemented

packages/devtools/test/formatAmount.test.ts
nazreen: fixed, thanks

shankars99 (dismissed):

nazreen:
@shankars99 your review was auto-dismissed as there was a merge conflict with one of the pnpm lockfiles

shankars99 (approved):

PR #1765, opened 2025-10-18
fix/upgradeable-deploy-script

shankars99:
`hardhat-deploy` does not work well with `TUP` deployments 

```js
if (options.deterministicDeployment &&
    diffResult.address &&
    diffResult.address.toLowerCase() !== deployment.address.toLowerCase())
```

is the boolean logic used in `hardhat-deploy` which makes it so you need `deterministicDeployment` enabled to deploy contracts, which makes it worse to deploy since you need to maintain a bytes key to differentiate between deployments.

The fix is to create our own Transparent Upgradeable Proxy (TUP) deployment script htat does it manually by deploying a `proxyAdmin`, `implementation`, `proxy`, and then initializing the proxy with the implementation.

### DIFF
--- a/examples/oft-alt/deploy/MyOFTAltUpgradeable.ts
+++ b/examples/oft-alt/deploy/MyOFTAltUpgradeable.ts
@@ -15,6 +15,7 @@ const deploy: DeployFunction = async (hre) => {
 
     console.log(`Deploying ${contractName} on network: ${hre.network.name} with ${deployer}`)
 
+    /// EndpointV2Alt deployments are also saved under the filename: EndpointV2
     const endpointV2AltDeployment = await hre.deployments.get('EndpointV2')
 
     const { address: proxyAdminAddress } = await deployProxyAdmin({
--- a/examples/oft-alt/deploy/MyOFTAdapterAlt.ts
+++ b/examples/oft-alt/deploy/MyOFTAdapterAlt.ts
@@ -31,7 +31,9 @@ const deploy: DeployFunction = async (hre) => {
     //     eid: EndpointId.AVALANCHE_V2_TESTNET
     //   }
     // }
-    const endpointV2Deployment = await hre.deployments.get('EndpointAltV2')
+
+    /// EndpointV2Alt deployments are also saved under the filename: EndpointV2
+    const endpointV2AltDeployment = await hre.deployments.get('EndpointV2')
 
     // The token address must be defined in hardhat.config.ts
     // If the token address is not defined, the deployment will log a warning and skip the deployment
@@ -45,7 +47,7 @@ const deploy: DeployFunction = async (hre) => {
         from: deployer,
         args: [
             hre.network.config.oftAdapter.tokenAddress, // LayerZero's EndpointV2 address
-            endpointV2Deployment.address, // LayerZero's EndpointV2 address
+            endpointV2AltDeployment.address, // LayerZero's EndpointV2 address
             deployer, // owner
         ],
         log: true,
--- a/examples/oft-alt/deploy/MyOFTAlt.ts
+++ b/examples/oft-alt/deploy/MyOFTAlt.ts
@@ -31,14 +31,16 @@ const deploy: DeployFunction = async (hre) => {
     //     eid: EndpointId.AVALANCHE_V2_TESTNET
     //   }
     // }
-    const endpointV2Deployment = await hre.deployments.get('EndpointV2')
+
+    /// EndpointV2Alt deployments are also saved under the filename: EndpointV2
+    const endpointV2AltDeployment = await hre.deployments.get('EndpointV2')
 
     const { address } = await deploy(contractName, {
         from: deployer,
         args: [
             'MyOFTAlt', // name
             'MOFTAlt', // symbol
-            endpointV2Deployment.address, // LayerZero's EndpointV2 address
+            endpointV2AltDeployment.address, // EndpointV2 address
             deployer, // owner
         ],
         log: true,


packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
```diff
@@ -0,0 +1,129 @@
...
+    const result = await deploy(proxyName, {
+        from: deployer,
+        contract: proxyContract,
+        args: [implementationAddress, proxyAdminAddress, initializeData],
+        log: true,
+        waitConfirmations: 1,
+        skipIfAlreadyDeployed: true,
+    })
+
+    return { address: result.address }
```
tinom9: I believe you want to return the whole deploy result here to mimic `hardhat-deploy`.

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @param deployer - Address that will deploy the ProxyAdmin
+ * @param owner - Address that will own the ProxyAdmin (defaults to deployer)
+ * @returns Object containing the deployed ProxyAdmin address
+ */
+export async function deployProxyAdmin(
+    hre: HardhatRuntimeEnvironment,
+    contractName: string,
+    deployer: string,
+    owner: string = deployer
+): Promise<{ address: string }> {
```
tinom9: This is a nit, but I like the idea of passing arguments as an object once it's >=3 args. Less error prone.

examples/oft-alt/deploy/MyOFTAltUpgradeable.ts
```diff
@@ -1,38 +1,35 @@
...
+    const signer = (await hre.ethers.getSigners())[0]
+    console.log(`Deploying ${contractName} on network: ${hre.network.name} with ${signer.address}`)
 
-    const { deploy } = deployments
-    const { deployer } = await getNamedAccounts()
+    const endpointV2Deployment = await hre.deployments.get('EndpointV2')
+    const endpointAddress = endpointV2Deployment.address
 
-    assert(deployer, 'Missing named deployer account')
+    const { address: proxyAdminAddress } = await deployProxyAdmin(hre, contractName, signer.address)
```
tinom9: Since this is an example, I would pass `signer.address` as the owner as well to be extra explicit.

examples/oft-alt/deploy/MyOFTAltUpgradeable.ts
```diff
@@ -1,38 +1,35 @@
...
-            execute: {
-                init: {
-                    methodName: 'initialize',
-                    args: ['MyOFT', 'MOFT', deployer],
-                },
-            },
-        },
-    })
+    const initializeInterface = new hre.ethers.utils.Interface([
+        'function initialize(string memory name, string memory symbol, address delegate) public',
```
tinom9:
```suggestion
        'function initialize(string memory name, string memory symbol, address delegate)',
```

examples/oft-upgradeable/deploy/MyOFTAdapterFeeUpgradeable.ts
tinom9, line 1: I like the idea of removing these!

examples/oft-alt/deploy/MyOFTAltUpgradeable.ts
shankars99: aight. im also gonna structure the parameters

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
shankars99: yeah that sounds better

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
shankars99: yeah agreed

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    // Use require for JSON to avoid ES module import assertion issues
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const proxyAdminContract = require('hardhat-deploy/extendedArtifacts/ProxyAdmin.json')
+    const proxyAdminName = `${contractName}_ProxyAdmin`
+
+    return await deploy(proxyAdminName, {
+        from: deployer,
+        contract: proxyAdminContract,
+        args: [owner],
```
ravinagill15, line 35: Should this be deployer if owner is undefined?

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
```diff
@@ -0,0 +1,151 @@
...
+export async function saveCombinedDeployment({
+    hre,
+    contractName,
+}: {
+    hre: HardhatRuntimeEnvironment
+    contractName: string
+}): Promise<void> {
+    const existing = await hre.deployments.getOrNull(contractName)
+    if (!existing) {
+        const proxyName = `${contractName}_Proxy`
```
ravinagill15, line 138: Should we also save a deployment file named the exact same as the contract name pointing to the proxy address? It would be mostly redundant, but I have seen this pattern before and am curious as to your thoughts on it. It could cause confusion or could be helpful when trying to get the "main" contract to interact with

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
shankars99: yeah. the default value of owner is set to the deployer

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
shankars99: ah i think we do that in devtools rn. yeah should prolly stick to equivalence

shankars99: oh nvm we do it `await hre.deployments.save(contractName, deployment)`

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
ravinagill15: Ah I see it a couple lines above

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
ravinagill15: Got it, also see it below

ravinagill15 (approved):

packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
tinom9: It's what `hardhat-deploy` does and I think it's what makes the most sense:
- You have the implementation file (`_Implementation`). It has implementation ABI and address. You use it to reference the implementation address.
- You have the proxy file (`${contractName}_Proxy`). It has proxy ABI and address. You use it if you need to use a proxy-specific function.
- You have the baseline file (`contractName`). It has implementation ABI and proxy address. You use it to interact with the contract as we do in most places.

PR #1769, opened 2025-10-21
Integrate Stargate Native asset pools into OVault SDK

arekLZ:
To use the stargate asset pools you should pass a `0x0` address or equivalent to the token address, and the pool address to the OFT address. 

This PR also changes the return type. Before you would supply the message fee as the message value, now value is returned separately. If using a native vault then the value will be the message fee + the amount you are vaulting. If it's an ERC vault then the value will just be the message fee.

Testing doc: https://docs.google.com/spreadsheets/d/1Gx_tzO8Egtc9kg_5GSpZ0lCEO4Jf9iYhEARW2fD4vVI/edit?gid=0#gid=0

### DIFF
--- a/packages/ovault-evm/src/oVaultSync/transfer.ts
+++ b/packages/ovault-evm/src/oVaultSync/transfer.ts
@@ -73,25 +73,20 @@ export class OVaultSyncMessageBuilder {
         dstAmount: bigint
         minDstAmount: bigint
     }> {
-        const { operation, amount, vaultAddress, hubChain, slippage, tokenAddress, oftAddress } = input
+        const { operation, vaultAddress, hubChain, slippage, tokenAddress, oftAddress } = input
 
         const client = createPublicClient({
             chain: hubChain,
             transport: http(),
         })
 
-        let sourceAmount = amount
-
-        if (hexToBigInt(tokenAddress ?? oftAddress) === 0n) {
-            // If it's a native transfer then there can be slippage or a fee taken during the inital hop.
-            sourceAmount = await this.getOutputAmount({
-                ...input,
-                dstAmount: 0n,
-                minDstAmount: 0n,
-                dstAddress: input.walletAddress,
-                tokenAddress: tokenAddress ?? oftAddress,
-            })
-        }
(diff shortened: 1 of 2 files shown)


.gitmodules
shankars99, line 1: undo this change pls. i will take a look once this is done

shankars99 (changes requested):
`.gitmodules` to be reverted pls

.gitmodules
arekLZ: Done

packages/ovault-evm/src/oVaultSync/transfer.ts
```diff
@@ -72,18 +73,31 @@ export class OVaultSyncMessageBuilder {
...
 
         const client = createPublicClient({
             chain: hubChain,
             transport: http(),
         })
 
+        let sourceAmount = amount
+
+        if (hexToBigInt(tokenAddress ?? oftAddress) === 0n) {
+            // If it's a native transfer then there can be slippage or a fee taken during the inital hop.
```
shankars99: there can also be a fee on the oft.send() for non-natives too like USDC stargate pools or just the regular share/asset ofts

packages/ovault-evm/src/types.ts
```diff
@@ -38,17 +39,20 @@ export interface SendParamsInput {
 
     slippage: number // How much slippage to allow for the redemption. This is a percentage, so 0.01 = 1%
 
-    oftAddress: `0x${string}`
+    oftAddress: `0x${string}` // If a hex code equal to 0x0, then the native token is being sent
```
shankars99: this comment should be on the `tokenAddress`

shankars99 (changes requested):
doc nits

packages/ovault-evm/src/types.ts
arekLZ: Ah good point.

packages/ovault-evm/src/oVaultSync/transfer.ts
arekLZ: Will remove the `if` statement then

PR #1776, opened 2025-10-24
Deposit-only vault composer

ravinagill15:
## In this PR:

- reverting on redeem
- will be used for vaults that have a cooldown period which means async txs, so redemptions should be disabled
- in the future we can implement a version of the composer that checks if the cooldown period is off before allowing redemption

### DIFF
--- a/packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
+++ b/packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
@@ -6,8 +6,8 @@ import { VaultComposerSync } from "./VaultComposerSync.sol";
 /**
  * @title VaultComposerSyncDepositOnly
  * @author LayerZero Labs (@ravinagill15)
- * @notice Cross-chain vault composer enabling omnichain vault operations via LayerZero
- * @notice This composer only supports Deposit actions. Redemption is disabled and will revert
+ * @notice Cross-chain vault composer enabling omnichain vault operations via LayerZero.
+ * @notice This composer only supports Deposit actions. Redemption is disabled and will revert.
  *
  * @dev IMPORTANT: This contract intentionally disables redemption functionality by overriding _redeem()
  *      to always revert. This causes unreachable code warnings in the parent VaultComposerSync
@@ -17,11 +17,15 @@ contract VaultComposerSyncDepositOnly is VaultComposerSync {
     error RedemptionDisabled();
 
     /**
-     * @notice Creates a new cross-chain vault composer
-     * @dev Initializes the composer with vault and OFT contracts for omnichain operations
-     * @param _vault The vault contract implementing ERC4626 for deposit/redeem operations
-     * @param _assetOFT The OFT contract for cross-chain asset transfers
-     * @param _shareOFT The OFT contract for cross-chain share transfers
+     * @notice Initializes the VaultComposerSyncDepositOnly contract with vault and OFT token addresses.
+     * @param _vault The address of the ERC4626 vault contract
+     * @param _assetOFT The address of the asset OFT (Omnichain Fungible Token) contract
+     * @param _shareOFT The address of the share OFT contract (must be an adapter)
+     *
+     * Requirements:
+     * - Share token must be the vault itself
+     * - Asset token should match the vault's underlying asset (overridable behavior)
+     * - Share OFT must be an adapter (approvalRequired() returns true)
      */
     constructor(
         address _vault,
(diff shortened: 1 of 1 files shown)


packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
```diff
@@ -0,0 +1,43 @@
...
+     * @param _shareOFT The OFT contract for cross-chain share transfers
+     */
+    constructor(
+        address _vault,
+        address _assetOFT,
+        address _shareOFT
+    ) VaultComposerSync(_vault, _assetOFT, _shareOFT) {}
+
+    /**
+     * @notice Cross-chain redemptions are disabled for this vault
```
tinom9:
```suggestion
     * @notice Redemptions are disabled for this vault
```

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
```diff
@@ -0,0 +1,43 @@
...
+     * @dev Users should interact with the vault directly for redemption
+     */
+    function _redeem(
+        bytes32 /*_redeemer*/,
+        uint256 /*_shareAmount*/
+    ) internal pure override returns (uint256 /*assetAmount*/) {
+        // Disable redemption to prevent cross-chain redemptions
```
tinom9:
```suggestion
        /// @dev Disable all redemptions
```

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
```diff
@@ -0,0 +1,43 @@
...
+     */
+    constructor(
+        address _vault,
+        address _assetOFT,
+        address _shareOFT
+    ) VaultComposerSync(_vault, _assetOFT, _shareOFT) {}
+
+    /**
+     * @notice Cross-chain redemptions are disabled for this vault
+     * @dev Users should interact with the vault directly for redemption
```
tinom9: I'd add a comment about `redeemAndSend` reverting atomically, and `lzCompose` calls refunding to the original chain.

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: UNLICENSED
```
tinom9, line 1: Does it have to be unlicensed? Should we make it MIT?

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
tinom9, line 1: Super nitpick warning.

I like to end all my `@notice` and `@dev` NatSpec comments (not `@param` comments though) with a dot, so that lines like this one don't look weird:

```solidity
 * @notice This composer only supports Deposit actions. Redemption is disabled and will revert
```

Leave it as you see fit though.

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
```diff
@@ -0,0 +1,43 @@
...
+contract VaultComposerSyncDepositOnly is VaultComposerSync {
+    error RedemptionDisabled();
+
+    /**
+     * @notice Creates a new cross-chain vault composer
+     * @dev Initializes the composer with vault and OFT contracts for omnichain operations
+     * @param _vault The vault contract implementing ERC4626 for deposit/redeem operations
+     * @param _assetOFT The OFT contract for cross-chain asset transfers
+     * @param _shareOFT The OFT contract for cross-chain share transfers
+     */
```
tinom9, line 29: Also super nit.

I would stick to the `VaultComposerSync` NatSpec comment.

```solidity
    /**
     * @notice Initializes the VaultComposerSyncDepositOnly contract with vault and OFT token addresses
     * @param _vault The address of the ERC4626 vault contract
     * @param _assetOFT The address of the asset OFT (Omnichain Fungible Token) contract
     * @param _shareOFT The address of the share OFT contract (must be an adapter)
     *
     * Requirements:
     * - Share token must be the vault itself
     * - Asset token should match the vault's underlying asset (overridable behavior)
     * - Share OFT must be an adapter (approvalRequired() returns true)
     */
```

Again, I'm ok with this as is.

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
ravinagill15: @TRileySchwarz thoughts?

shankars99 (approved):

packages/ovault-evm/contracts/VaultComposerSyncDepositOnly.sol
TRileySchwarz: Should be same as the regular vault composer

shankars99:
closed since we did it in https://github.com/LayerZero-Labs/devtools/pull/1778

PR #1777, opened 2025-10-24
patch/custom-upgradeableOft-deployer

shankars99:
Addding in:
1. `skipIfDeployed`
2. deploymentFile name override for multi-mesh repos
3. `mergeABIs` to simulate `hardhat-deploy` to emulate 
```
if ( options.proxy.proxyContract === 'OpenZeppelinTransparentProxy') {
        checkABIConflict = false;
        proxyContract = OpenZeppelinTransparentProxy;
        viaAdminContract = 'DefaultProxyAdmin';
}
```


packages/devtools-evm-hardhat/src/deployment/upgradeable.ts
```diff
@@ -1,151 +1,206 @@
...
-    return await deploy(proxyAdminName, {
-        from: deployer,
+    const proxyAdminFileName = deploymentName ? deploymentName + '_ProxyAdmin' : 'DefaultProxyAdmin'
+    await checkDeploymentExists(hre, proxyAdminFileName)
+
+    return await deploy(proxyAdminFileName, {
         contract: proxyAdminContract,
-        args: [owner],
         log: true,
         waitConfirmations: 1,
```
tinom9, line 41: Should we have an opinion on `{ log: true, waitConfirmations: 1 }`?

shankars99: log always. wait confirmations i dont think so

tinom9 (approved):
I don't think we should've gone the way of "make this compatible with `hardhat-deploy`.

Merging this as is is okay to me, but I'd prefer a big and opinionated `deployTUP` function with `hardhat-deploy` syntax.

shankars99:
> I don't think we should've gone the way of "make this compatible with `hardhat-deploy`.
> 
> Merging this as is is okay to me, but I'd prefer a big and opinionated `deployTUP` function with `hardhat-deploy` syntax.

we definitely wanna be backward compatible with hardhat-deploy, no? a single `deployTUP` is nicer but its imo just a wrapper of these internal functions and most users do not open deploy scripts so it doesn't really affect them

tinom9:
> > I don't think we should've gone the way of "make this compatible with `hardhat-deploy`.
> > Merging this as is is okay to me, but I'd prefer a big and opinionated `deployTUP` function with `hardhat-deploy` syntax.
> 
> we definitely wanna be backward compatible with hardhat-deploy, no? a single `deployTUP` is nicer but its imo just a wrapper of these internal functions and most users do not open deploy scripts so it doesn't really affect them

We're still having 3 code blocks (4 if you count proxy + implementation deployment file) which are a wrap of `hardhat-deploy::deploy`. If we really want to make this `hardhat-deploy`-compatible, the path of least resistance is to just remove the wrappers. If we want to make it easy to use, I'd prefer an opinionated `deployTUP` function.

PR #1686, opened 2025-08-25
chore(examples/solana): bump up anchor version

nazreen:

PR #1782, opened 2025-10-30
experiment: try removing

nazreen:


nazreen:
If CI here fails, and CI in #1686 passes, this proves that `RUSTUP_TOOLCHAIN=1.84.1` is needed

PR #1783, opened 2025-11-03
feat: use latest initiad & initia.js 

beer-1:
## Summary of Changes

- Allow usage with the latest Initiad release
- Update dependency to the latest initia.js
- Fix to correctly pass local decimal values when creating MoveOFTAdapter on Initia
    - In many cases, tokens on Initia have 0 decimals because the IBC transfer module does not support transmitting decimal metadata

### DIFF
--- a/packages/devtools-evm/src/signer/sdk.ts
+++ b/packages/devtools-evm/src/signer/sdk.ts
@@ -81,7 +81,7 @@ export class GnosisOmniSignerEVM<
             ethers,
             signerOrProvider: signer,
         }),
-        protected readonly apiKit = new SafeApiKit({ txServiceUrl: safeUrl, ethAdapter }),
+        protected readonly apiKit = new SafeApiKit({ txServiceUrl: safeUrl, apiKey: safeConfig.safeApiKey, chainId }),
         protected readonly safeSdkPromise: Safe | Promise<Safe> = Safe.create({
             ethAdapter,
             safeAddress: safeConfig.safeAddress!,
@@ -142,7 +142,7 @@ export class GnosisOmniSignerEVM<
 
         return safeSdk.createTransaction({
             safeTransactionData: transactions.map((transaction) => this.#serializeTransaction(transaction)),
-            options: { nonce },
+            options: { nonce: parseInt(nonce) },
             onlyCalls: true,
         })
     }


packages/devtools-move/cli/operations/evm-send.ts
```diff
@@ -66,6 +73,7 @@ async function sendOFT(args: any): Promise<MessagingFee> {
     const to = ethers.utils.hexZeroPad(args.to, 32)
     const amount = args.amount
     const minAmount = args.min_amount
+    const sendingNativeToken = args.native_token ? true : false
```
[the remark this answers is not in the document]
beer-1: fixed

nazreen:
@beer-1 , can you confirm what the main blocker that this PR is meant to solve?

I might ask you to separate out the client only changes, as the contract changes might take longer to approve.

beer-1:
Hey @nazreen,

There are three points
- Use latest initia.js to use coin type 60 instead of 118
- Allow to set local decimals when you create `OFTAdapter` on initia because IBC tokens do not have decimals information, so it will be created with 0 decimals always. 
- In object code deployment, we have added some interface and address computation method. so in current version of cli, we cannot deploy move contract without this fixes.

We already deployed the contract with this fixes locally, but good to merge this for future usages.

packages/devtools-move/tasks/move/utils/config.ts
```diff
@@ -410,3 +379,154 @@ export function isVersionLessThanOrEqualTo(installed: string, required: string):
...
+        let stdout = ''
+
+        childProcess.stdout?.on('data', (data) => {
+            stdout += data.toString()
+        })
+
+        childProcess.on('close', (code) => {
+            if (code === 0) {
+                const accountInfo = JSON.parse(stdout)
+                resolve(BigInt(accountInfo.info.sequence) + BigInt(2))
```
nazreen, line 480: @beer-1 what's the significance of the value `2` here?

beer-1: same reason with this https://github.com/initia-labs/initia/blob/0621b9de571ffde395da7671ad261cf1b4e32557/cmd/move/deploy.go#L110

so sequence number will be increased during this deployment tx. 

and in contract we are using sequence+1 as object seed [here](https://github.com/initia-labs/movevm/blob/dc215a4b7e8f4d113acee042b5e854b0efe21f5e/precompile/modules/initia_stdlib/sources/object_code_deployment.move#L114)

PR #1787, opened 2025-11-04
generate lockfiles script: throw if pnpm used is not v8

nazreen:
The `verify lockfiles` CI job expects lockfiles to be generated by pnpm v8 but in some cases a different version of pnpm might be used - leading to incorrect lockfiles being generated.

## Problem

Currently, running `pnpm lockfiles:generate` on a local machine may result in lockfiles that would be rejected by the `verify lockfiles` CI job. This can happen due to the generate lockfiles script using the local machine's version of `pnpm`, disregarding the `packageManager` version of pnpm from `package.json`. This results in lockfiles with `lockfileVersion: 9.0` whereas the CI job expects `lockfileVersion: 6.0`

## Changes

Validate that the pnpm version used matches the one set in `package.json`


scripts/generate-lockfiles.mjs
```diff
@@ -43,3 +52,25 @@ if (failures) {
...
 console.log('\nDone.');
+
+function getExpectedVersionFromPackageJson() {
+    const packageJson = JSON.parse(readFileSync(join(process.cwd(), 'package.json'), 'utf8'));
+    const packageManager = packageJson.packageManager;
+    if (!packageManager?.startsWith('pnpm@')) {
+        throw new Error(`Invalid packageManager: ${packageManager || 'missing'}. Expected format: "pnpm@x.y.z"`);
+    }
+    return packageManager.slice(5); // Remove "pnpm@" prefix
+}
```
tinom9, line 63: This assumes that we will always have fixed versions and no trailing hash (example here https://github.com/LayerZero-Labs/docs/blob/main/package.json#L102).

I think it's a safe assumption given the timeline of the repo though.

PR #1792, opened 2025-11-06
hyperliquid sdk updates

shankars99:
PR introduces changes to update the hyperliquid sdk

1. New quote token `USDH` on mainnet and testnet as default and known tokens in addition to `USDC` and `USDT0`

2. SDK `--network` flag defaults to `mainnet` when not passed. Aliases for `--network <m | t>` resolve to `mainnet/testnet`

3. `--ci` mode for `set-block` which bypasses the confirmation prompt in `wallet`.

4. Building off (3) updating the hyperliquid examples to use `skipPrompt` so that `npx hardhat lz:deploy` is non-blocking

### DIFF
--- a/.changeset/funny-turkeys-walk.md
+++ b/.changeset/funny-turkeys-walk.md
@@ -1,0 +1,6 @@
+---
+"@layerzerolabs/hyperliquid-composer": patch
+"@layerzerolabs/oft-hyperliquid-example": patch
+---
+
+sdk improvements - add USDH, ci mode for set-block, and default --network to mainnet


packages/hyperliquid-composer/src/types/constants.ts
```diff
@@ -21,20 +21,22 @@ export const ENDPOINTS = {
...
+        { tokenId: 0, name: 'USDC' },
+        { tokenId: 268, name: 'USDT0' },
+        { tokenId: 360, name: 'USDH' },
+    ],
+    TESTNET: [
+        { tokenId: 0, name: 'USDC' },
+        { tokenId: 1204, name: 'USDT0' },
+        { tokenId: 1452, name: 'USDH' },
+    ],
+} as const
```
tinom9, line 39: I'm guessing there's no API we can fetch these from?

lzErfan (approved):

packages/hyperliquid-composer/src/types/constants.ts
shankars99: none

PR #1794, opened 2025-11-07
feat: add multi-hop composer contract

tinom9:
### Description

- Add `MultiHopComposer` and `IMultiHopComposer` contracts to `oft-evm` package.


shankars99:
dont think we should be merging this version of the multihop contract. i can create a more generic implementation. also maybe create a new folder called "composers" for this could be better. will have to think about it

PR #1803, opened 2025-11-10
patch/non-evm-support

shankars99:
Updates core `@layerzerolabs/lz-definitions` packages from `3.0.75` to `3.0.145` which adds in more chains.


tinom9:
Why don't we update it in every package? There's more than shown in the picture.

image

We could create a pnpm catalog (https://pnpm.io/catalogs) for the internal packages (not `examples/` as we cannot use workspace-specific features there), to ensure we share the same external dependency version in every package.

tinom9 (approved):

PR #1805, opened 2025-11-12
Support Safe API keys

DanL0:
Safe will be making API keys mandatory soon.

https://docs.safe.global/core-api/how-to-use-api-keys

Example Safe config:
```
{
  "safeUrl": "https://api.safe.global/tx-service/sep/api",
  "safeAddress": "0x1329D6234df5EB52c1f6A2e31638c6ae3C0FCE6c",
  "safeApiKey": "XYZ"
}
```

Testing:
1. Clone devtools locally and checkout `safe-api-keys` branch
2. Go to examples/oft and prepare hardhat.config.ts with OFT on Sepolia and Avalanche Testnet
3. Wire OFTs, transfer delegate of Sepolia OFT to Sepolia Safe you control
4. Create Safe API key: https://developer.safe.global/api-keys
5. Add safeConfig to hardhat.config.ts, with safeApiKey equal to your key
6. In Safe Dashboard -> API keys -> edit your API key -> choose allowed IP only 192.168.1.1 -> press + button -> press Update
7. Change number of confirmations in layerzero.config.ts
8. Run wire command again with `--safe` flag
9. Your transactions on Sepolia should fail with "Your IP is not allowed"
10. Now edit Safe API key in Safe Dashboard again - remove IP whitelist
11. Re-run wire command, this time transactions should be submitted to Safe
12. Execute Safe transactions


packages/devtools-evm/package.json
```diff
@@ -37,7 +37,7 @@
     "test:jest": "jest --ci"
   },
   "dependencies": {
-    "@safe-global/api-kit": "^1.3.0",
+    "@safe-global/api-kit": "~4.0.0",
     "@safe-global/protocol-kit": "^1.3.0",
```
tinom9, line 41: I believe `@safe-global/protocol-kit` at the current version relies in contract deployments that may not be available for new chains.

There's no current need for `MultiSend` anywhere in our code I believe (we're passing the `onlyCalls: true` flag that bundles transactions with `MultiSendCallOnly` instead of `MultiSend`), so I'm leaning to think it's an initialization method that grabs all contracts for a given chain ID and throws if it doesn't find it.

This is outside of the PR scope, but we should fix it if we want to use it in new chains.

I tested with Unichain.

Image

tinom9 (dismissed):
Looking great! 🚀

nazreen:
Screenshot 2025-11-14 at 3 22 11 PM
I wanted to test but got stuck at creating a safe on Base Sepolia
Their indexers are behind by an hour and it doesn't seem like I can plug in my own RPC URL

packages/devtools-evm/package.json
DanL0: Devtools already support it:
```
safeConfig: {
                safeUrl: 'xyz',
                safeAddress: '0x1',
                contractNetworks: {
                    '9745': {
                        // the EVM chain id, not to be confused with LayerZero Endpoint ID.
                        multiSendAddress: '0x38869bf66a61cF6bDB996A6aE40D5853Fd43B526', // Multicall1.4.1 from https://berachain.notion.site/Furtim-Safe-Contracts-1826120abced804e912dd888e917ab64
                        multiSendCallOnlyAddress: '0x9641d764fc13c8B624c04430C7356C1C7C8102e2', // same as above
                    },
                },
            },
```

nazreen:
Tested, works as described!

## Proof of Test

When my IP address is not part of allowlist:
```
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ error               Error: Your IP address is not allowed                                                                                                                                          │
│ Endpoint            BASESEP_V2_TESTNET                                                                                                                                                             │
│ OmniAddress         0x6EDCE65403992e310A62460808c4b910D972f10f    
```

When my IP address is not restricted, setConfig txn popped up in the Safe and could be executed: https://sepolia.basescan.org/tx/0xa67d8b6fa233658c57080fac068107f769b4a77398397df716bec3889644aff5

nazreen (approved):

PR #1806, opened 2025-11-12
DEVREL-914 DEVREL-924 oapp-solana: update README + fix gen:api

nazreen:
closes DEVREL-914
closes DEVREL-926

## Changes

- fixed `pnpm gen:api` which broke after the upgrade of Anchor versions (IDL schema changed)
- deleted `exchangeIdlJson` which is no longer needed after the upgrade to Anchor v0.31.1 (it was needed to reconcile the differing IDL schemas)
- removed call to `moveGenEventFiles` as that's no longer needed
- given the above 2, deleted `exchange.ts` as it's no longer needed

## Proof of Test

Solana to Arbitrum: https://testnet.layerzeroscan.com/tx/4PQ591pcz6DbEntTmzoy16yckwgse7g9q5Ke3RkLR7EJdRHeogqbcdRhGcjR6dWe3LKg9P1cDC8fpYKnkniLkn2s

Arbitrum to Solana: 
https://testnet.layerzeroscan.com/tx/0x1e0e1d7b35e6d7708b85b893c02619c35df73dfc357750d8899113cc6049bfec

## Testing

Init

```
LAYERZERO_EXAMPLES_REPOSITORY_REF=solana-oapp-updates LZ_ENABLE_SOLANA_OAPP_EXAMPLE=1 npx create-lz-oapp@latest
```

Follow the README


examples/oapp-solana/lib/scripts/generate.ts
```diff
@@ -1,26 +1,23 @@
+import * as fs from 'fs'
 import * as path from 'path'
 
-import { IdlV00, rootNodeFromAnchor } from '@kinobi-so/nodes-from-anchor'
+import { Idl } from '@coral-xyz/anchor/dist/cjs/idl'
```
DanL0: import { Idl } from '@coral-xyz/anchor'

examples/oapp-solana/lib/scripts/generate.ts
```diff
@@ -1,26 +1,23 @@
...
+import { AnchorIdl, rootNodeFromAnchor } from '@kinobi-so/nodes-from-anchor'
 import { renderVisitor } from '@kinobi-so/renderers-js-umi'
 import { createFromRoot } from 'kinobi'
 
-import { exchangeIDLJson, moveGenEventFiles } from './exchange'
-
 async function sleep(ms: number): Promise<void> {
```
DanL0: unused

examples/oapp-solana/README.md
```diff
@@ -279,3 +313,19 @@ pnpm test
...
+## Modifying for your use case
+
+After going through the above and having successfully sent cross-chain messages using the template, you might want to start customizing the Solana program to fit your actual use case.
+
+First, ensure you have built the program after making the necessary updates, to ensure the latest changes to the IDL are reflected.
+
+Next, run:
+
+```bash
+pnpm gen:api
```
DanL0, line 325: would:
```
MYOAPP_ID=AQWUXAcCXoapzJ2qUnkkQabyhUZ1NmQAseBWXHkSvixw pnpm gen:api
```

be more flexible and easy to use for users (telling them they need to replace it too explicitly)?

DanL0 (changes requested):

examples/oapp-solana/lib/scripts/generate.ts
nazreen: fixed

examples/oapp-solana/lib/scripts/generate.ts
nazreen: fixed

examples/oapp-solana/README.md
nazreen: this is a good idea, thanks!

EDIT: this might not be necessary

nazreen: Just to clarify,

(buckle in)

it's not actually necessary to replace the address in the generate script 

(the following line:)
```
anchorIdl.address = 'HFyiETGKEUS9tr87K1HXmVJHkqQRtw8wShRNTMkKKxay'
```

Is purely 'cosmetic', because

that affects lines such as in `initStore`:

```
  const programId = context.programs.getPublicKey(
    'myOapp',
    'HFyiETGKEUS9tr87K1HXmVJHkqQRtw8wShRNTMkKKxay'
  );
 ```

But, (the program client is written in such a way that) every time the SDK needs to be used, it has to be instantiated in the following way:

```
const myoappInstance: myoapp.MyOApp = new myoapp.MyOApp(publicKey(programId))
```

the `programId` public key variable is actually mandator
this is the MyOApp constructor in `lib/client/myoapp.ts`:

```
    constructor(
        public readonly programId: PublicKey,
```

What I found is that, the line

```
anchorIdl.address = 'HFyiETGKEUS9tr87K1HXmVJHkqQRtw8wShRNTMkKKxay'
```

is cosmetic, because the address in the IDL, is actually like

```
"address": "anchorlangsolanaprogrampubkeyPubkeynewfromarrayprogramidfromenvMYOAPPIDn41NCdrEvXhQ4mZgyJkmqYxL6A1uEmnraGj31UJ6PsXd3",
```

Which results from how the IDL generation step can only ready static values (and not env vars), which is an issue because we declare the ID like so

```
declare_id!(anchor_lang::solana_program::pubkey::Pubkey::new_from_array(program_id_from_env!(
    "MYOAPP_ID",
    "41NCdrEvXhQ4mZgyJkmqYxL6A1uEmnraGj31UJ6PsXd3" // It's not necessary to change the ID here if you are building using environment variable
)));
```

Comment (nazreen) [reply examples/oapp-solana/README.md:325]: Ok, I think 

```
anchorIdl.address = 'HFyiETGKEUS9tr87K1HXmVJHkqQRtw8wShRNTMkKKxay'
```

can just be updated to 

```
anchorIdl.address = ''
```

to tidy up. I think having a stock address is more confusing than having one

Comment (nazreen) [reply examples/oapp-solana/README.md:325]: Setting the address to an empty string is in line with the OFT SDK in the monorepo

```
// Instruction.
export function lzReceiveTypes(
  context: Pick<Context, 'programs'>,
  input: LzReceiveTypesInstructionAccounts & LzReceiveTypesInstructionArgs
): TransactionBuilder {
  // Program ID.
  const programId = context.programs.getPublicKey('oft', '');
```

**Comment by DanL0** [comment]
For anyone wanting to test themselves:
```
LAYERZERO_EXAMPLES_REPOSITORY_REF=solana-oapp-updates LZ_ENABLE_SOLANA_OAPP_EXAMPLE=1 npx create-lz-oapp@latest
```

DanL0 (approved):

PR #1807, opened 2025-11-12
verify-contract upgrade to V2 and improvements (#1844)

jadijadi:
- Migration to API V2
- Moving hardcoded configs to a Yaml file


packages/verify-contract/src/common/etherscan.ts
```diff
@@ -178,9 +179,22 @@ const isAlreadyVerifiedResult = (result: string | null | undefined): boolean =>
...
     try {
-        const response = await got(apiUrl, {
+        // For Etherscan API v2, include chainid in the URL
+        // Only add chainid for Etherscan v2 URLs, not for custom explorers
+        const url = new URL(apiUrl)
+        const isEtherscanV2 = apiUrl.includes('api.etherscan.io/v2')
+
+        if (chainId !== undefined && isEtherscanV2) {
+            url.searchParams.set('chainid', String(chainId))
+        }
```
tinom9: Does this block need to be inside of the try-catch? `checkGuid` seems to handle it outside.

packages/verify-contract/src/common/etherscan.ts
```diff
@@ -178,9 +179,22 @@ const isAlreadyVerifiedResult = (result: string | null | undefined): boolean =>
 
 const isApiRateLimitedResult = (result: string | null | undefined): boolean => !!result?.match(/rate/)
 
-const submitRequest = async (apiUrl: string, request: SubmitForVerificationRequest): Promise<ScanResponse> => {
+const submitRequest = async (
+    apiUrl: string,
+    request: SubmitForVerificationRequest,
+    chainId?: number
```
tinom9: Do we not need the `apiKey` here as we do in `checkGuid`?

packages/verify-contract/README.md
```diff
@@ -16,129 +16,355 @@
...
+
+**Note:** Environment variable names are case-insensitive and support both hyphenated and underscored network names:
+- `SCAN_API_KEY_ethereum` or `SCAN_API_KEY_ETHEREUM`
+- `SCAN_API_KEY_base-sepolia` or `SCAN_API_KEY_BASE_SEPOLIA`
+
+### Default Network Configuration
+
+The package is preconfigured with scan API URLs and chain IDs for well-known networks.
+
+#### Etherscan v2 Compatible Networks
```
tinom9: This is a nit, but I'd avoid having the full list here, so we don't need to update it if we add new networks to the list.

I'd replace if with a couple of examples and a link to the source of truth (the networks file).

packages/verify-contract/src/common/networks.yaml
tinom9, line 1: Also a nit, but is there any specific reason why this is a `yaml` and not a `ts` or `json` file?

It seems like it's not very consistent with the rest of the codebase.

packages/verify-contract/src/common/etherscan.ts
jadijadi: You are right. Not needed. removed it.

packages/verify-contract/src/common/etherscan.ts
jadijadi: Right. will do this. Better Consistency and API compatibility.

packages/verify-contract/README.md
jadijadi: good point. Agree. Will mention the API2 URL (as our source) and after showing a few chains as an example, will point tot he `src/common/networks.yaml` file.

packages/verify-contract/src/common/networks.yaml
jadijadi: I went with YAML because it’s more human readable and easier to edit. And this file will be updated -mostly- by people.

But I totally get your point about using ts/json instead. It’s more consistent and avoids adding a new dependency. Was going to disagree with this one first.. but after thinking aloud (on keyboard) I think I should use ts/json. 

- If no one’s used YAML in this project before, I’m probably not the best person to introduce it
- Consistency with the rest of the project
- The future edits will be AI assisted and YAML do not have a huge advantage over json/ts.

jadijadi:
Dear @DanL0 & @tinom9, Thanks for the reviews. addressed all comments in the latest commits. CI is green.

packages/verify-contract/src/common/networks.ts
tinom9, line 1: I believe we should add aliases for every network that we have with a different slug in [`@layerzerolabs/lz-definitions`](https://github.com/LayerZero-Labs/monorepo/blob/main/packages/layerzero-definitions/src/enums/chain-evm.ts).

packages/verify-contract/tsup.config.ts
tinom9, line 1: Are the diffs in this file still required?

packages/verify-contract/README.md
```diff
@@ -16,129 +16,315 @@
...
+SCAN_CHAIN_ID_polygon=137
+
+# Browser URL for displaying verified contract links
+SCAN_BROWSER_URL_ethereum=https://etherscan.io
+SCAN_BROWSER_URL_polygon=https://polygonscan.com
+```
+
+**Note:** Environment variable names are case-insensitive and support both hyphenated and underscored network names:
+- `SCAN_API_KEY_ethereum` or `SCAN_API_KEY_ETHEREUM`
+- `SCAN_API_KEY_base-sepolia` or `SCAN_API_KEY_BASE_SEPOLIA`
```
tinom9: Is `SCAN_API_KEY_base-sepolia` ever valid?

packages/verify-contract/src/common/etherscan.ts
```diff
@@ -178,31 +179,59 @@ const isAlreadyVerifiedResult = (result: string | null | undefined): boolean =>
...
+    if (apiKey) {
+        url.searchParams.set('apikey', apiKey)
     }
+
+    const response = await got(url.toString(), {
+        method: 'POST',
+        form: request,
+        headers: {
+            'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8',
+        },
```
tinom9: Is this required or just a nice-to-have?

packages/verify-contract/src/common/networks.ts
jadijadi: Tried to review all and add the missing ones.

packages/verify-contract/tsup.config.ts
jadijadi: Agree. reverted to the initial state.

packages/verify-contract/src/common/etherscan.ts
jadijadi: Checked. This is already sent via the 'got' library. removed it.

packages/verify-contract/README.md
jadijadi: it can be valid in an .env file. but for maximum compatibility, added a note and used and proposed uppercased, underscored version only.

tinom9 (approved):
Great work! 🔝

jadijadi:
> Code looks good. Happy path works too.
> 
> I have tested three commands:
> 
> ```
> node ../../packages/verify-contract/dist/cli.js target --network avalanche-testnet --deployments ./deployments --api-url https://api.routescan.io/v2/network/testnet/evm/43113/etherscan
> ```
> 
> ```
> node ../../packages/verify-contract/dist/cli.js target --network avalanche-testnet --deployments ./deployments --api-key XYZ
> ```
> 
> ```
> node ../../packages/verify-contract/dist/cli.js target --network sepolia-testnet --api-key XYZ --deployments ./deployments
> ```
> 
> Result:
> 
> * https://testnet.routescan.io/address/0xa5F4E778826fbD20b71b3C38a4B58b608D91A129/contract/43113/code
> * https://sepolia.etherscan.io/address/0x78C148083697a4132289eA8D66B345690d2e3CCC#code
> * https://docs.etherscan.io/api-reference/endpoint/getsourcecode 43113 0xa5F4E778826fbD20b71b3C38a4B58b608D91A129
> 
> One thing, please add changeset before merging! (npx changeset)

added the changeset. Thanks for the review.

DanL0 (approved):

PR #1808, opened 2025-11-12
feat: optimized rate limiter

tinom9:
### Description

- Reintroduced and updated rate limiter with:
  - Storage struct packing for optimization and avoiding edge-case overflows.
  - New `_resetRateLimits` internal function to reset inflight state.
  - `getAmountCanBeSent` visibility update from `external` to `public`.


TRileySchwarz (dismissed):

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
shankars99, line 1: i wish we can move this to `packages/oft-evm` since this contract is tied to transfer of tokens

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
```diff
@@ -0,0 +1,264 @@
...
+ *            |                        ✧                              *
+ *            |                         ✧                                  *
+ *            |                          ✧                                     *
+ *            |                           ✧                                        *
+ *        0 - +---|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----> Time
+ *            0   5    10   15   20   25   30   35   40   45   50   55   60   65   70   75   80   85   90 (seconds)
+ *            [  Initial 30 Second Window  ]
+ *                          [ --------------- Extended 60 Second Window --------------- ]
+ */
+abstract contract RateLimiter {
```
shankars99, line 103: interface would be nice. also this _doesnt_ really need to be `abstract` right?

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * @notice Get the current amount that can be sent to this destination endpoint id for the given rate limit window.
+     * @param _dstEid The destination endpoint id.
+     * @return currentAmountInFlight The current amount that was sent.
+     * @return amountCanBeSent The amount that can be sent.
+     */
+    function getAmountCanBeSent(
+        uint32 _dstEid
+    ) external view virtual returns (uint256 currentAmountInFlight, uint256 amountCanBeSent) {
```
shankars99: can just be `public` if for some reason the same contract wants to read it they can by just passing the `dstEid` without having to copy over the `RateLimit storage` read line

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
```diff
@@ -0,0 +1,264 @@
...
+     * @notice Rate Limit Configuration struct.
+     * @param dstEid The destination endpoint id.
+     * @param limit This represents the maximum allowed amount within a given window.
+     * @param window Defines the duration of the rate limiting window.
+     */
+    struct RateLimitConfig {
+        uint32 dstEid;
+        uint192 limit;
+        uint64 window;
+    }
```
shankars99, line 128: nit: would rewrite as 

uint192
uint64
uint32

so that the last slot isn't a complete one and the ones before that are used up

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
shankars99: improvement nit

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
shankars99: nit for standards

tinom9: I'd tackle the interface in follow-up version.

Abstract contract makes sense to me for the same reason that OZ's ERC20 is abstract.

I don't see anyone deploying the contract as is (no functionality to rate-limit).

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
tinom9: That'd be nicer for future upgrades, I agree, but it's just 8 bytes. I'd disregard it for now.

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
shankars99: true sounds good

packages/oapp-evm/contracts/oapp/utils/RateLimiter.sol
tinom9: Updated.

shankars99 (approved):

PR #1817, opened 2025-11-19
DEVREL-945 fix default ALTs not being used

nazreen:
## Problem

When `--address-lookup-tables` isn't provided, `sendOFT.ts` defaults `addressLookupTables` to `[]`. In `sendSolana.ts`, the condition `addressLookupTables ? ...` is truthy for an empty array, so it maps an empty array instead of using the default lookup table from `DEFAULT_LOOKUP_TABLE_ADDRESS`. This caused the lookup table addresses array to be empty when it should contain the default lookup table address.

## Changes

- Updated the condition in `sendSolana.ts` to check `addressLookupTables && addressLookupTables.length > 0` instead of `addressLookupTables ? ...`
- Ensures the default lookup table is used when no lookup tables are provided via CLI, instead of returning an empty array

## Test

You can test the following on `main`, before the fix is applied. 
In `sendSolana.ts`, add a `console.log` below where the ALTs are evaluated:

```
    const lookupTableAddresses = addressLookupTables
        ? addressLookupTables.map((addr) => publicKey(addr))
        : [(await getDefaultAddressLookupTable(connection, umi, srcEid)).lookupTableAddress]
   console.log(lookupTableAddresses)
 ```
 
 Run a send (from Solana) command like normal:
 ```
 npx hardhat lz:oft:send --src-eid 40168 --dst-eid 40231 --to <EVM_RECIPIENT>  --amount 1
 ```
 
 You'll see that what's printed is an empty array. This is **not expected** as we expect the default ALT to be used.

### DIFF
--- a/.changeset/chatty-crabs-sin.md
+++ b/.changeset/chatty-crabs-sin.md
@@ -1,7 +1,0 @@
----
-"@layerzerolabs/devtools-evm-hardhat": patch
-"@layerzerolabs/oft-upgradeable-example": patch
-"@layerzerolabs/oft-alt-example": patch
----
-
-patch: make upgradeable deploy script similar to hardhat-deploy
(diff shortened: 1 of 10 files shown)


ravinagill15 (approved):
Can add a changeset

PR #1840, opened 2025-11-25
DEVREL-782 Feat: support native composer in ovault example

St0rmBr3w:
adds updates to README and code change to use the native OVault native composer

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -52,7 +52,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: ['javascript-typescript', 'rust']
+        language: ['javascript-typescript']
 
     # Skip autobuild - CodeQL can analyze source code directly
     # No need to compile or install dependencies for static analysis
--- a/bin/env
+++ b/bin/env
@@ -10,5 +10,6 @@
 # (The "." will execute this inside the current shell instead of creating a child one)
 
 set -a
-source ${1:-.env}
+# shellcheck source=/dev/null
+source "${1:-.env}"
 set +a
--- a/packages/protocol-devtools-solana/CHANGELOG.md
+++ b/packages/protocol-devtools-solana/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @layerzerolabs/protocol-devtools-solana
 
+## 8.0.6
+
+### Patch Changes
+
+- c45bec8: shell linting
+
 ## 8.0.5
 
 ### Patch Changes
--- a/packages/protocol-devtools-solana/bin/test
+++ b/packages/protocol-devtools-solana/bin/test
@@ -3,5 +3,5 @@
 if [ -z "${LZ_DEVTOOLS_ENABLE_SOLANA_TESTS}" ]; then
     echo 'Solana tests can be enabled by setting LZ_DEVTOOLS_ENABLE_SOLANA_TESTS environment variable to a non-empty value'
 else
-    jest --ci $@
+    jest --ci "$@"
 fi
--- a/packages/protocol-devtools-solana/package.json
+++ b/packages/protocol-devtools-solana/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@layerzerolabs/protocol-devtools-solana",
-  "version": "8.0.5",
+  "version": "8.0.6",
   "description": "Utilities for LayerZero Solana protocol programs",
   "repository": {
     "type": "git",
--- a/packages/ua-devtools-solana/CHANGELOG.md
+++ b/packages/ua-devtools-solana/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @layerzerolabs/ua-devtools-solana
 
+## 8.0.7
+
+### Patch Changes
+
+- c45bec8: shell linting
+- Updated dependencies [c45bec8]
+  - @layerzerolabs/protocol-devtools-solana@8.0.6
+
 ## 8.0.6
 
 ### Patch Changes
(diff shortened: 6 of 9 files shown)


examples/ovault-evm/contracts/MyOVaultComposerERC20.sol
```diff
@@ -7,12 +7,12 @@ import { VaultComposerSync } from "@layerzerolabs/ovault-evm/contracts/VaultComp
  * @title MyOVaultComposer
  * @notice Cross-chain vault composer enabling omnichain vault operations via LayerZero
  */
-contract MyOVaultComposer is VaultComposerSync {
+contract MyOVaultComposerERC20 is VaultComposerSync {
```
ravinagill15, line 10: nit: may want to add Sync in the contract name as well. That is the standard we followed [in monorepo-internal](https://github.com/LayerZero-Labs/monorepo-internal/blob/main/apps/yield-xyz-share-oft-app/contracts/evm/ovault/contracts/YieldXyzOVaultComposerSync.sol)

examples/ovault-evm/contracts/MyOVaultComposerNative.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.22;
+
+import { VaultComposerSyncNative } from "@layerzerolabs/ovault-evm/contracts/VaultComposerSyncNative.sol";
+
+/**
+ * @title MyOVaultComposerNative
+ * @notice Cross-chain vault composer enabling omnichain vault operations via LayerZero
+ */
+contract MyOVaultComposerNative is VaultComposerSyncNative {
```
ravinagill15, line 25: nit: same concept, may wanna add sync to be extra clear that async vaults are not supported

ravinagill15 (approved):

examples/ovault-evm/contracts/MyOVaultComposerNative.sol
St0rmBr3w: I don't think this matters as much, very few people are looking at this from the POV of synchronous vs async yet.

shankars99: would be good to maintain it going forward since async could be a thing we build out anytime soon then need to tell people to update in their head?

examples/ovault-evm/contracts/MyOVaultComposerERC20.sol
shankars99: second this

examples/ovault-evm/contracts/MyOVaultComposerNative.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.22;
+
+import { VaultComposerSyncNative } from "@layerzerolabs/ovault-evm/contracts/VaultComposerSyncNative.sol";
+
+/**
+ * @title MyOVaultComposerNative
+ * @notice Cross-chain vault composer enabling omnichain vault operations via LayerZero
+ */
```
shankars99, line 24: prolly comment that it defaults to using `WETH9` interface to convert Native to Wrapped-Native. 
This works with ETH->WETH and HYPE->WHYPE but there could be chains where the `wNative` contract does not follow `WETH9`. 

In that case developers should override `function lzCompose(...args)` to something like 

```
    /**
     * @dev Reduction of ComposerNative into ComposerBase by wrapping ETH into WETH
     * @dev All internal logic handles WETH as the asset token making deposit symmetric to redemption
     * @dev The native token used here was sent during lzReceive from the Pool
     @dev Inherits entry authorization from parent contract
     */
    function lzCompose(
        address _composeSender, // The OFT used on refund, also the vaultIn token.
        bytes32 _guid,
        bytes calldata _message, // expected to contain a composeMessage = abi.encode(SendParam hopSendParam,uint256 minMsgValue)
        address _executor,
        bytes calldata _extraData
    ) public payable override virtual {
        /// @dev Wrap ETH received during lzReceive into WETH
        /// @dev Conversion happens here to avoid running out of gas in the receive function
        if (_composeSender == ASSET_OFT) ICustomNative(ASSET_ERC20).customFunction{ value: _message.amountLD() }();

        /// @dev Since lzCompose is public, the msg.value called to pay the tx Fee is automatically forwarded
        /// @dev VaultComposerSync and not native because the parent calls `super.` which is `VaultComposerSync`
        VaultComposerSync.lzCompose(_composeSender, _guid, _message, _executor, _extraData);
    }
    ```

File: examples/ovault-evm/README.md [hunk 4]
```diff
@@ -45,6 +46,34 @@ OVault extends the ERC-4626 tokenized vault standard with LayerZero's omnichain
...
+- Automatically wraps native tokens (ETH) into WETH before depositing to the vault
+- Necessary because ERC-4626 vaults only support ERC20 tokens
+- Transparent to end users - they send native tokens and receive shares
+
+**Which Composer Should You Use?**
+
+The composer type depends on your asset OFT's underlying token:
+
+- If `assetOFT.token()` returns an ERC20 address → use `MyOVaultComposerERC20`
+- If `assetOFT.token()` returns `address(0)` (native token) → use `MyOVaultComposerNative`
```
shankars99, line 81: also document that we have constructor overridable functions for initializing asset and share tokens if they do not adhere to our default patterns?

`_initializeAssetToken(...args)` and `_initializeShareToken(...args)`

examples/ovault-evm/contracts/MyOVaultComposerNative.sol
```diff
@@ -0,0 +1,23 @@
...
+contract MyOVaultComposerNative is VaultComposerSyncNative {
+    /**
+     * @notice Creates a new cross-chain vault composer where the vault asset is the chain's native asset
```
shankars99, line 27: also document that `OFT.token()` != `Vault.asset()` because vault asset needs to be an `ERC20` but there could exist a custom `ERC4626` vault that takes in ETH. In that case you would need a custom OVaultComposerNative?

lzJxhn (approved):

.changeset/quiet-toes-occur.md
ravinagill15, line 1: 2 changesets?

ravinagill15 (dismissed):

ravinagill15 (approved):

PR #1851, opened 2025-11-28
DEVREL-955 bump versions of monorepo packages

nazreen:


nazreen:
> 
> The job failed during the pnpm fetch step, while installing native Node.js modules (such as utf-8-validate, bufferutil, bigint-buffer) on ARM64/node@20.19.5. The log shows a SyntaxError in Python while running node-gyp, specifically:
> 
> SyntaxError: unterminated string literal (detected at line 566) in gyp_main.py
> This resulted in a failure to build utf-8-validate, leading to an ELIFECYCLE error and the pnpm fetch step exiting with code 1.
>

St0rmBr3w (approved):
LGTM!

PR #1853, opened 2025-11-29
Avoid node gyp build

nazreen:


nazreen:
build failure turns out is failing on main too
https://github.com/LayerZero-Labs/devtools/actions/runs/19776463219/job/56669775441

nazreen:
no longer needed as #1851 now passes CI

PR #1864, opened 2025-12-08
feat: codeql workflow for ts and rust

shankars99:
CodeQL extension does not run on fork prs into devtools. This workflow yaml should effectively replace the extension.

Also took the liberty to add in:
1. `Dockerfile` scanning
2. `Shell script` scanning 

ontop of `codeql`'s `js`-`ts` and `rust`

It also runs weekly scans on `main`

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -2,7 +2,7 @@
 #  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ / / \ \ / / \ \ / / \
 # `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-'   `-`-'   `-`-'   `-`-'
 #
-#                       CodeQL Security Analysis
+#                       CodeQL
 #
 #  This workflow replaces the GitHub CodeQL extension to support fork PRs.
 #  The extension doesn't trigger on fork PRs due to security restrictions.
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,135 @@
...
+jobs:
+  analyze:
+    name: CodeQL Analysis
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+
+    permissions:
+      contents: read
+      security-events: write
+      actions: read
```
ravinagill15: nit: seems redundant since we define the same permissions at the top level

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v4
+        with:
+          category: "/language:${{ matrix.language }}"
+
+  dockerfile:
+    name: Dockerfile Lint
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    needs: analyze
```
ravinagill15, line 82: just want to confirm this is expected -- if codeql fails, dockerfile and shellcheck checks won't run i believe

shankars99: yep intended

.github/workflows/codeql.yaml
shankars99: yep. removing them

St0rmBr3w (approved):
lgtm

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,154 @@
...
+    branches:
+      - main
+      - develop
+  schedule:
+    # Run weekly security scans every Monday at midnight UTC
+    - cron: '0 0 * * 1'
+
+# Cancel in-progress runs when a new commit is pushed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
tinom9, line 36: I think this will only cancel in-progress runs in a same PR.

I'm guessing it's ok.

shankars99: yeah thats what we wanna do

PR #1867, opened 2025-12-08
DEVREL-1009 fix: solana enforced options setter logic

St0rmBr3w:
### In this PR

Fixes a bug in the Solana OFT SDK where `setEnforcedOptions` incorrectly resets enforced options for msgTypes that aren't being updated.

**Problem:**

The wiring configuration (`configureEnforcedOptions` in `packages/ua-devtools/src/oapp/config.ts`) only passes msgTypes that differ from the current on-chain state to `setEnforcedOptions`. However, the Solana SDK's `setEnforcedOptions` defaulted any missing msgType to empty options rather than preserving the existing on-chain value. Since `_setPeerEnforcedOptionsIx` writes both `send` and `sendAndCall` atomically, this caused unintended overwrites.

This resulted in an infinite toggle:
1. Run 1: msgType 2 differs → SDK sets msgType 2, resets msgType 1 to empty
2. Run 2: msgType 1 now differs → SDK sets msgType 1, resets msgType 2 to empty
3. Repeat forever

**Solution:**

When a msgType isn't included in the update array, fetch its current on-chain value instead of defaulting to empty options. The implementation only makes RPC calls when necessary (i.e., when that msgType isn't being updated).

**Before:**
```typescript
const sendOption = optionsByMsgType.get(MSG_TYPE_SEND) ?? emptyOptions
const sendAndCallOption = optionsByMsgType.get(MSG_TYPE_SEND_AND_CALL) ?? emptyOptions
```

**After:**
```typescript
const sendOption = optionsByMsgType.has(MSG_TYPE_SEND)
    ? optionsByMsgType.get(MSG_TYPE_SEND)!
    : Options.fromOptions(await this.getEnforcedOptions(eid, MSG_TYPE_SEND)).toBytes()

const sendAndCallOption = optionsByMsgType.has(MSG_TYPE_SEND_AND_CALL)
    ? optionsByMsgType.get(MSG_TYPE_SEND_AND_CALL)!
    : Options.fromOptions(await this.getEnforcedOptions(eid, MSG_TYPE_SEND_AND_CALL)).toBytes()
```

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,154 @@
+#   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-.   .-.-  .-.-.   .-.-.   .-.-
+#  / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ \ / / \ / / \ \ / / \ \ / / \
+# `-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-`-'   `-'   `-`-'   `-`-'   `-`-'
+#
+#                       CodeQL
+#
+#  This workflow replaces the GitHub CodeQL extension to support fork PRs.
+#  The extension doesn't trigger on fork PRs due to security restrictions.
+#  This workflow uses the pull_request event which works for all PRs.
+#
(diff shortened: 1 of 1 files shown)


DanL0 (approved):
given very specific way how this SDK is laid out, it should work and not have a huge impact on the performance, however there probably is a way to avoid a new RPC call to fetch enforced options, given it already happens in configureOApp; however overriding that might not be simple given devtools architecture

shankars99 (approved):

PR #1944, opened 2026-06-22
fix(protocol-devtools): treat explicitly-empty ULN config values as NIL sentinels

St0rmBr3w:
## What does this PR do?
Makes the devtools ULN302 serializers distinguish an explicitly-empty config field from an omitted one, matching the protocol contract's sentinel semantics:

- `confirmations: 0n` → `NIL_CONFIRMATIONS` (`type(uint64).max`)
- `optionalDVNs: []` → `NIL_DVN_COUNT` (`0xff`), matching the existing `requiredDVNs: []`
- an omitted field still serializes to `0` (inherit the on-chain default)

To support this, `Uln302UlnConfig`/`UlnReadUlnConfig` now carry `optionalDVNCount` (and `UlnReadUlnConfig` also carries `requiredDVNCount`) so a stored sentinel round-trips through the config diff. The on-chain read path no longer re-applies the empty→NIL mapping, keeping `hasAppUlnConfig` idempotent. The library-wide DEFAULT config still serializes literal values (it rejects NIL on-chain). On Solana, `confirmations` is now encoded as a `BN` so the `u64` NIL sentinel survives without precision loss (the previous `Number(...)` cast was lossy). Covers EVM (uln302 + ulnRead) and Solana.

## Why is this change needed?
Previously an explicitly-empty field (`confirmations: 0`, `optionalDVNs: []`) was treated identically to an omitted one — both fell back to the protocol default. An integrator pinning "zero confirmations" or "no optional DVNs" silently got the default instead, a correctness and security gap.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / Tech debt
- [ ] Docs update
- [ ] Chore
- [ ] Performance improvement
- [ ] Other

## How was this tested?
- [x] Unit tests added/updated
- [ ] Integration tests pass
- [x] Manual testing performed

protocol-devtools-evm: 70 tests pass (added confirmations/optionalDVNs matrices + `hasAppUlnConfig` round-trip/idempotency tests). Solana unit tests added for the serializer NIL behavior and the `SetConfigSchema` BN precision (11 pass). metadata-tools: 39 tests + 9 snapshots unchanged. All three packages build (typecheck) and lint clean. Solana mainnet-RPC integration tests are CI-disabled; the two ULN-config ones were updated to the new semantics.

## Impact/Screenshots
N/A

## Checklist
- [x] PR is scoped and not too large
- [x] Self-review completed
- [x] Code follows style guidelines
- [x] Tests added for new functionality
- [x] Documentation updated (changeset with migration note)
- [x] Hard-to-understand code is commented

> ⚠️ **Migration (major):** if you wrote `confirmations: 0` or `optionalDVNs: []` expecting the default, **omit the field** instead — an explicit empty value now pins literal zero/none. For `confirmations` this means zero block confirmations (security-relevant).

## AI Provenance
- **Model**: Opus 4.8 (1M context)
- **Tool**: Claude Code
- **Prompt**: "The devtools repo has a critical bug — if a config is passed empty it falls back to defaults instead of the NIL sentinels. Only an omitted field should be a default fallback; an explicitly-empty DVNs/confirmations should be the NIL sentinel. Investigate and plan, then implement."
- **Iterations**: Plan mode → implementation across one Claude Code session


packages/ua-devtools-evm-hardhat/src/oapp/typescript/typescript.ts
```diff
@@ -283,31 +289,61 @@ export const creatExecutorConfig = ({ maxMessageSize, executor }: Uln302Executor
...
-        ),
-        factory.createPropertyAssignment(
-            factory.createIdentifier(OPTIONAL_DVN_THRESHOLD),
-            factory.createNumericLiteral(optionalDVNThreshold)
-        ),
-    ])
+        )
+    }
+
+    const optionalDVNsToEmit = dvnsFromCount(optionalDVNCount, optionalDVNs)
```
tinom9, line 328: `tests/ua-devtools-evm-hardhat-test/test/task/oapp/config.init.test.ts:239` and `read.config.init.test.ts:245` seem to be broken now that the generator omits `optionalDVNs` and `optionalDVNThreshold` when the value is 0?

packages/protocol-devtools/src/uln302/constants.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Sentinel values for ULN302 configuration, matching the on-chain contract (`UlnBase.sol`).
+ *
+ * A config field left at `0` falls back to the default config, so these sentinels are used
+ * to pin a literal zero/none instead:
+ *
+ * - `NIL_DVN_COUNT` pins "no DVNs" (the accompanying DVN array must be empty).
+ * - `NIL_CONFIRMATIONS` pins "zero confirmations".
+ */
+export const NIL_DVN_COUNT = (1 << 8) - 1 // type(uint8).max = 255
```
tinom9, line 10: This seems to duplicate `metadata-tools` and `devtools-move` utils, should we drop any now?

examples/oft/layerzero.config.ts
```diff
@@ -36,6 +36,10 @@ const pathways: TwoWayConfig[] = [
     [
         baseContract, // Chain A contract
         arbitrumContract, // Chain B contract
+        // DVN config: [ requiredDVN[], [ optionalDVN[], optionalDVNThreshold ] ]
+        //   - requiredDVNs: every listed DVN must verify each message — set these explicitly.
```
tinom9, line 40: Should we update `examples/oft/tasks/simple-workers-mock/utils/setSendConfig.ts:76` and `setReceiveConfig.ts:64` as well?

tinom9 (commented):
Checked that EVM and Solana empty optional DVNs now correctly resolve to the NIL sentinels. 👍

packages/protocol-devtools/src/uln302/constants.ts
St0rmBr3w: I'd like to avoid for now since this involves adding `@layerzerolabs/protocol-devtools` as a brand-new direct dependency to both packages.

tinom9 (dismissed):

TRileySchwarz (approved):

tinom9 (approved):

tinom9 (approved):

TRileySchwarz (approved):

PR #1947, opened 2026-07-21
OAPP-2067 fix(ua-devtools): skip already-pinned libraries with areBytes32Equal

St0rmBr3w:
## What does this PR do?

Fixes wire idempotency for message libraries so already-pinned `setSendLibrary` / `setReceiveLibrary` configs are not re-proposed.

`configureSendLibraries` / `configureReceiveLibraries` (and the oapp-read equivalents) now compare on-chain vs configured library addresses with `areBytes32Equal` instead of `toLowerCase()` / strict `===`. Checksum or padding mismatches no longer produce a second `set*Library` that reverts with `LZ_SameValue`.

Keeps the existing `!isDefaultLibrary` gate so "lock defaults" behavior is unchanged: when the OApp is still on the endpoint default and config matches the resolved default ULN, the first wire still pins storage.

## Why is this change needed?

After libraries are already set on-chain, fragile address equality caused tooling to re-propose the same library. EndpointV2 rejects that with `LZ_SameValue`, breaking idempotent re-wires.

Note: this is unrelated to #1944 (ULN NIL sentinel pinning). That PR did not touch message-lib skip logic.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / Tech debt
- [ ] Docs update
- [ ] Chore
- [ ] Performance improvement
- [ ] Other

## How was this tested?
- [x] Unit tests added/updated
- [ ] Integration tests pass
- [ ] Manual testing performed

Added hardhat regression tests that pin send/receive/read libraries, then re-run configure with lowercase config addresses and expect `[]` transactions. Package build + lint for `@layerzerolabs/ua-devtools` pass locally; full hardhat suite left to CI.

## Impact/Screenshots

N/A

## Checklist
- [x] PR is scoped and not too large
- [x] Self-review completed
- [x] Code follows style guidelines
- [x] Tests added for new functionality
- [x] Documentation updated
- [x] Hard-to-understand code is commented

## AI Provenance
- **Model**: Cursor Grok 4.5
- **Tool**: Cursor
- **Prompt**: Investigate Keno's report that old tooling re-proposes already-set setSendLibrary/setReceiveLibrary (likely from #1944), identify the SameValue bug, propose an idempotency fix, then create a PR
- **Iterations**: Plan mode investigation → agent implementation + PR in one follow-up turn

Made with [Cursor](https://cursor.com)


St0rmBr3w:
## Checksum canary (proved against main-style skip logic)

Added `packages/ua-devtools/test/oapp/library-checksum-canary.test.ts` (no Docker).

**Result when temporarily restoring main's skip predicates:**
- `configureReceiveLibraries` with checksummed on-chain + lowercase config → **FAILS** (proposes `setReceiveLibrary` → would hit `LZ_SameValue`)
- `configureSendLibraries` with the same casing mismatch → **passes** on main already (`toLowerCase()`)
- Documentation assertion: `checksummed === lowercase` is `false`, while `areBytes32Equal` is `true`

**With this PR's fix:** all 3 canary tests pass.

```bash
pnpm --filter @layerzerolabs/ua-devtools test:jest -- test/oapp/library-checksum-canary.test.ts
```

tests/ua-devtools-evm-hardhat-test/test/oapp/config.test.ts
```diff
@@ -576,6 +612,60 @@ describe('oapp/config', () => {
...
+                            },
+                        },
+                    ],
+                }
+
+                expect(await configureReceiveLibraries(graph, oappSdkFactory)).toEqual([])
+            })
+        })
+
         describe('configureReceiveLibraryTimeouts', () => {
```
tinom9, line 669: I believe we also want to run the normalized equality checks on `configureReceiveLibraryTimeouts` (`packages/ua-devtools/src/oapp/config.ts:287`).

.changeset/library-samevalue-idempotency.md
```diff
@@ -0,0 +1,14 @@
+---
+"@layerzerolabs/ua-devtools": patch
+---
+
+Fix library wire skip checks with chain-aware address equality
+
+`configureSendLibraries` / `configureReceiveLibraries` (and read-library equivalents)
+previously compared addresses with `toLowerCase()` or strict `===`. Checksum or
+padding mismatches caused already-pinned libraries to be re-proposed, which reverts
```
tinom9, line 9: Do we want to add some simple tests for hex padding and Solana base58 format?

